PR #38, opened 2020-09-24
Fix #37: Change default concurrency in 0.24.x release

### LINKED ISSUE #37: Change default concurrency in 0.24.x release

petrdvorak:


pom.xml
```diff
@@ -57,7 +57,7 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcprov-ext-jdk15on</artifactId>
-            <version>1.65</version>
+            <version>1.66</version>
```
romanstrobl, line 60: I think we should stick to `1.65` for version `0.24.0`, as documented, e.g. here:
https://developers.wultra.com/docs/2020.05/powerauth-server/Installing-Bouncy-Castle#bouncy-castle-on-tomcat
I always try to keep the same version of BC per release to avoid any unexpected behaviour.

romanstrobl (approved):
Looks good, just 1 small nit-pick.

PR #43, opened 2020-09-25
Fix #42: Improve the string validations, comments and depencencies

### LINKED ISSUE #42: Improve the string validations, comments and depencencies

petrdvorak:


romanstrobl (approved):
I find it a bit silly to add a library because of 1 method call on `StringUtils`, but ok...

petrdvorak:
@romanstrobl I actually like it - maybe we can evaluate if new Java has some better String util methods...

petrdvorak:
@romanstrobl Maybe one more comment, the dependency is already in the project, transitively added by SpringDoc. However, I felt this is kinda added by "side-effect" and so I added an explicit dependency.

PR #49, opened 2021-01-02
Extending server into mobile token

petrdvorak:
Please have a look at the PR - if the code is OK, I will create a separate repository for this. The Enrollment Server should probably remain a customizable stub project, this PR feels a bit too "feature-complete" and introduces some undesired dependencies (such as the requirement for the DB).

### DIFF
--- a/src/main/java/com/wultra/app/enrollmentserver/database/OperationTemplateRepository.java
+++ b/src/main/java/com/wultra/app/enrollmentserver/database/OperationTemplateRepository.java
@@ -1,0 +1,37 @@
+/*
+ * PowerAuth Enrollment Server
+ * Copyright (C) 2021 Wultra s.r.o.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.wultra.app.enrollmentserver.database;
+
+import com.wultra.app.enrollmentserver.database.entity.OperationTemplate;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+
+/**
+ * Repository for the operation localization templates.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Repository
+public interface OperationTemplateRepository extends CrudRepository<OperationTemplate, Long> {
+
+    Optional<OperationTemplate> findFirstByLanguageAndPlaceholder(String language, String placeholder);
+
+}
--- a/src/main/java/com/wultra/app/enrollmentserver/database/entity/OperationTemplate.java
+++ b/src/main/java/com/wultra/app/enrollmentserver/database/entity/OperationTemplate.java
@@ -36,8 +36,6 @@ public class OperationTemplate implements Serializable {
     private static final long serialVersionUID = 5914420785283118800L;
 
     @Id
-    @SequenceGenerator(name = "es_operation_template", sequenceName = "es_operation_template_seq")
-    @GeneratedValue(strategy = GenerationType.AUTO, generator = "es_operation_template")
     @Column(name = "id", nullable = false)
     private Long id;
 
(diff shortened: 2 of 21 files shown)


src/main/java/com/wultra/app/enrollmentserver/database/entity/OperationTemplate.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Entity representing an operation template.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Data
+@Entity
+@Table(name = "es_operation_template")
+public class OperationTemplate implements Serializable {
```
romanstrobl, line 34: Should we define nullability for fields in `@Column`?

src/main/java/com/wultra/app/enrollmentserver/database/entity/OperationTemplateParam.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Data
+public class OperationTemplateParam implements Serializable {
+
+    private String type;
+    private String id;
+    private String text;
+    private final Map<String, String> params = new HashMap<>();
```
romanstrobl, line 38: Isn't it better to use a `LinkedHashMap` to keep the same order of parameters during serialization?

src/main/java/com/wultra/app/enrollmentserver/database/entity/OperationTemplateParam.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Class representing an operation template attribute.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Data
+public class OperationTemplateParam implements Serializable {
```
romanstrobl, line 33: Perhaps we need to define `serialVersionUID`, too?

src/main/java/com/wultra/app/enrollmentserver/impl/service/MobileTokenService.java
```diff
@@ -0,0 +1,160 @@
...
+        throw new MobileTokenException("POWERAUTH_AUTH_FAIL", "Authentication failed");
+    }
+
+    private void handleStatus(String status) throws MobileTokenException {
+        switch (status) {
+            case "PENDING": {
+                // OK, this operation is still pending
+                break;
+            }
+            case "CANCELLED": {
```
romanstrobl, line 144: Typo: `CANCELLED` and `OPERATION_ALREADY_CANCELED`, choose one way. I prefer the variant with single L, which we use in Web Flow.

src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -0,0 +1,139 @@
...
+            case "HEADING": {
+                return new HeadingAttribute(id, text);
+            }
+            case "NOTE": {
+                final String noteKey = templateParam.getParams().get("note");
+                final String note = params.get(noteKey);
+                return new NoteAttribute(id, text, note);
+            }
+            case "KEY_VALUE":
+            default: { // attempt fallback to key-value type
```
romanstrobl, line 131: I would prefer to throw an error for an unknown type to avoid surprises (e.g. null values).

romanstrobl (changes requested):
I found several minor issues to consider.

romanstrobl:
Concerning the old JPA generators, I suggest to keep them because they are not a bottleneck for PowerAuth. It is better to keep same JPA generators across the whole stack, I think. Should any performance issue arise, we can migrate to new generators in the whole stack.

src/main/java/com/wultra/app/enrollmentserver/database/OperationTemplateRepository.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Repository for the operation localization templates.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Repository
+public interface OperationTemplateRepository extends CrudRepository<OperationTemplate, Long> {
+
+    Optional<OperationTemplate> findFirstByLanguageAndPlaceholder(String language, String placeholder);
```
romanstrobl, line 35: It is a bit hard to understand what it means to search by `language` and `placeholder`. Is the word `placeholder` the best term to call a localized operation template name? I get it now, but the word placeholder has meanings outside localization and it is not clear that it is a placeholder for an operation name, too. Feel free to ignore.

src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
petrdvorak: We generally handle this with the fallbacks in the mobile app too. I will keep the fallback and add error logging instead.

src/main/java/com/wultra/app/enrollmentserver/database/OperationTemplateRepository.java
petrdvorak: I will probably leave it for now, since calling it "template" would probably cause even more confusion due to different meanings. This is a candidate for the future improvement...

romanstrobl (approved):

PR #52, opened 2021-02-12
Prepare release 1.0.0

petrdvorak:


pom.xml
```diff
@@ -122,19 +107,14 @@
         <dependency>
             <groupId>com.sun.xml.messaging.saaj</groupId>
```
romanstrobl, line 108: Are the libraries related to Web Services still required? If you don't use the old SOAP interface, I think we can remove them.

romanstrobl (changes requested):
Overall looks ok, let's just check that whether we still need all the SOAP-related libraries for Java 11 after migration to REST.

pom.xml
petrdvorak: Good point. We do not need this and we decrease the size of a resulting WAR file from 71.7MB to 48.8MB by removing these dependencies. 👍

romanstrobl (approved):
Looks good

PR #61, opened 2021-05-24
Fix #60: Review logic for signature type conversion

### LINKED ISSUE #60: Review logic for signature type conversion
We should review the logic for signature conversion:

https://github.com/wultra/enrollment-server/blob/54e5d9c845308d1bc97528c84700428022a28ace/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java#L58

This part of code does not make sense to me:
```
if (signatureType.contains("possession")) {            
   allowedSignatureType.setType(AllowedSignatureType.Type.MULTIFACTOR_2FA);        
}
```

petrdvorak:


romanstrobl (approved):
Much better :)

PR #69, opened 2021-06-17
Improvements before release

petrdvorak:
Add operation history endpoint
Add JavaDoc
Add operation approval failure call when authetnication fails
Convert status in operation details


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java
```diff
@@ -106,14 +158,20 @@ public Response operationApprove(@RequestBody ObjectRequest<OperationApproveRequ
                 throw new MobileTokenAuthException();
             }
 
+            final String operationId = requestObject.getId();
+            final String data = requestObject.getData();
+            if (operationId == null) {
+                throw new MobileTokenAuthException();
```
romanstrobl, line 164: We should probably log the error reason, otherwise this will result in a very generic log message.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java
```diff
@@ -106,14 +158,20 @@ public Response operationApprove(@RequestBody ObjectRequest<OperationApproveRequ
...
                 final String userId = auth.getUserId();
                 final Long applicationId = auth.getApplicationId();
-                final String operationId = requestObject.getId();
-                final String data = requestObject.getData();
                 final PowerAuthSignatureTypes signatureFactors = auth.getSignatureFactors();
                 return mobileTokenService.operationApprove(userId, applicationId, operationId, data, signatureFactors);
             } else {
+                // make sure to fail operation as well, to increase the failed number
+                mobileTokenService.operationFailApprove(operationId);
                 throw new MobileTokenAuthException();
```
romanstrobl, line 175: We should probably log the error reason, otherwise this will result in a very generic log message.

romanstrobl (approved):
The changes are ok, I suggest we improve the logging a bit for error handling.

PR #80, opened 2021-06-23
Fix #79: Prepare Oracle script

### LINKED ISSUE #79: Prepare Oracle script

petrdvorak:


docs/sql/oracle/create-schema.sql
```diff
@@ -0,0 +1,28 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+CREATE TABLE ES_OPERATION_TEMPLATE (
+    ID NUMBER(19) NOT NULL PRIMARY KEY,
+    PLACEHOLDER VARCHAR2(255) NOT NULL,
```
romanstrobl, line 21: This should be VARCHAR2(255 CHAR).

docs/sql/oracle/create-schema.sql
```diff
@@ -0,0 +1,28 @@
...
+    LANGUAGE VARCHAR2(8) NOT NULL,
```
romanstrobl, line 22: dtto

docs/sql/oracle/create-schema.sql
```diff
@@ -0,0 +1,28 @@
...
+    TITLE VARCHAR2(255) NOT NULL,
```
romanstrobl, line 23: dtto

romanstrobl (changes requested):
I found a minor issue with the VARCHARs.

docs/sql/oracle/create-schema.sql
romanstrobl: To explain a bit better, `VARCHAR2(255)` is a `VARCHAR` with 255 bytes, which for UTF-8 is much less characters than 255. `VARCHAR2(255 CHAR)` is a `VARCHAR` with 255 characters, which is what we probably want.

romanstrobl (approved):

PR #87, opened 2021-09-14
Fix #86: Do not log activation OTP

### LINKED ISSUE #86: Do not log activation OTP
Activation OTP specified during activation code request is logged on `INFO` level. The OTP code is sensitive, we should avoid logging it.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #101, opened 2021-11-12
Add primary id generators to all es entities

saalistaja:


docs/sql/oracle/create-schema.sql
```diff
@@ -16,6 +16,13 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+--
+--  Create sequences. Maximum value for PostgreSQL is 9223372036854775807.
+--- See: https://www.postgresql.org/docs/9.6/sql-createsequence.html
+--
+CREATE SEQUENCE "es_document_result_seq" MINVALUE 1 MAXVALUE 9223372036854775807 INCREMENT BY 10 START WITH 1 CACHE 20;
```
romanstrobl, line 23: For Oracle we can use `9999999999999999999999999999` as MAXVALUE (28 digits).

romanstrobl (approved):
Let's review the `MAXVALUE` for Oracle DB.

docs/sql/oracle/create-schema.sql
saalistaja: fixed and removed the comment

PR #103, opened 2021-11-15
Remove 1FA signature verification for requests with large data

romanstrobl:


saalistaja (approved):
Good idea, perfect solution!

PR #110, opened 2021-11-30
Reset identity verification process during cleanup

romanstrobl:
The activation flags are reset during cleanup, allowing the next round of identity verification process. The client should call `/api/identity/init` to continue with the next round.


saalistaja (approved):
All changes approved.

PR #111, opened 2021-11-30
Fix JPA queries when multiple verification entities are present per activation

romanstrobl:


saalistaja (approved):
Perfect, this is the fix!

PR #113, opened 2021-12-01
Implement foreign key relationships in JPA, allow empty filter in document status, fine tune database indexes

romanstrobl:


saalistaja (approved):
Well done. This change helps a lot.

PR #117, opened 2021-12-22
Fix #95: Add applicationId (the string ID sent from client) to all callback methods

### LINKED ISSUE #95: Add applicationId (the string ID sent from client) to all callback methods
Currently, only `fetchDestinationApplicationId` contains `String applicationId` as parameter. We should add this to `addActivationFlags` and `didReturnActivationCode` too.

romanstrobl:
I added the parameter, please check the Javadoc if it can be further improved because there are 3 application identifiers present in the methods.


petrdvorak (approved):
Looks OK

PR #125, opened 2022-01-17
Fix #124: Try adding the Docker support right in the repository

### LINKED ISSUE #124: Try adding the Docker support right in the repository
Let's evaluate if we can add Docker support to each executable bundle, instead of to the centralized repository. This is an experimental issue and fix - feel free to comment on this or close, if it does not look convincing...

petrdvorak:
@romanstrobl Let me know what you think about this appoach of keeping Docker support by the particular project...


romanstrobl:
@petrdvorak This pull request is for me an equivalent of adding an almost identical class into three different projects. It works, but it creates a maintenance hurdle which can be avoided. I think we should be making changes into our software which decrease maintenance costs, not increase it, because we spend a lot of time on maintenance already.

romanstrobl:
I am a bit concerned that when building the whole stack on CI using `docker-compose` inside a job we will need to clone all the individual repos which contain all the Dockerfiles and XML configurations and checkout the correct branch. This seems a bit strange.

I personally prefer the approach taken by Oracle, see: https://github.com/oracle/docker-images. But if we want to take the distributed Dockerfile route, then we have to make little sacrifices, I guess. 

Or do you have another suggestion for running CI? I really don't want to duplicate the `Dockerfiles` and XML configurations, so cloning all the individual repos seems necessary. Maybe I am just missing something...

petrdvorak:
> Maybe I am just missing something...

With our current setup, you still need to clone all repositories and build the images. Then you need to manually copy WAR files to the cloned Docker folder with git-ignored subfolder, and build the Docker images. This is more work, than cloning the repo and running `mvn package -Pdocker`...

petrdvorak:
Btw `docker-compose` does not "build the entire stack" - it pulls and launches containers from container repositories. These images already need to be present, and you need to build them independently.

romanstrobl:
> With our current setup, you still need to clone all repositories and build the images.

From the point of view of a customer the wars will be published already on GitHub, so no need to clone all the repos or run any `mvn` for this use case. The process seems more complex now when building a release by a customer, which is also an important use case.

romanstrobl:
> Btw `docker-compose` does not "build the entire stack" - it pulls and launches containers from container repositories. These images already need to be present, and you need to build them independently.

Ok, I see, in the new build process we would rely on individual jobs for publishing the Docker images for individual components, makes sense. Hopefully we will not need any additional logic (scripts, configuration, etc.) because this would be scattered across the individual repositories.

romanstrobl:
Let me try to identify all the use cases we need to support:
1. build Docker images during PowerAuth development:
   - build all PowerAuth components using `mvn package`
   - build Docker images 
   - deploy Docker images for testing
2. build Docker images for publishing the PowerAuth release:
   - download published PowerAuth wars for the release 
   - build Docker images
   - deploy Docker images for testing
   - release Docker images to Docker Hub
3. build Docker images for deployment by customers:
   - download published PowerAuth wars for the release
   - build `enrollment-server` using `mvn package`
   - build Docker images 
   - deploy Docker images for testing or release

It seems to me that each of the use cases has different advantages/disadvantages for the centralized vs. decentralized storage of `Dockerfiles` and variable configuration, depending if it is:
- development - decentralized is better, we work with individual components, so we clone all the repos anyway
- release - centralized is better, we don't want to clone more than just one repo

petrdvorak:
Your analysis actually perfectly explains why Dockerfiles should be decentralized. For development, this is clear - running local build will just do the trick. For release, the customers either take the Docker image as is (PowerAuth Server - when you run the release build, the result is the WAR bundle and published Docker container), or they need to customize the app (Enrollment server), hence they need to change the code and build the customized version, and this again makes Dockerizing easier if the Dockerfile is in the same repo...

Arguments against decentralization would be "We need to have this with 10 linux variants and 5 java versions, and we publish to apt-get and yum repositories anyway..."

romanstrobl:
Ok, I see, we will support the development use case better. Let's merge this change and figure out the base images later.

romanstrobl (approved):
Ok, I guess we can switch to the suggested approach.

PR #128, opened 2022-01-26
Allow run of swagger codegen on JDK16

saalistaja:
- Swagger codegen was failing during the build
- https://github.com/swagger-api/swagger-codegen/issues/10966
- Fixed by setting direct dependency on latest handlebars 4.3.0 library


romanstrobl (approved):
Thanks for the fix. The commit mentions JDK6, maybe we can still ammend the message? If not, then ignore this comment.

PR #131, opened 2022-01-27
Implement presence check restarts with an already enrolled photo

saalistaja:
- Allowed repeated starts of already initialized presence check at expected statuses
- Refactored presence check code a little
- Added more logging
- Removed not thrown DocumentVerificationException from checkPresenceVerification
- Added integration test for repeated presence check
- Fixed verify-token context

Updated diagram at https://lucid.app/lucidchart/4e66ce07-cdfc-4505-a284-e71f385a9299/edit?page=0_0&invitationId=inv_6a5fae3f-23c6-46c4-910e-de92dfd69aa0#


romanstrobl (approved):
Great, thanks for covering the scenario with a test, too.

PR #132, opened 2022-01-28
Allow disabling presence check phase in the identity verification process

saalistaja:
- Added a configuration option to enable/disable presence check
- Adaptate phase/state flow when disabled presence check
- Return bad request when calling services on disabled presence check

### DIFF
--- a/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/IdentityVerificationStatusService.java
+++ b/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/IdentityVerificationStatusService.java
@@ -17,6 +17,7 @@
  */
 package com.wultra.app.enrollmentserver.impl.service;
 
+import com.wultra.app.enrollmentserver.configuration.IdentityVerificationConfig;
 import com.wultra.app.enrollmentserver.database.IdentityVerificationRepository;
 import com.wultra.app.enrollmentserver.database.entity.IdentityVerificationEntity;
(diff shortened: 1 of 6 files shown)


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/IdentityVerificationStatusService.java
```diff
@@ -194,4 +197,21 @@ private void evaluatePresenceCheckResult(OwnerId ownerId,
...
+    /**
+     *
+     *
+     * @param ownerId Owner identification.
+     * @param idVerification Verification identity.
+     */
+    private void startVerification(OwnerId ownerId, IdentityVerificationEntity idVerification) {
+        try {
+            identityVerificationService.startVerification(ownerId);
+        } catch (DocumentVerificationException e) {
```
romanstrobl, line 209: We'll need to merge latest changes from in error handling:
https://github.com/wultra/enrollment-server/pull/134/files#diff-dbc4228c7aa1cf01460a26f414926790d475030a5497e0a03353a2ba36661c1dR179

romanstrobl (approved):
Great, we'll just need to merge the code with recent changes in error handling (separation of identity verification exceptions, document-related exceptions and remote communication exceptions).

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/IdentityVerificationStatusService.java
saalistaja: resolved

PR #133, opened 2022-01-28
Allow asynchronous upload of documents to verification provider

saalistaja:
- Allowed async upload and data extraction in the used document provider
- Introduced a regular task to check process of document upload and data extraction at the provider side
- Added scheduler lock support to prevent parallel run of regular tasks
- Adapted and fixed unit tests to latest changes

### DIFF
--- a/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/DocumentResultRepository.java
+++ b/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/DocumentResultRepository.java
@@ -36,10 +36,10 @@ public interface DocumentResultRepository extends CrudRepository<DocumentResultE
     /**
      * @return All not finished document uploads (in progress status and no extracted data filled)
      */
-    @Query("select doc from DocumentResultEntity doc where" +
+    @Query("SELECT doc FROM DocumentResultEntity doc WHERE" +
             " doc.documentVerification.status = com.wultra.app.enrollmentserver.model.enumeration.DocumentStatus.UPLOAD_IN_PROGRESS" +
-            " and doc.extractedData is null " +
-            " order by doc.timestampCreated asc")
+            " AND doc.extractedData IS NULL " +
+            " ORDER BY doc.timestampCreated ASC")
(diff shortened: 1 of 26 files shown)


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/DocumentResultRepository.java
```diff
@@ -30,4 +33,13 @@
 @Repository
 public interface DocumentResultRepository extends CrudRepository<DocumentResultEntity, Long> {
 
+    /**
+     * @return All not finished document uploads (in progress status and no extracted data filled)
+     */
+    @Query("select doc from DocumentResultEntity doc where" +
```
romanstrobl, line 39: Perhaps we should capitalize SQL commands like in other queries?

enrollment-server/src/test/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProviderTest.java
```diff
@@ -47,7 +52,7 @@
  * @author Lukas Lukovsky, lukas.lukovsky@wultra.com
  */
 @SpringBootTest(classes = EnrollmentServerTestApplication.class)
-@ActiveProfiles("external-service")
+@ActiveProfiles({"external-service"})
```
romanstrobl, line 55: Is this change required?

romanstrobl (approved):
Nice addition, I found several minor issues.

enrollment-server/src/test/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProviderTest.java
saalistaja: aah, thanks
I had two profiles there and removed the other, no need for an array any more

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/DocumentResultRepository.java
saalistaja: fixed

PR #134, opened 2022-01-31
Check activation flags during status 

romanstrobl:
- Check activation flags during status to identify state when re-initialization is required after cleanup.
- Improve error handling.


saalistaja (approved):
Looks good.

PR #135, opened 2022-02-01
Use secured user identification in iProov API requests

saalistaja:
- Allows API cost optimization
- Hides potentially sensitive user identification with hashed value


romanstrobl (approved):
Good, works for me.

PR #136, opened 2022-02-01
Separate api and domain models to own project modules

saalistaja:
- Moved all domain model classes to `enrollment-server-domain-model` module
- Moved all api model classes to `enrollment-server-api-model` module


romanstrobl (approved):
Perfect, thanks.

PR #137, opened 2022-02-03
Token-based authentication and onboarding process changes

romanstrobl:
- token-based authentication is used for status and for document endpoints where signatures cannot be used
- onboarding process is now terminated after identity verification, not immediately after activation
- `processId` parameter is sent and verified for all endpoints where it is applicable to improve auditing and improve request validation
- `process_id` field added to table `es_identity_verification`
- unified service method calls - do not pass `PowerAuthApiAuthentication` parameter into services, pass `OwnerId` instead


saalistaja (approved):
Looks good.

PR #138, opened 2022-02-03
Fix securing of the userId value

saalistaja:


romanstrobl (approved):
Yes, this should do it.

PR #141, opened 2022-02-09
Add verification during document submit

saalistaja:
- Allow immediate verification of submitted documents
- Keeping documents as UPLOAD_IN_PROGRESS until the verification finishes
- Added a regular task to check verification results on submitted documents


romanstrobl (approved):
Looks great, thanks!

PR #143, opened 2022-02-10
Allow multiple document submit

saalistaja:
- Prevent change of uploaded timestamp


romanstrobl (approved):
Simple improvement, nice.

PR #146, opened 2022-02-14
Changes of OTP delivery based on requirements from mobile application

romanstrobl:
- Send OTP when changing status automatically.
- Change `otp/send` endpoint to `otp/resend`.
- Return only 1 result when checking OTP detail.


saalistaja (approved):
More robust and simple approach than before, very well!

PR #147, opened 2022-02-14
Always send OTP code after completed document verification

romanstrobl:
Always send OTP code after completed document verification, process document verification results later when OTP verification is enabled.


saalistaja (approved):
Changes explained, it's ok.

PR #151, opened 2022-02-15
Allow repeated presence check initialization

saalistaja:
- temporary solution until the enrolled entity deletion is implemented


romanstrobl (approved):
Ok, I understand that this is just a temporary workaround, let's use it for now.

PR #153, opened 2022-02-15
Add custom simple validations on submitted documents

saalistaja:
- check document type
- check card side

### DIFF
--- a/enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
+++ b/enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
@@ -543,7 +543,7 @@ public class ZenidDocumentVerificationProvider implements DocumentVerificationPr
     @Nullable
     private DocumentType toDocumentType(ZenidSharedMineAllResult.DocumentRoleEnum documentRoleEnum) {
         switch (documentRoleEnum) {
-            case CAR:
+            case DRV:
                 return DocumentType.DRIVING_LICENSE;
             case IDC:
                 return DocumentType.ID_CARD;
--- a/enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
+++ b/enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
@@ -371,7 +371,9 @@ public class ZenidDocumentVerificationProvider implements DocumentVerificationPr
         }
         DocumentType zenIdDocType = minedData.getDocumentRole() == null ?
                 null : toDocumentType(minedData.getDocumentRole());
-        if (zenIdDocType == null || documentType != zenIdDocType) {
+        if (documentType != zenIdDocType) {
+            logger.info("Received different document type {} ({}) than expected {} from ZenID, {}",
+                    zenIdDocType, minedData.getDocumentRole(), documentType, id);
             documentSubmitResult.setRejectReason(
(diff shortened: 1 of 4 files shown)


enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
```diff
@@ -499,4 +544,29 @@ private DocumentVerificationStatus toStatus(ZenidWebInvestigateResponse.StateEnu
...
+    private DocumentType toDocumentType(ZenidSharedMineAllResult.DocumentRoleEnum documentRoleEnum) {
+        switch (documentRoleEnum) {
+            case DRV:
+                return DocumentType.DRIVING_LICENSE;
+            case IDC:
+                return DocumentType.ID_CARD;
+            case PAS:
+                return DocumentType.PASSPORT;
+            default:
+                return null;
```
romanstrobl, line 557: I am not sure about the `null` value. I would like to suggest to add a new `DocumentType.UNKNOWN` enum value, what do you think?

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
```diff
@@ -499,4 +544,29 @@ private DocumentVerificationStatus toStatus(ZenidWebInvestigateResponse.StateEnu
...
+        }
+    }
+
+    private CardSide toCardSide(ZenidSharedMineAllResult.PageCodeEnum pageCodeEnum) {
+        switch (pageCodeEnum) {
```
romanstrobl, line 562: I also suggest to handle the `null` value of `pageCodeEnum` parameter just to be sure.

enrollment-server/src/main/resources/application.properties
```diff
@@ -118,6 +118,7 @@ enrollment-server.presence-check.iproov.serviceHostname=${IPROOV_SERVICE_HOSTNAM
 spring.security.oauth2.client.provider.app.token-uri=http://localhost:6060/oauth/token
 
 # ZenID configuration
+enrollment-server.document-verification.zenid.additionalDocSubmitValidationsEnabled=${ZENID_ADDITIONAL_DOC_SUBMIT_VALIDATIONS_ENABLED:false}
```
romanstrobl, line 121: I am not sure we need to configure this aspect of the application. If the validations work properly, we should probably just use them. We already have a lot of configuration properties.

romanstrobl (approved):
Nice addition. I have several suggestions for changes, but feel free to ignore them, these are just minor points.

enrollment-server/src/main/resources/application.properties
saalistaja: My aim was to pilot the feature slowly, just to be able to turn it off when we would need.
I will still add a TODO for now, we can reconsider removing the configuration later.

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
saalistaja: The null check is before calling this method, but I fixed it to be more robust at this place.

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
saalistaja: Fixed

PR #155, opened 2022-02-16
Enable encryption in identity cleanup endpoint, consistently check process ID wherever possible

romanstrobl:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OnboardingService.java
```diff
@@ -228,6 +227,26 @@ public Response performCleanup(OnboardingCleanupRequest request) throws Onboardi
...
+    public void verifyProcessId(OwnerId ownerId, String processId) throws OnboardingProcessException {
+        Optional<OnboardingProcessEntity> processOptional = onboardingProcessRepository.findProcessByActivationId(ownerId.getActivationId());
+        if (!processOptional.isPresent()) {
+            logger.error("Onboarding process not found, {}", ownerId);
+            throw new OnboardingProcessException();
+        }
+        String expectedProcessId = processOptional.get().getId();
+
+        if (!expectedProcessId.equals(processId)) {
+            logger.warn("Invalid process ID received in request: {}", processId);
```
saalistaja, line 245: Maybe we should log also the ownerId context here? But this will be solved also later by logging with MDC.

saalistaja (approved):
Nice bunch of changes. One small comment but feel free to merge it as it is.
I will adapt the cleanup call in https://github.com/wultra/powerauth-tests/pull/41 then.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OnboardingService.java
romanstrobl: Yes, good point, I will change it.

romanstrobl:
I already adapted the cleanup call: https://github.com/wultra/powerauth-tests/pull/41/commits/fae7ad91fcbccc483c1817ff34aa0f884b522022

PR #157, opened 2022-02-22
Resolve #156: Use consistent DocumentMetadata in responses

### LINKED ISSUE #156: Use consistent DocumentMetadata
We should use consistent data, currently two fields (type, side) or only in one entity. Consider using one common `DocumentMetadata`.

https://github.com/wultra/enrollment-server/blob/issues/88-customer-identification-verification/enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/response/DocumentSubmitResponse.java

https://github.com/wultra/enrollment-server/blob/issues/88-customer-identification-verification/enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/response/DocumentStatusResponse.java

saalistaja:


enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/response/data/DocumentMetadataResponseDto.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Processing status of the document
+     */
+    private DocumentStatus status;
+
+    /**
+     * Errors discovered during processing of the document
+     */
+    private List<String> errors;
```
romanstrobl, line 63: Very minor suggestion: perhaps we could initialize the List as an empty `ArrayList` and make it final to avoid null value handling?

romanstrobl (approved):
Nice improvement, I have just one very minor suggestion, feel free to ignore.

enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/response/data/DocumentMetadataResponseDto.java
saalistaja: There is still a big doubt in me with such defaults. If it should be lazy on getter, initialized in all cases even when not needed, or just checked by the caller. I will leave now as it is.

PR #158, opened 2022-02-22
Use our REST client for iProov calls

saalistaja:


romanstrobl (approved):
I am glad we could use our own REST client. Nice touch with specifying the user agent.

PR #163, opened 2022-02-24
Fix #162: Add flag conditioning for operations APIs

### LINKED ISSUE #162: Add flag conditioning for operations APIs

petrdvorak:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/entity/OperationTemplate.java
```diff
@@ -18,17 +18,25 @@
...
 
 /**
  * Entity representing an operation template.
  *
  * @author Petr Dvorak, petr@wultra.com
  */
-@Data
+@Getter
+@Setter
+@ToString
```
romanstrobl, line 38: I suggest to add a `@NoArgsConstructor` because the class implements serializable interface, which indicates serialization is possible. Some frameworks (e.g. Jackson ObjectMaper) fail to deserialize classes with missing no-args constructor.

romanstrobl (approved):
Looks good, I just added one minor idea for improvement.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/entity/OperationTemplate.java
petrdvorak: Good suggestion - added.

PR #164, opened 2022-02-25
Improve OTP verification and timeout configuration

saalistaja:
- In case timeout for SMS OTP is exceeded, the verification attempt should fail
- Providing configured timeout value of the OPT resend period in the `api/identity/status` response
- Refactored OTP code verification to cover more cases
- Tracking OTP expiration time, using it for expiration check

Sql update for already existing database schemas:
```sql
alter table es_onboarding_otp add timestamp_expiration timestamp;

update es_onboarding_otp
set timestamp_expiration = timestamp_created + interval '30 seconds'
where timestamp_expiration is null;

alter table es_onboarding_otp alter column timestamp_expiration set not null;
```


petrdvorak (changes requested):
When creating OTP, please store the expiration timestamp in a separate DB column. We need to provide a way to verify that SMS was really approved before expiration, just having property that can change does not provide it.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/entity/OnboardingOtpEntity.java
```diff
@@ -82,6 +82,9 @@
     @Column(name = "timestamp_created", nullable = false)
     private Date timestampCreated;
 
+    @Column(name = "timestamp_expiration")
+    private Date timestampExpiration;
```
romanstrobl, line 86: Missing `nullable=false`.

romanstrobl (approved):
Implemented as expected, I just found one minor missing nullability check configuration.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/entity/OnboardingOtpEntity.java
saalistaja: Good catch! Fixed now.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OtpService.java
```diff
@@ -129,34 +128,46 @@ public OtpVerifyResponse verifyOtpCode(String processId, String otpCode, OtpType
...
+            expired = true;
+            otp.setStatus(OtpStatus.FAILED);
+            otp.setErrorDetail(OnboardingOtpEntity.ERROR_EXPIRED);
+            otp.setTimestampLastUpdated(new Date());
+            onboardingOtpRepository.save(otp);
+        } else if (otp.getOtpCode().equals(otpCode)) {
+            verified = true;
+            otp.setStatus(OtpStatus.VERIFIED);
+            otp.setTimestampVerified(new Date());
+            otp.setTimestampLastUpdated(new Date());
```
petrdvorak, line 160: Store `new Date()` in variable in the beginning, otherwise you can have different dates in subsequent lines. We should review all places with the `new Date()` code.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OtpService.java
```diff
@@ -91,14 +91,10 @@ public String createOtpCode(OnboardingProcessEntity process, OtpType otpType) th
...
         // Do not allow spamming by OTP codes
-        Date lastDate = onboardingOtpRepository.getNewestOtpCreatedTimestamp(processId, otpType);
-        int resendPeriod = onboardingConfig.getOtpResendPeriod();
-        Calendar c = GregorianCalendar.getInstance();
-        c.setTime(lastDate);
-        c.add(Calendar.SECOND, resendPeriod);
-        Date allowedDate = c.getTime();
-        if (allowedDate.after(new Date())) {
-            logger.warn("Resend OTP functionality is not available yet for process ID: {}", processId);
+        Date otpLastCreatedDate = onboardingOtpRepository.getNewestOtpCreatedTimestamp(processId, otpType);
```
petrdvorak, line 94: Consider adding `final` (also to the variables below).

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OtpService.java
```diff
@@ -165,8 +176,9 @@ public OtpVerifyResponse verifyOtpCode(String processId, String otpCode, OtpType
         OtpVerifyResponse response = new OtpVerifyResponse();
         response.setProcessId(processId);
         response.setOnboardingStatus(process.getStatus());
+        response.setExpired(expired);
         response.setVerified(verified);
-        response.setRemainingAttempts(remainingAttempts);
+        response.setRemainingAttempts(expired ? 0 : maxFailedAttempts - failedAttempts);
```
petrdvorak, line 181: I am not sure if this is correct - we are losing information "expired with 3 remaining attempts"

petrdvorak (changes requested):
Please have a look at the review comments.

petrdvorak (approved):
Looks good

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OtpService.java
saalistaja: Fixed

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OtpService.java
saalistaja: Fixed

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OtpService.java
saalistaja: Fixed

PR #165, opened 2022-02-25
Add preconditions check to otp send call

saalistaja:
- expected phase/status is OTP_VERIFICATION / OTP_VERIFICATION_PENDING


romanstrobl (approved):
Looks great. I was also recently thinking about adding constructors with messages for all exception classes to improve logging.

PR #167, opened 2022-03-01
Improve selection of the person photo for presence check

saalistaja:


romanstrobl (approved):
Excellent, just what was needed.

PR #168, opened 2022-03-01
Speed up mining of document data with query params

saalistaja:


romanstrobl (approved):
Nice speed improvement!

PR #172, opened 2022-03-17
Fix #171: Add support for request context calls

### LINKED ISSUE #171: Add support for request context calls
We should pass IP address, activation ID and user agent to PAS when approving or rejecting operation.

petrdvorak:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/RequestContextConverter.java
```diff
@@ -0,0 +1,85 @@
...
+     * @param request HttpServletRequest instance.
+     * @return Best-effort information about the client IP address.
+     */
+    private String getClientIpAddress(HttpServletRequest request) {
+        if (request == null) { // safety null check
+            return null;
+        }
+        for (String header: HTTP_HEADERS_IP_ADDRESS) {
+            final String ip = request.getHeader(header);
+            if (ip != null && ip.length() != 0 && !"unknown".equalsIgnoreCase(ip)) {
```
romanstrobl, line 78: Can we use `!ip.isEmpty()` instead of `ip.length() != 0`? I think it's easier to understand.

romanstrobl (approved):
Looks good. I have just one very small idea for improvement of readability.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/MobileTokenService.java
```diff
@@ -146,6 +151,10 @@ public Response operationApprove(
         approveRequest.setUserId(userId);
         approveRequest.setSignatureType(SignatureType.enumFromString(signatureFactors.name())); // 'toString' would perform additional toLowerCase() call
         approveRequest.setApplicationId(applicationId);
+        // Prepare additional data
+        approveRequest.getAdditionalData().put("activation_id", activationId);
```
romanstrobl, line 155: Perhaps it would be better to define constants for these 3 strings used in additional data.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/RequestContextConverter.java
petrdvorak: Fixed

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/MobileTokenService.java
petrdvorak: Fixed

PR #176, opened 2022-04-04
Resolve #175: Add an API to initialize ZenID SDK

### LINKED ISSUE #175: Add init SDK API for ZenID integration

saalistaja:


enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
```diff
@@ -323,6 +325,40 @@ public void cleanupDocuments(OwnerId id, List<String> uploadIds) throws Document
...
+    @Override
+    public VerificationSdkInfo initVerificationSdk(OwnerId id, Map<String, String> initAttributes) throws DocumentVerificationException {
+        Preconditions.checkArgument(initAttributes.containsKey(ZenidConst.SDK_INIT_TOKEN), "Missing initialization token for ZenID SDK");
+        String token = initAttributes.get(ZenidConst.SDK_INIT_TOKEN);
+
+        ResponseEntity<ZenidWebInitSdkResponse> responseEntity;
+        try {
+            responseEntity = zenidApiService.initSdk(token);
+        } catch (RestClientException e) {
+            logger.warn("Failed REST call to init ZenID SDK, " + id, e);
```
banterCZ, line 337: Why don't we write `...init ZenID SKD, {}", id, e`?

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
```diff
@@ -323,6 +325,40 @@ public void cleanupDocuments(OwnerId id, List<String> uploadIds) throws Document
...
+            logger.error("Unexpected error when initializing ZenID SDK, " + id, e);
+            throw new DocumentVerificationException("Unexpected error when initializing ZenID SDK");
+        }
+
+        if (responseEntity.getBody() == null) {
+            logger.error("Missing response body when initializing ZenID SDK, " + id);
+            throw new DocumentVerificationException("Unexpected error when initializing ZenID SDK");
+        }
+
+        if (!HttpStatus.OK.equals(responseEntity.getStatusCode())) {
```
banterCZ, line 349: What about to call `!responseEntity.getStatusCode().is2xxSuccessful()`

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/service/ZenidRestApiService.java
```diff
@@ -184,6 +181,18 @@ public String getFilename() {
...
     }
 
+    /**
+     * Initializes the SDK of ZenID
+     *
+     * @param token Initialization token
+     * @return Response entity with the SDK initialization result
+     */
+    public ResponseEntity<ZenidWebInitSdkResponse> initSdk(String token) {
+        String apiPath = String.format("/api/initSdk?token=%s", token);
```
banterCZ, line 191: UriVariables may be passed directly to the method `exchange(String url, HttpMethod method, @Nullable HttpEntity<?> requestEntity, Class<T> responseType, Object... uriVariables)`

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
saalistaja: Because I want also the logged exception with the message.

The implementation of logback does not seem to try to detect it from multiple arg call
```
    public void warn(String format, Object arg1, Object arg2) {
        this.filterAndLog_2(FQCN, (Marker)null, Level.WARN, format, arg1, arg2, (Throwable)null);
    }
```

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
saalistaja: Yeps, this is also a possible but it is much more wide variant. The REST API is returning 200, all other response codes I do consider as strange and weird.

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/service/ZenidRestApiService.java
saalistaja: Nice tip! But I will not include it here as there would be an inconsistent approach as the building of full url path is on all other places already. We could apply it later during refactor to our https://github.com/wultra/lime-java-core

enrollment-server/src/main/java/com/wultra/app/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
banterCZ: Hmm, I see, it is strange to me, because there is slf4j facade and developers should not care (much) about underlying implementation. Ok, let it be, but I will double check the logback + slf4j.

romanstrobl (approved):
No issues found.

PR #179, opened 2022-04-07
Use slf4j logging of parametrized messages with an exception

saalistaja:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/IdentityVerificationStatusService.java
```diff
@@ -192,11 +192,11 @@ public IdentityVerificationStatusResponse checkIdentityVerificationStatus(Identi
                     identityVerificationFinishService.finishIdentityVerification(ownerId);
                 }
             } catch (DocumentVerificationException e) {
-                logger.error("Checking identity verification result failed, " + ownerId, e);
+                logger.error("Checking identity verification result failed, {}", ownerId, e);
                 response.setIdentityVerificationStatus(IdentityVerificationStatus.FAILED);
                 return response;
             } catch (OnboardingProcessException e) {
-                logger.error("Updating onboarding process failed, " + ownerId, e);
+                logger.error("Updating onboarding process failed, {}", ownerId, e);
```
banterCZ, line 199: Out of scope of this PR, but what do you think about writing `Updating onboarding process failed, ownerId={}`, IMHO it provides better context for readers.

banterCZ (approved):

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/IdentityVerificationStatusService.java
saalistaja: Well it is a way but with repeated words. The logged message would look like `ownerId=OwnerId(activationId=123,userId=456)` which I am not fan of.

banterCZ: Oh, I see, you are right, I have missed that `ownerId` is not a primitive.

romanstrobl (approved):
Nice improvement!

PR #184, opened 2022-04-14
Use OID as the image source for iProov

saalistaja:


romanstrobl (approved):
Let's test this change, hopefully this is correct.

PR #203, opened 2022-06-12
Fix #202: Decrease increment for the template configuration table

### LINKED ISSUE #202: Decrease increment for the template configuration table
The `es_operation_template_seq` refers to table that is not traffic driven, but configuration. We do not need efficient allocation there, having a clear sequential records is beneficial.

petrdvorak:


docs/sql/oracle/create-schema.sql
```diff
@@ -17,7 +17,7 @@
  */
 
 CREATE SEQUENCE "es_document_result_seq" MINVALUE 1 MAXVALUE 9999999999999999999999999999 INCREMENT BY 10 START WITH 1 CACHE 20;
-CREATE SEQUENCE "es_operation_template_seq" MINVALUE 1 MAXVALUE 9999999999999999999999999999 INCREMENT BY 10 START WITH 1 CACHE 20;
+CREATE SEQUENCE "es_operation_template_seq" MINVALUE 1 MAXVALUE 9999999999999999999999999999 INCREMENT BY 1 START WITH 1 CACHE 20;
```
banterCZ, line 20: Is this sequence new? Don't we need to update existing ones as well?

banterCZ (approved):

docs/sql/oracle/create-schema.sql
romanstrobl: I think this change requires a DB migration for both `enrollment-server` and `powerauth-cloud`.

petrdvorak: No, it does not. This is how it was until you changed it...

romanstrobl: I doubt it was me (I am just author of large merge commit), but I see, it's a mistake.

PR #217, opened 2022-06-15
Fix #216: Consent for Onboarding

### LINKED ISSUE #216: Consent for Onboarding
Implement consent for onboarding.

- obtain consent text
- store approval

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/OnboardingController.java
```diff
@@ -183,4 +187,50 @@ public Response performCleanup(@EncryptedRequestBody ObjectRequest<OnboardingCle
...
+            description = "Obtain a text of user consent in specified language."
+    )
+    public ObjectResponse<OnboardingConsentTextResponse> fetchConsentText(final @RequestBody OnboardingConsentTextRequest request) throws OnboardingProcessException{
+        logger.debug("Returning consent for {}", request);
+        validate(request);
+        final OnboardingConsentTextResponse onboardingConsentTextResponse = onboardingService.fetchConsentText(request);
+        return new ObjectResponse<>(onboardingConsentTextResponse);
+    }
+
+    @PostMapping("consent/approve")
```
banterCZ, line 202: I am open to discuss this path.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingService.java
```diff
@@ -316,6 +318,58 @@ public void terminateInactiveProcesses() {
...
+                .build();
+
+        try {
+            onboardingProvider.approveConsent(providerRequest);
+        } catch (OnboardingProviderException e) {
+            throw new OnboardingProcessException("An error when approving consent.", e);
+        }
+    }
+
+    private static com.wultra.app.onboardingserver.provider.ConsentType convert(final ConsentType source) {
```
banterCZ, line 365: This seems like duplicity in `ConsentType` but I want to isolate possible changes in REST a provider API.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/OnboardingProvider.java
```diff
@@ -32,4 +32,21 @@ public interface OnboardingProvider {
...
     void sendOtpCode(String userId, String otpCode, boolean resend) throws OnboardingProviderException;
 
+    /**
+     * Provide consent text.
+     *
+     * @param request consent text request
+     * @return consent text
+     * @throws OnboardingProviderException if there is a problem to fetch consent text
+     */
+    String fetchConsent(ConsentTextRequest request) throws OnboardingProviderException;
```
banterCZ, line 42: _Context object_ approach makes keeping backward compatibility easier. We may add parameters and don't break compatibility.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/OnboardingController.java
```diff
@@ -183,4 +187,50 @@ public Response performCleanup(@EncryptedRequestBody ObjectRequest<OnboardingCle
...
+            throw new ValidationException("Missing mandatory attributes");
+        }
+        try {
+            LocaleUtils.toLocale(request.getLanguage());
+        } catch (IllegalArgumentException e) {
+            throw new ValidationException("Invalid language: " + request.getLanguage(), e);
+        }
+    }
+
+    private static void validate(final OnboardingConsentApprovalRequest request) {
```
banterCZ, line 228: Or do you prefer to replace it by _javax.validation_?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/OnboardingController.java
petrdvorak: This looks sufficient. 👍

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/OnboardingController.java
petrdvorak: Typically, we would isolate the logic into a separate `OnboardingConsentApprovalRequestValidator` class.

enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/request/ConsentType.java
```diff
@@ -0,0 +1,28 @@
...
+package com.wultra.app.enrollmentserver.api.model.request;
+
+/**
+ * Consent type.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public enum ConsentType {
+
+    GDPR
```
petrdvorak, line 27: I think we need to have this freeform string. Can you name consent types the customers will use? Validation can be done on some other layer.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingService.java
petrdvorak: this method should be isolated to separate converter class.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingService.java
```diff
@@ -316,6 +318,58 @@ public void terminateInactiveProcesses() {
...
+            throw new OnboardingProcessException("An error when fetching consent text.", e);
+        }
+    }
+
+    /**
+     * Record dis/approval of consent
+     *
+     * @param request approval request
+     */
+    public void approveConsent(final OnboardingConsentApprovalRequest request) throws OnboardingProcessException {
```
petrdvorak, line 350: Let's consider if we can live with void method here.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/ConsentType.java
```diff
@@ -0,0 +1,28 @@
...
+package com.wultra.app.onboardingserver.provider;
+
+/**
+ * Consent type.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public enum ConsentType {
+
+    GDPR
```
petrdvorak, line 27: dtto - not sure if we can live with enum here...

petrdvorak (changes requested):
Please consider several changes, maily separating validator/converter logic from the service/controller classes.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingService.java
banterCZ: 👍  You are right. From compatibility point of view, it's better to return even empty object expandable in the future.

enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/request/ConsentType.java
romanstrobl: Agree.

banterCZ: Changed to String.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/ConsentType.java
banterCZ: Changed to String.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingService.java
romanstrobl: Agree with empty return type.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/OnboardingProvider.java
romanstrobl: This is a good idea, however we should stick to one approach within this interface, we use two approaches now (see methods above).

romanstrobl (approved):
Nice improvements. Let's finalize the migration of consent type to String and keep the interface tidy.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/ApproveConsentRequest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.UUID;
+
+/**
+ * Request object for {@link OnboardingProvider#approveConsent(ApproveConsentRequest)}.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@Builder
+@Getter
```
banterCZ, line 31: This should be annotated by `PublicApi` when #222 merged

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/OnboardingProvider.java
banterCZ: Yes, I was about make it consinstent in a separate PR.

PR #209, opened 2022-06-14
Fix #208: Change compiler target from 9 to 8

banterCZ:

PR #228, opened 2022-06-21
Fix #227: Setup Compiler for Java 11 LTS

### LINKED ISSUE #227: Setup Compiler for Java 11 LTS
Setup Compiler for Java 11 LTS. In pom.xml, there is confusing mixture of 8 and 9.

banterCZ:


romanstrobl (approved):
Yes, we can use Java 11 (which is an LTS version) as a minimum supported version. We should probably reflect this change in documentation.

banterCZ:
> Yes, we can use Java 11 (which is an LTS version) as a minimum supported version. We should probably reflect this change in documentation.

In Deploying-Enrollment-Server.md there is:
> ## Supported Java Runtime Versions
> 
> The following Java runtime versions are supported:
> - Java 11 (LTS release)
> - Java 17 (LTS release)

romanstrobl:
Ah ok, then it is fine.

saalistaja:
Thus can we then rollback all the nice merged changes from https://github.com/wultra/enrollment-server/pull/209 ?

banterCZ:
> Thus can we then rollback all the nice merged changes from #209 ?

#209 has never been merged, see https://github.com/wultra/enrollment-server/pull/209#issuecomment-1156159362

saalistaja:
> > Thus can we then rollback all the nice merged changes from #209 ?
> 
> #209 has never been merged, see [#209 (comment)](https://github.com/wultra/enrollment-server/pull/209#issuecomment-1156159362)

Thanks God! :)

PR #235, opened 2022-06-22
Fix #233: Separate application.properties in enrollment and onboarding

### LINKED ISSUE #233: Separate application.properties in enrollment and onboarding
After split of enrollment server into two applications we should clean up `application.properties`.

romanstrobl:


enrollment-server/src/main/resources/application.properties
```diff
@@ -76,80 +76,6 @@ powerauth.service.security.clientSecret=
...
-enrollment-server.presence-check.iproov.restClientConfig.proxyEnabled=false
-enrollment-server.presence-check.iproov.restClientConfig.proxyHost=
-enrollment-server.presence-check.iproov.restClientConfig.proxyPort=0
-enrollment-server.presence-check.iproov.restClientConfig.proxyUsername=
-enrollment-server.presence-check.iproov.restClientConfig.proxyPassword=
-
-spring.security.oauth2.client.provider.app.token-uri=http://localhost:6060/oauth/token
-
-# ZenID configuration
-enrollment-server.document-verification.zenid.apiKey=${ZENID_API_KEY}
```
banterCZ, line 133: So this belongs to the downstream project?

banterCZ (approved):

enrollment-server/src/main/resources/application.properties
banterCZ: I see, I have to expand, it is here https://github.com/wultra/enrollment-server/pull/235/files#diff-53ea8234fd009b798ac2ab82ae426290ba589a47313322d4215a7728152dedd2L129

romanstrobl: Yes, this has nothing to do with downstream, just a bit of mess due to duplicit configuration and older property names as left overs.

PR #237, opened 2022-06-22
Fix #234: Missing DocumentVerificationProvider

### LINKED ISSUE #234: Missing DocumentVerificationProvider
In case property `enrollment-server-onboarding.document-verification.provider` is not configured in `enrollment-server`, the startup fails with following error:

```
No qualifying bean of type 'com.wultra.app.onboardingserver.provider.DocumentVerificationProvider' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

We should configure the application so that this property is not required, because it is only related to onboarding.

romanstrobl:
This change will cause that mock providers are autowired in case no provider is specified.


banterCZ (approved):
Approving and just wondering about empty implementation such as `EmptyOnboardingProvider`.

romanstrobl:
Feel free to raise another ticket for improvement.

PR #240, opened 2022-06-23
Fix #238: Move consent endpoints into IdentityVerificationController

### LINKED ISSUE #238: Move consent endpoints into IdentityVerificationController
The `OnboardingController` contains endpoints which are authenticated using `APPLICATION_SCOPE` and generally these endpoints are related to the activation process. I believe we should move consent endpoints into `IdentityVerificationController`, because this controller uses `ACTIVATION_SCOPE` and we will also use this scope for consent authentication. 

Furthermore, consent processing is the next step after `/api/identity/init` endpoint which is present in `IdentityVerificationController`.

banterCZ:


romanstrobl (approved):
Great, thanks!

PR #243, opened 2022-06-23
Fix #242: Configure Maven for snapshot release of libraries

### LINKED ISSUE #242: Configure Maven for snapshot release of libraries
We should configure Maven to allow deployment of snapshot for libraries created in the `enrollment-server` project, but disallow deployment of war files.

romanstrobl:


enrollment-server/pom.xml
```diff
@@ -168,6 +168,13 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-deploy-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
```
banterCZ, line 175: Or just use a shorter way and define the maven property `maven.deploy.skip` https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html#skip

romanstrobl: It's ok this way for consistency reasons.

PR #245, opened 2022-06-24
Fix #244: Skip deploying of onboarding war

### LINKED ISSUE #244: Skip deploying onboarding war
Do not deploy onboarding war to maven repository.

banterCZ:


enrollment-server-onboarding/pom.xml
```diff
@@ -157,6 +157,13 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-deploy-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
```
banterCZ, line 164: I would rather use only `maven.deploy.skip` https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html#skip but prefer to be consistent.

romanstrobl: We can change it one day in all projects.

romanstrobl (changes requested):
My concern is that this configuration would stop deployment of both `war` and `jar` files. Can you please check if this is the case?

banterCZ:
> My concern is that this configuration would stop deployment of both `war` and `jar` files. Can you please check if this is the case?

Yes, you are absolutely correct. Thanks! We should go this wya https://maven.apache.org/plugins/maven-deploy-plugin/examples/deploying-with-classifiers.html

romanstrobl (approved):
Approved.

PR #253, opened 2022-06-27
Fix #252: Add support to return CONSENT_REQUIRED

### LINKED ISSUE #252: Support to receive consent required attribute
Add support to return receive `consent required` flag. It will be ignored so far and not propagated to the client, which means  always required.

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/EmptyOnboardingProvider.java
```diff
@@ -27,7 +27,7 @@
 class EmptyOnboardingProvider implements OnboardingProvider {
 
     @Override
-    public String lookupUser(LookupUserRequest request) throws OnboardingProviderException {
+    public LookupUserResponse lookupUser(LookupUserRequest request) throws OnboardingProviderException {
```
banterCZ, line 30: This should be enough so far. We are able to receive the flag but do not propagate to the client yet.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/LookupUserResponse.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@Builder
+@Getter
+@PublicApi
+public final class LookupUserResponse {
+
+    @NonNull
+    private String userId;
+
+    private boolean consentRequired;
```
romanstrobl, line 38: Given that the default logic is to always show consent I suggest to set `consentRequired` = `true` by default.

romanstrobl (approved):
I have one suggestion for improvement.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/LookupUserResponse.java
banterCZ: Set default value to `true`.

PR #255, opened 2022-06-27
Fix #254: Remove any traces of onboarding code from the enrollment server

### LINKED ISSUE #254: Remove any traces of onboarding code from the enrollment server
The point of separating enrollment server from the onboarding component was to completely remove unfinished code from enrollment server and add it only for those customers who explicitly need to use onboarding.

petrdvorak:


enrollment-server-onboarding-common/pom.xml
```diff
@@ -27,16 +27,15 @@
         <version>1.4.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>enrollment-server-common</artifactId>
-    <name>enrollment-server-common</name>
+    <artifactId>enrollment-server-onboarding-common</artifactId>
```
banterCZ, line 30: Shouldn't we rename the module folder as well?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/configuration/OnboardingComponentsConfiguration.java
```diff
@@ -39,14 +38,6 @@
  */
 @Configuration
 @ConditionalOnProperty(name = "enrollment-server.onboarding.enabled", havingValue = "true")
-@EnableJpaRepositories(basePackages = {
```
banterCZ, line 42: And how are entities and daos from common initialized now? I mean for example `com.wultra.app.onboardingserver.common.database.OnboardingOtpRepository`?

romanstrobl:
There is:

`enrollment-server.onboarding.enabled=true`

in `enrollment-server/application-dev.properties`. We should probably also remove this configuration property.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/configuration/OnboardingComponentsConfiguration.java
banterCZ: Now I think that `OnboardingComponentsConfiguration` should be moved to downstream and property `enrollment-server.onboarding.enabled` should be removed.

enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/request/DocumentStatusRequest.java
```diff
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package com.wultra.app.enrollmentserver.api.model.request;
+package com.wultra.app.enrollmentserver.api.model.onboarding.request;
```
banterCZ, line 18: What about to introduce a new module `enrollment-server-onboarding-api-model`?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/configuration/OnboardingComponentsConfiguration.java
petrdvorak: We can discuss on grooming, we never needed those annotations.

enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/request/DocumentStatusRequest.java
petrdvorak: I was about to do it and then I thought: "I am doing the minimal fix, they will have to fix this themselves"

I admit: Big motivation for leaving it up to you is that we also have "domain model" package... which probably also should be considered.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/configuration/OnboardingComponentsConfiguration.java
banterCZ: Just for explanation: It is needed in the downstream project after the split.

enrollment-server-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/request/DocumentStatusRequest.java
banterCZ: Opps, _api-model_ depends on `domain-model`, another reason to split.

romanstrobl (approved):
I didn't notice any issues. We should run automated tests to validate these changes.

PR #263, opened 2022-06-29
Fix #262: Fix authentication of consent/text endpoint

### LINKED ISSUE #262: Fix authentication of consent/text endpoint
Fix authentication of consent/text endpoint. Remove `PowerAuthApiAuthentication` and get userId from process stored in db.

banterCZ:


romanstrobl (approved):
Yes, I overlooked this in first review.

PR #265, opened 2022-07-01
Fix #256: Implement limit for max attempts of identity verification

### LINKED ISSUE #256: Implement limit for max attempts of identity verification
When the max. attempt count for identity verification is exceeded, the onboarding process fails. The default number of attempts is 5.

romanstrobl:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/errorhandling/OnboardingProcessLimitException.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * PowerAuth Enrollment Server
+ * Copyright (C) 2021 Wultra s.r.o.
```
saalistaja, line 3: 2022

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationLimitService.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Check attempt limit of failed identity verifications. Fail onboarding process in the attempt count has exceeded the limit.
+     * @param ownerId Owner identification.
+     * @throws PowerAuthClientException Thrown when activation flag request fails.
+     * @throws IdentityVerificationException Thrown in case identity verification is invalid.
+     * @throws OnboardingProcessLimitException Thrown when maximum failed attempts for identity verification have been reached.
+     */
+    public void checkIdentityVerificationLimit(OwnerId ownerId) throws PowerAuthClientException, IdentityVerificationException, OnboardingProcessLimitException {
+        List<IdentityVerificationEntity> identityVerifications = identityVerificationRepository.findByActivationIdOrderByTimestampCreatedDesc(ownerId.getActivationId());
```
saalistaja, line 81: Maybe not clear that `findByActivationIdOrderByTimestampCreatedDesc` returns all so far failed identity verifications. The sorting is not needed here at all.

But obviously it will work
- when checking the verification limit on creation, reset

romanstrobl: It returns both `FAILED` and `IN_PROGRESS` verifications, we are interested in the total number. The sorting is used just for consistency with other methods, in case we need to get this list anywhere else, but I can remove it.

saalistaja (approved):
Looks great. I have not found any issue.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/errorhandling/OnboardingProcessLimitException.java
romanstrobl: Ah, right, copy-paste error. My template should be fine already.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationLimitService.java
saalistaja: Aaah, then `maxFailedAttempts` naming is a little bit confusing, maybe `maxAttempts`. But for sure a small comment above the line would be enough too.

romanstrobl: I will review this and add some comments to make it clear, thanks.

PR #267, opened 2022-07-04
Fix #257: Implement limit for max attempts of document upload

romanstrobl:
This PR builds on changes introduced in: https://github.com/wultra/enrollment-server/pull/266 and adds a limit for maximum number of attempts for document upload.

PR #268, opened 2022-07-07
Fix #258: Implement limit for max attempts of selfie check and OTP verification

### LINKED ISSUE #258: Implement limit for max attempts of selfie check and OTP verification
When the max. number of attempts for selfie check and/or OTP verification fails, the identity verification should fail. The default number of max attempts is 5.

romanstrobl:
This PR builds on changes introduced in: https://github.com/wultra/enrollment-server/pull/267 and adds a limit for maximum number of presence checks with OTP, verified using count of OTP entities with type = `USER_VERIFICATION`.


petrdvorak (approved):
Looks good

saalistaja (approved):
Looks good.

PR #279, opened 2022-07-18
Fix #278: Invalid property key identity-verification.data-retention

### LINKED ISSUE #278: Invalid property key identity-verification.data-retention.hours
Remove _hours_ from `@Value("${enrollment-server-onboarding.identity-verification.data-retention.hours:1h}")` because it does not match _application.properties_

banterCZ:


romanstrobl (approved):
Ah yes, thanks for noticing.

PR #300, opened 2022-07-28
Fix #299: Handle missing process ID before user lookup

### LINKED ISSUE #299: Handle missing process ID before user lookup
The lookup request requires a process ID, however the process ID is determined after lookup using obtained `user ID`. This is a chicken and egg problem.

We can resolve this problem using following approach:
- Use `identification data` to lookup the process ID for existing processes.
- Perform the user lookup with existing process ID or new process ID in case process is not found.
- Obtain `user ID` from lookup.
- Check that there is no other process for user with status `ACTIVATION_IN_PROGRESS`, in case there is such a process, fail it, because it contains obsolete `identification data`.

banterCZ:
Find the existing process by identificationData instead of userId


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -405,11 +382,49 @@ public boolean hasProcessExpired(OnboardingProcessEntity onboardingProcess) {
...
+                    .processId(processId)
+                    .build();
+            return onboardingProvider.lookupUser(lookupUserRequest).getUserId();
+        } catch (OnboardingProviderException e) {
+            throw new OnboardingProcessException("User lookup failed, error: " + e.getMessage(), e);
         }
+    }
 
-        return false;
+    @SneakyThrows(OnboardingProcessException.class)
```
banterCZ, line 400: IMHO we exaggerate it with checked exceptions. Anyway, this Lombok annotation enables to use the checked exception in lambdas. Declaration at public level remains.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -405,11 +382,49 @@ public boolean hasProcessExpired(OnboardingProcessEntity onboardingProcess) {
...
+        final String userId = lookupUser(process.getId(), identification);
+        process.setUserId(userId);
+        return process;
     }
 
+    private OnboardingProcessEntity createNewProcessWithTempUserId(final String identificationData) {
+        final OnboardingProcessEntity process = new OnboardingProcessEntity();
+        process.setIdentificationData(identificationData);
+        process.setStatus(OnboardingStatus.ACTIVATION_IN_PROGRESS);
+        process.setUserId("user-id-not-looked-up-yet");
```
banterCZ, line 413: `userId` has _not-null_ constrain. It is filled later on but still in the same transaction.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -100,43 +101,18 @@ public OnboardingServiceImpl(
...
-        } else {
-            // Create an onboarding process
-            process = new OnboardingProcessEntity();
-            process.setIdentificationData(identificationData);
-            process.setStatus(OnboardingStatus.ACTIVATION_IN_PROGRESS);
-            process.setUserId(userId);
-            process.setTimestampCreated(new Date());
-        }
-        process = onboardingProcessRepository.save(process);
+        final OnboardingProcessEntity process = onboardingProcessRepository.findExistingProcessByIdentificationData(identificationData, OnboardingStatus.ACTIVATION_IN_PROGRESS)
```
banterCZ, line 107: The change is covered by this new test case https://github.com/wultra/powerauth-tests/pull/86

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
```diff
@@ -36,10 +36,8 @@
 @Repository
 public interface OnboardingProcessRepository extends CrudRepository<OnboardingProcessEntity, String> {
 
-    @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.status = :status " +
-            "AND p.userId = :userId " +
-            "ORDER BY p.timestampCreated DESC")
-    Optional<OnboardingProcessEntity> findExistingProcessForUser(String userId, OnboardingStatus status);
+    @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.status = :status AND p.identificationData = :identificationData")
```
petrdvorak, line 39: This code may not work unless we normalize the way JSON data is serialized in the database. We should consider:

```java
ObjectMapper mapper = new ObjectMapper();
mapper.setConfig(mapper.getSerializationConfig()
   .with(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY));
   .with(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS)
```

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
petrdvorak: I don't like it... :) At least, we should introduce a new variable for it. Alternatively, we should consider loosening the constraint.

petrdvorak (changes requested):
There are two changes I think we need to make to ensure proper functionality and code style.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
petrdvorak: ... plus we should probably ensure that the other typical serialization features (formatting, date serialization, etc.) work in the same manner every time...

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
romanstrobl: I also don't like this. We should loose the constraint.

romanstrobl (approved):
I fixed the issues related to JSON normalization and initial user ID value.

petrdvorak (approved):
Thank you for the fixes.

romanstrobl:
Ok, I am merging this pull request.

PR #301, opened 2022-07-29
Fix #194: Fetch configuration endpoint

### LINKED ISSUE #194: Handle situation when onboarding is disabled on the server
If the server has onboarding disabled and client has onboarding enabled, the client would report errors. We should add endpoint that would tell the client app that the feature is not available.

If the server feature is not available, the client should inform the user about it.

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/ConfigurationController.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+@RestController
+@RequestMapping(value = "api/configuration")
+public class ConfigurationController {
+
+    @Value("${enrollment-server-onboarding.onboarding-process.enabled}")
+    private boolean onboardingEnabled;
+
+    @GetMapping
+    public Object fetchConfiguration() {
```
banterCZ, line 40: Quite free structure to extend later on. Exposed at _onboarding-server_, may be aggregated by _mobile utility server_ (if needed), that's why the signature is not required.

`GET /api/configuration`
```json
{
  "onboarding": {
    "enabled": true
  }
}
```

kober32 (approved):
OK with me (as far as an impact on frontend)

romanstrobl (approved):
Looks OK.

PR #316, opened 2022-08-22
Fix #314: Unify @Transactional annotations

### LINKED ISSUE #314: Unify @Transactional annotations
Two types of `@Transactional` annotations are used in the project: Spring and Java EE. We should use only one type and consider using `propagation = Propagation.REQUIRES_NEW` wherever applicable.

banterCZ:
Use `org.springframework.transaction.annotation.Transactional` instead of `javax.transaction.Transactional`.

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationCreateService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationCreateService.java
@@ -17,20 +17,19 @@
  */
 package com.wultra.app.onboardingserver.impl.service;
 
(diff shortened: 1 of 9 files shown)


romanstrobl (approved):
This is a step in right direction. The issue https://github.com/wultra/enrollment-server/issues/314 also suggests to `consider using propagation = Propagation.REQUIRES_NEW wherever applicable`. Do we want to do this in a separate pull request or this one?

banterCZ:
> This is a step in right direction. The issue #314 also suggests to `consider using propagation = Propagation.REQUIRES_NEW wherever applicable`. Do we want to do this in a separate pull request or this one?

Let's keep it for another PR. IMHO usage of `REQUIRES_NEW` should be limited to very special use-cases where really needed. I would rather remove this declaration that add it.

PR #321, opened 2022-08-23
Fix #320: Use fake OTP authentication in case user does not exist

### LINKED ISSUE #320: Use fake OTP authentication in case user does not exist

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -458,4 +438,51 @@ private OnboardingProcessEntity resumeExistingProcess(final OnboardingProcessEnt
...
+        try {
+            return normalizedMapper.writeValueAsString(identification);
+        } catch (JsonProcessingException ex) {
+            logger.warn("Invalid identification data: {}", identification);
+            throw new InvalidRequestObjectException();
+        }
+    }
+
+    private static boolean isFakeUser(final String userId) {
+        return userId.startsWith(FAKE_USER_ID_PREFIX);
```
banterCZ, line 452: The question is, whether such differentiation of fake user is solid enough. I guess so.

romanstrobl (approved):
👏

PR #326, opened 2022-08-25
Fix #325: OperationTemplateEntity equals uses generated id

### LINKED ISSUE #325: OperationTemplateEntity equals uses generated id

banterCZ:


romanstrobl (approved):
Looks OK, thanks for the fix.

PR #328, opened 2022-08-26
Fix #327: Change machine states without calling status endpoint

### LINKED ISSUE #327: Change machine states without calling status endpoint

banterCZ:


banterCZ:
Done in https://github.com/wultra/enrollment-server/pull/331

PR #331, opened 2022-08-29
Fix #327: Change machine states by scheduler

### LINKED ISSUE #327: Change machine states without calling status endpoint

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/database/IdentityVerificationRepository.java
```diff
@@ -70,4 +70,23 @@ public interface IdentityVerificationRepository extends CrudRepository<IdentityV
             " AND id.status = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus.IN_PROGRESS " +
             " ORDER BY id.timestampLastUpdated ASC")
     Stream<IdentityVerificationEntity> streamAllInProgressClientEvaluations();
+
+    /**
+     * Return all identity verifications eligible for change to next state.
+     *
+     * @return identity verifications
+     */
+    @Query("SELECT id FROM IdentityVerificationEntity id WHERE" +
```
banterCZ, line 79: As much phases as possible should be done out of the `StateMachineTask`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -123,6 +134,33 @@ public Message<OnboardingEvent> createMessage(OwnerId ownerId, String processId,
                 .build();
     }
 
+    /**
+     *
+     */
+    @Transactional(readOnly = true)
+    public void changeMachineStatesInBatch() {
+        try (Stream<IdentityVerificationEntity> stream = identityVerificationService.streamAllIdentityVerificationsToChangeState().parallel()) {
```
banterCZ, line 142: Do we want to use `.parallel()`?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -123,6 +134,33 @@ public Message<OnboardingEvent> createMessage(OwnerId ownerId, String processId,
...
+            stream.forEach(this::changeMachineState);
+        }
+    }
+
+    private void changeMachineState(final IdentityVerificationEntity identityVerification) {
+        final String processId = identityVerification.getProcessId();
+        final OwnerId ownerId = new OwnerId();
+        ownerId.setActivationId(identityVerification.getActivationId());
+        ownerId.setUserId(identityVerification.getUserId());
```
banterCZ, line 151: 👀 Needed to have the real `userId` from `IdentityVerificationEntity` especially for the presence check provider.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/database/IdentityVerificationRepository.java
saalistaja, line 79: Maybe we should consider a refactor of existing `DocumentsVerificationSyncTask` and `ClientEvaluationSyncTask`  tasks and cover all `status = IN_PROGRESS` here. But I see this would require adding new actions to the machine. So rather postpone.

Additionally some transitions from the `ACCEPTED` states could be maybe switched automatically by the state machine to the next state.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/StateMachineTask.java
```diff
@@ -0,0 +1,54 @@
...
+    private final StateMachineService stateMachineService;
+
+    public StateMachineTask(final StateMachineService stateMachineService) {
+        this.stateMachineService = stateMachineService;
+    }
+
+    /**
+     * Scheduled task to change machine state.
+     */
+    @Scheduled(cron = "${enrollment-server-onboarding.state-machine.changeMachineState.cron:0/1 * * * * *}", zone = "UTC")
```
saalistaja, line 47: I am aware of only simple cron patterns. Maybe there is some magic which I don't see here.

What's the frequency of this cron, every second? Is there a difference with `* * * * * *`?
Maybe we can use e.g. `*/3`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/database/IdentityVerificationRepository.java
banterCZ: I tried to limit `where` clause as much as possible. I may think about `ACCEPTED` once again.

saalistaja (approved):
Well done, looks really promising. There is no issue to address.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/database/IdentityVerificationRepository.java
```diff
@@ -70,4 +70,23 @@ public interface IdentityVerificationRepository extends CrudRepository<IdentityV
...
+            "   AND id.status = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus.IN_PROGRESS)" +
+            " OR (id.phase = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationPhase.DOCUMENT_UPLOAD" +
+            "   AND id.status = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus.VERIFICATION_PENDING)" +
+            " OR (id.phase = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationPhase.DOCUMENT_VERIFICATION" +
+            "   AND id.status = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus.ACCEPTED)" +
+            " OR (id.phase = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationPhase.CLIENT_EVALUATION" +
+            "   AND id.status = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus.ACCEPTED)" +
+            " OR (id.phase = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationPhase.PRESENCE_CHECK" +
+            "   AND id.status = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus.IN_PROGRESS)"
+    )
```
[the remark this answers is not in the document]
banterCZ: Do we need it? I mean, in that case we should have an index over updated time, right?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/database/IdentityVerificationRepository.java
banterCZ: I recalled, we need to keep them because another jobs such as `DocumentsVerificationSyncTask`. We may consolidate it in an extra PR/issue.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/StateMachineTask.java
romanstrobl: The default value seems to be inconsistent with the value set in properties file.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -123,6 +134,33 @@ public Message<OnboardingEvent> createMessage(OwnerId ownerId, String processId,
                 .build();
     }
 
+    /**
+     *
+     */
+    @Transactional(readOnly = true)
+    public void changeMachineStatesInBatch() {
+        try (Stream<IdentityVerificationEntity> stream = identityVerificationService.streamAllIdentityVerificationsToChangeState().parallel()) {
+            stream.forEach(this::changeMachineState);
```
romanstrobl, line 143: It would be nice to add the number of processed identity verifications per batch, probably in `debug` level.

romanstrobl (approved):
This looks very promising. Let's deploy the new functionality for testing.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/StateMachineTask.java
banterCZ: Ok, we may change it to three seconds, but all cron expressions in enrollment using zero, like `0/5`, which should be same.

PR #337, opened 2022-08-31
Fix #336: Remove public modifier from JUnit5 tests

### LINKED ISSUE #336: Remove public modifier from JUnit5 tests

banterCZ:


romanstrobl (approved):
Ok, why not.

PR #347, opened 2022-09-02
Fix #346: Remove @Lob and @Type

### LINKED ISSUE #346: Remove @Lob and @Type
Remove `@Lob` and `@Type` completely from String fields.
Relates to #342, #345
https://www.jpa-buddy.com/blog/how-to-store-text-in-postgresql-tips-tricks-and-traps/

banterCZ:


romanstrobl (approved):
Ok, the performance optimization was premature, agree.

PR #364, opened 2022-09-08
Fix #354: Validate that user submitted correct ID documents

### LINKED ISSUE #354: Validate that user submitted correct ID documents
Validate that user submitted either of these primary documents before starting document verification:
- ID card
- passport

The driving license is a secondary document.

Two documents of different type are required before document verification.

In case the user restarts the submission of any documents, the backend should also check that updated set of documents contains required document types.

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/IdentityVerificationController.java
```diff
@@ -319,17 +320,17 @@ public ObjectResponse<DocumentStatusResponse> checkDocumentStatus(@EncryptedRequ
...
+        checkRequestObject(request, operationDescription);
 
         final OwnerId ownerId = PowerAuthUtil.getOwnerId(apiAuthentication);
         final String processId = request.getRequestObject().getProcessId();
 
         onboardingService.verifyProcessId(ownerId, processId);
 
-        // Process upload document request
-        final DocumentStatusResponse response = identityVerificationService.checkIdentityVerificationStatus(request.getRequestObject(), ownerId);
+        final DocumentStatusResponse response = identityVerificationService.fetchDocumentStatusResponse(request.getRequestObject(), ownerId);
```
banterCZ, line 333: Just refactored confusing naming.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -175,39 +182,26 @@ public List<DocumentVerificationEntity> submitDocuments(DocumentSubmitRequest re
...
             throw new DocumentSubmitException("Invalid process ID");
         }
 
-        if (!IdentityVerificationPhase.DOCUMENT_UPLOAD.equals(idVerification.getPhase())) {
-            logger.error("The verification phase is {} but expected {}, {}",
-                    idVerification.getPhase(), IdentityVerificationPhase.DOCUMENT_UPLOAD, ownerId
-            );
+        final IdentityVerificationPhase phase = idVerification.getPhase();
+        final IdentityVerificationStatus status = idVerification.getStatus();
+        if (phase == IdentityVerificationPhase.DOCUMENT_VERIFICATION && status == IdentityVerificationStatus.IN_PROGRESS) {
```
banterCZ, line 196: We need to allow upload document later on.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -304,6 +298,11 @@ public void checkVerificationResult(IdentityVerificationPhase phase, OwnerId own
             return;
         }
 
+        if (!requiredDocumentTypesGuard.evaluate(idVerification.getDocumentVerifications(), idVerification.getId())) {
```
banterCZ, line 301: Here there is a guard to stop until all required documents are present and accepted.

banterCZ:
Works with integration test but still problem with non-mock providers.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/DocumentsVerificationPendingGuard.java
```diff
@@ -0,0 +1,74 @@
...
+        this.documentVerificationRepository = documentVerificationRepository;
+    }
+
+    @Override
+    @Transactional(readOnly = true)
+    public boolean evaluate(final StateContext<OnboardingState, OnboardingEvent> context) {
+        final IdentityVerificationEntity identityVerification = context.getExtendedState().get(ExtendedStateVariable.IDENTITY_VERIFICATION, IdentityVerificationEntity.class);
+
+        final List<DocumentVerificationEntity> documentVerifications = documentVerificationRepository.findAllUsedForVerification(identityVerification);
+        if (documentVerifications.isEmpty()) {
```
banterCZ, line 57: The guard is extracted from service and the logic for empty collections is fixed https://github.com/wultra/enrollment-server/pull/364/files#diff-0454fa519907321b8ba689699b8fc5e5d8835ea5172e7d29137e24ef4e9eb0baL491

petrdvorak (approved):
Looks OK

romanstrobl (changes requested):
As discussed, for starting the verification we should check the following conditions:
- two distinct document types are present (ID card - both sides, passport, driving license)
  - a filter is applied to only take into account most recent versions of documents (`usedForVerification` = `true`)
- at least 1 primary document is present (ID card or passport) - there is no other way currently, but we can check this as a double-check
- at least 1 of the documents is in state `VERIFICATION_PENDING`, the other one is either in state `VERIFICATION_PENDING` or `ACCEPTED`

Consequences (to double check the logic or for tests):
- if no document is present, the verification does not start, it waits for both documents
- if only one document is present in state `VERIFICATION_PENDING`, the verification does not start yet and waits for second document
- if more than two documents are present, this state is invalid and all document verifications should fail immediately
- if both documents are in state `ACCEPTED`, the verification does not start, because it was already executed
- if any of the required documents is in state `DISPOSED`, `REJECTED` or `FAILED`, a re-upload of the document is required so that the verification can start
- in case only half of the document is present (e.g. one side of an ID card), the verification should wait for the second half to appear, otherwise the verification would be incomplete

Note that this logic may become dynamic in the future (e.g. configured by a DB table), but for now static logic is fine.

banterCZ:
> in case only half of the document is present (e.g. one side of an ID card), the verification should wait for the second half to appear, otherwise the verification would be incomplete

I have opened a separate issue for this #371

banterCZ:
Successfully tested end-to-end on dev env.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesGuard.java
```diff
@@ -0,0 +1,75 @@
...
+     * @return true when all required document types present and accepted
+     */
+    public boolean evaluate(final Collection<DocumentVerificationEntity> documentVerifications, final String identityVerificationId) {
+        if (documentVerifications.isEmpty()) {
+            logger.debug("There is no document uploaded yet for identity verification ID: {}", identityVerificationId);
+            return false;
+        } else if (!containsIdOrPassport(documentVerifications)) {
+            logger.debug("There is no accepted ID card or travel passport yet for identity verification ID: {}", identityVerificationId);
+            return false;
+        } else if (!containsDrivingLicence(documentVerifications)) {
```
romanstrobl, line 53: This seems incorrect, because such condition does not allow ID card and passport as submitted documents, which I think should pass.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/DocumentsVerificationPendingGuard.java
```diff
@@ -0,0 +1,74 @@
...
+            logger.debug("No document uploaded yet for {}", identityVerification);
+            return false;
+        }
+
+        final boolean allDocumentsPendingVerification = documentVerifications.stream()
```
romanstrobl, line 62: We should improve this check, although it is almost correct, the expected behaviour is that exactly 2 documents are available (not 3 or more, this would be allowed at the moment as far as I can tell).

romanstrobl (changes requested):
The change looks really good, we just need to properly cover the corner cases.

romanstrobl:
As discussed, the current logic also does not check that submitted document types are distinct (e.g. 2 submitted passports should not be accepted).

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesGuard.java
```diff
@@ -44,31 +46,58 @@ public class RequiredDocumentTypesGuard {
...
+
+    private static boolean containsPassport(final Collection<DocumentVerificationEntity> documentVerifications) {
         return documentVerifications.stream()
-                .filter(it -> it.getStatus() == DocumentStatus.ACCEPTED)
                 .map(DocumentVerificationEntity::getType)
-                .anyMatch(it -> it == DocumentType.ID_CARD || it == DocumentType.PASSPORT);
+                .anyMatch(it -> it == DocumentType.PASSPORT);
+    }
+
+    private static boolean containsSecondaryDocument(final Collection<DocumentVerificationEntity> documentVerifications) {
```
romanstrobl, line 93: I do not like this method name.

A primary document is: ID card, passport.
A secondary document is: driving license.

The method name should be e.g. `containsSecondDocument` or `containsAnotherDocument`, not `secondary`, this word has a special meaning.

romanstrobl (changes requested):
This looks great, there is just one minor issue with name of a method which is misleading.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesGuard.java
banterCZ: Renamed to `containsSecondDocument`.

romanstrobl (approved):
I am happy now.

PR #367, opened 2022-09-12
Fix #366: Move logic from ClientEvaluationSyncTask to state machine

### LINKED ISSUE #366: Move logic from ClientEvaluationSyncTask to state machine
Some logic is done in tasks, some in the state machine. It is quite complicated. Originally it was needed because of changing state by `status` endpoint.

Move the code from task to state machine action.

banterCZ:


petrdvorak (approved):
Looks OK.

romanstrobl (approved):
Nice cleanup.

PR #374, opened 2022-09-19
Fix #373: Use response object for onboarding configuration

### LINKED ISSUE #373: Use response object for onboarding configuration
The endpoint `/api/configuration` does not follow the same response conventions as the rest of the API.

romanstrobl:


kober32 (approved):
Not implemented on the frontend yet, but this is a "more expected" format, so OK with me.

banterCZ (approved):
Thanks for adjusting according to the conventions.

Issue #375, opened 2022-09-19
Change default expiration time for OTP


romanstrobl:
The default OTP timeout of 30 seconds makes sense for development, however it is too short for production use.

PR #379, opened 2022-09-20
Fix CleaningServiceTest after change of otp expiration in #375

banterCZ:


romanstrobl (approved):
My bad, thanks for the fix.

PR #381, opened 2022-09-20
Fix #380: Remove double quotes from sequence definition

### LINKED ISSUE #380: Remove double quotes from sequence definition
Quotes in DDL definition enable case sensitivity, there is no need for it: https://github.com/wultra/enrollment-server/blob/develop/docs/sql/postgresql/onboarding/create-schema.sql#L23

romanstrobl:


banterCZ (approved):
A good catch. I thought we had already got rid of all quotes.

PR #386, opened 2022-09-22
Error handling from adapter and related refactoring

### LINKED ISSUE #272: Add blacklisted ID info to response from client evaluation service
Information about blacklisted ID should be added into the REST API response from client evaluation service and such event should be processed on the backend (special `errorDetail` saved into DB, updated identity verification state).

The change has impact on the REST API of Digital Onboarding Adapter.

### LINKED ISSUE #384: Improve handling of error responses in User Lookup
In case User Lookup fails, we should process the reason why User Lookup failed. Ideally the reason why lookup failed should be logged and stored in the database in column `error_detail`.

romanstrobl:
The pull request contains following changes:
- Fix #272
- Fix #384
- Set timestampFinished for identity verification at the end of process 
- Model package cleanup for REST Onboarding Provider

The concrete part of error handling will be handled in downstream (mapping of error codes to error detail).

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
@@ -267,7 +267,7 @@ public class PresenceCheckService {
             case ACCEPTED:
                 idVerification.setStatus(IdentityVerificationStatus.ACCEPTED);
                 idVerification.setTimestampLastUpdated(ownerId.getTimestamp());
-                // The timestampFinished parameter is not set yet, there are may be other steps ahead
+                // The timestampFinished parameter is not set yet, there may be other steps ahead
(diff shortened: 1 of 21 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
```diff
@@ -267,7 +267,7 @@ private void evaluatePresenceCheckResult(OwnerId ownerId,
             case ACCEPTED:
                 idVerification.setStatus(IdentityVerificationStatus.ACCEPTED);
                 idVerification.setTimestampLastUpdated(ownerId.getTimestamp());
-                idVerification.setTimestampFinished(ownerId.getTimestamp());
+                // The timestampFinished parameter is not set yet, there are may be other steps ahead
```
banterCZ, line 270: The verb _are_ appears to be unnecessary.

banterCZ (approved):
Looks good.

PR #392, opened 2022-09-23
Fix #391: Disallow approval of operations with onboarding flags

### LINKED ISSUE #391: Disallow approval of operations with onboarding flags
Disallow approval of operations with onboarding flags in vanilla enrollment server.

romanstrobl:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java
```diff
@@ -62,6 +63,9 @@ public class MobileTokenController {
 
     private static final Logger logger = LoggerFactory.getLogger(MobileTokenController.class);
 
+    // Disallowed flags contain onboarding flags used before onboarding process is finished
+    private static final List<String> DISALLOWED_FLAGS = Arrays.asList("VERIFICATION_PENDING", "VERIFICATION_IN_PROGRESS");
```
banterCZ, line 67: You may use `List#of`, as bonus, you get immutable collection.

banterCZ (approved):

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java
romanstrobl: Right, we moved away from Java 8 support, I forgot.

PR #402, opened 2022-09-27
Fix #400: Improve logging

### LINKED ISSUE #400: Improve logging
Prefer OwnerId in the log context rather than processId.

banterCZ:
- replace processId by ownerId in log messages changing state
- pass message to the exception instead of logging it right away
- use idiom orElseThrow for optional


romanstrobl (approved):
Looks good.

PR #403, opened 2022-09-29
Fix #351: Service for automatic cleanup of activations with failed onboarding

### LINKED ISSUE #351: Service for automatic cleanup of activations with failed onboarding
An activation cleanup service will be implemented which will remove all activations for which the customer identification process was not finished within specified timeframe (default: 3 hours). In case the onboarding process failed after activation, the activation status should be checked on PowerAuth server and we should mark that the activation was removed, too, perhaps using the database to avoid any future checks.

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/ActivationCleaningService.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Service to cleaning activations.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@Service
+@Slf4j
+class ActivationCleaningService {
+
+    private static final int BATCH_SIZE = 100;
```
banterCZ, line 40: This number is just an initial guess.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/ActivationCleaningService.java
```diff
@@ -0,0 +1,90 @@
...
+            final ActivationService activationService) {
+
+        this.onboardingProcessRepository = onboardingProcessRepository;
+        this.activationService = activationService;
+    }
+
+    /**
+     * Cleanup activations of failed onboarding processes.
+     */
+    public void cleanupActivations() {
```
banterCZ, line 58: Missing `@Transactional` on purpose. It is possible to load processes anyway. We do not want to do REST calls in transaction. Each process is updated in a separate transaction, because `save` in Spring Data is annotated by default.

romanstrobl (approved):
I really like the implementation. The only change I would introduce is to handle all activations for removal per one scheduled run instead of limiting them by `BATCH_SIZE` to avoid some corner cases where many activations are created (e.g. perf tests) when the cleanup would take quite long.

PR #410, opened 2022-10-03
Fix #351: ActivationCleaningService needs transaction for stream processing

### LINKED ISSUE #351: Service for automatic cleanup of activations with failed onboarding

banterCZ:


romanstrobl (approved):
This should work fine.

Issue #352, opened 2022-09-02
Auditing of important events using auditing library


romanstrobl:
See: https://github.com/wultra/lime-java-core#wultra-auditing-library

romanstrobl:
Please include audit records related to activation, too, including expiration and removal.

banterCZ:
Already implemented.

PR #416, opened 2022-10-05
Add Audit library dependency and its configuration

banterCZ:
Issue #352


romanstrobl (approved):
Implemented perfectly according to documentation.

PR #417, opened 2022-10-05
Add initial github action

banterCZ:


romanstrobl (changes requested):
Can we synchronize this with https://github.com/wultra/powerauth-crypto/pull/414/files ?

banterCZ:
I have renamed the file and add `workflow_dispatch`. I am not sure about publish. May we postpone it to another PR?

romanstrobl:
We will probably create a reusable workflow we can use across all PowerAuth backends, so let's postpone this pull request until this is ready.

romanstrobl:
See: https://github.com/wultra/lime-java-core/pull/116/files

romanstrobl (approved):
I am satisfied with the changes now.

PR #419, opened 2022-10-05
Fix #418: Allow serialization of ActivationOtpErrorResponse

### LINKED ISSUE #418: Allow serialization of ActivationOtpErrorResponse
The class is missing a default constructor and the field is final, thus it cannot be used by ObjectMapper in tests.

romanstrobl:
Unfortunately we need to modify the model class for test usage due to ObjectMapper requirements.

### DIFF
--- a/enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/error/ActivationOtpErrorResponse.java
+++ b/enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/error/ActivationOtpErrorResponse.java
@@ -28,7 +28,14 @@ import javax.validation.constraints.NotBlank;
  */
 public class ActivationOtpErrorResponse extends ErrorResponse {
 
-    private final Integer remainingAttempts;
+    private Integer remainingAttempts;
+
+    /**
+     * Default constructor.
+     */
+    public ActivationOtpErrorResponse() {
+        super();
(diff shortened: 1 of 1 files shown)


enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/error/ActivationOtpErrorResponse.java
```diff
@@ -28,7 +28,14 @@
...
+
+    /**
+     * Default constructor.
+     */
+    public ActivationOtpErrorResponse() {
+        super();
+    }
 
     /**
      * Create a new error response with response object with provided code, error message and remaining attempts.
```
banterCZ, line 41: You may try `@JsonCreator` annotation,  if you want to. https://www.tutorialspoint.com/jackson_annotations/jackson_annotations_jsoncreator.htm

banterCZ (approved):

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/error/ActivationOtpErrorResponse.java
romanstrobl: We prefer not to use Jackson-specific annotations because someone may use another serialization framework.

PR #423, opened 2022-10-06
Audit Document

banterCZ:
Issue #352

### DIFF
--- a/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/AuditService.java
+++ b/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/AuditService.java
@@ -17,6 +17,7 @@
  */
 package com.wultra.app.onboardingserver.common.service;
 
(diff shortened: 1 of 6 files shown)


enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/AuditService.java
```diff
@@ -32,6 +33,12 @@
 @Service
 public class AuditService {
 
+    private static final String IDENTITY_VERIFICATION_ID = "identityVerificationId";
```
romanstrobl, line 36: I see this declaration for the third time in the review. Perhaps we can share these constants?

romanstrobl (approved):
Looks good. I have one suggestion related to shared constant declaration.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/AuditService.java
banterCZ: Sorry, I did the same change in several PRs, but yes, it is a shared constant.

PR #426, opened 2022-10-07
Audit PresenceCheckProvider

banterCZ:
Issue #352

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
@@ -21,14 +21,17 @@ import com.google.common.base.Ascii;
 import com.google.common.base.Preconditions;
 import com.wultra.app.enrollmentserver.model.enumeration.*;
 import com.wultra.app.enrollmentserver.model.integration.*;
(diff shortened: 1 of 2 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
```diff
@@ -129,6 +137,7 @@ public SessionInfo init(OwnerId ownerId, String processId)
             } else {
                 Image photo = selectPhotoForPresenceCheck(ownerId, docsWithPhoto);
                 presenceCheckProvider.initPresenceCheck(ownerId, photo);
+                auditService.auditPresenceCheckProvider(idVerification, "Presence check initialized for user: {}", ownerId.getUserId());
```
banterCZ, line 140: This was tough architecture decision, but at the end, I believe that abstraction level is correct. Auditing should be done where the provider interface is called. Auditing must be same for whatever implementation, not only _iProov_. Of course, we may improve logging in `IProovPresenceCheckProvider` or `IProovRestApiService`, but if some details are important for, they should be propagated higher.

romanstrobl: Agree.

romanstrobl (approved):
Looks good.

PR #427, opened 2022-10-07
Audit DocumentVerificationProvider

banterCZ:
Issue #352

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentProcessingService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentProcessingService.java
@@ -202,6 +202,8 @@ public class DocumentProcessingService {
         DocumentSubmitResult docSubmitResult;
         try {
             docsSubmitResults = documentVerificationProvider.checkDocumentUpload(ownerId, docVerification);
+            final IdentityVerificationEntity identityVerification = documentResultEntity.getDocumentVerification().getIdentityVerification();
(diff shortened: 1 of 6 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentProcessingService.java
```diff
@@ -202,6 +202,8 @@ public void checkDocumentSubmitWithProvider(OwnerId ownerId, DocumentResultEntit
         DocumentSubmitResult docSubmitResult;
         try {
             docsSubmitResults = documentVerificationProvider.checkDocumentUpload(ownerId, docVerification);
+            final IdentityVerificationEntity identityVerification = documentResultEntity.getDocumentVerification().getIdentityVerification();
+            auditService.auditDocumentVerificationProvider(identityVerification, "Check document upload for user: {}", ownerId.getUserId());
```
banterCZ, line 206: This was tough architecture decision, but at the end, I believe that abstraction level is correct. Auditing should be done where the provider interface is called. Auditing must be same for whatever implementation, not only _ZenId_. Of course, we may improve logging in `ZenidDocumentVerificationProvider` or `ZenidRestApiService`, but if some details are important for, they should be propagated higher. Similar to #426

romanstrobl: Agree.

romanstrobl (approved):
Looks good.

PR #429, opened 2022-10-10
Fix #428: Remove explicit lombok version

### LINKED ISSUE #428: Remove explicit lombok version

banterCZ:


pom.xml
```diff
@@ -146,7 +146,6 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>1.18.24</version>
```
banterCZ, line 149: This version is driven by `spring-boot-dependencies`. Let's do not override it if there is no need for it. When we update Spring Boot, similar issues keep us behind. For temporary updates, use predefined `${lombok.version}` and comment the reason of the change.

romanstrobl (approved):
Good idea.

PR #436, opened 2022-10-11
Fix #413: A separate error code when ZenId is down

### LINKED ISSUE #413: A separate error code when ZenId is down
When ZenId is down and returning 503, we are getting `DOCUMENT_VERIFICATION_FAILED` which is not clear enough. Let's introduce a separate error code.

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/DocumentVerificationProvider.java
```diff
@@ -105,8 +112,9 @@ public interface DocumentVerificationProvider {
      * @param id Owner identification.
      * @param initAttributes Initialization attributes
      * @return Info with data related to the verification process in SDK.
-     * @throws DocumentVerificationException When an error during initializing the verification SDK occurred
+     * @throws RemoteCommunicationException In case of remote communication error
+     * @throws DocumentVerificationException In case of business logic error
      */
-    VerificationSdkInfo initVerificationSdk(OwnerId id, Map<String, String> initAttributes) throws DocumentVerificationException;
+    VerificationSdkInfo initVerificationSdk(OwnerId id, Map<String, String> initAttributes) throws RemoteCommunicationException, DocumentVerificationException;
```
banterCZ, line 118: The main change of this PR is to distinguish business logic error from communication error.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
```diff
@@ -91,30 +89,30 @@ public ZenidDocumentVerificationProvider(
...
         DocumentsSubmitResult result = new DocumentsSubmitResult();
         ResponseEntity<ZenidWebUploadSampleResponse> responseEntity;
 
         try {
             responseEntity = zenidApiService.syncSample(document.getUploadId());
         } catch (RestClientException e) {
-            logger.warn("Failed REST call to check {} upload in ZenID, statusCode={}, responseBody='{}', {}",
-                    document, e.getStatusCode(), e.getResponse(), id, e);
-            throw new DocumentVerificationException("Unable to check document upload due to a REST call failure");
+            throw new RemoteCommunicationException(
```
banterCZ, line 99: REST error leads to 503 REMOTE_COMMUNICATION_ERROR.
In addition, double logging is limited and the whole context is passed to the exception.

kober32:
Do I understand correctly that there will be a new possible 503 for the `init-sdk` call and that's it?

banterCZ:
> Do I understand correctly that there will be a new possible 503 for the `init-sdk` call and that's it?

I believe that it is possible even for `document/submit`.

kober32 (approved):
OK with me

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
```diff
@@ -497,13 +483,13 @@ private DocumentsVerificationResult toResult(OwnerId id, ZenidWebInvestigateResp
             List<DocumentVerificationResult> verificationResults = new ArrayList<>();
 
             String extractedData = toExtractedData(id, response.getMinedData());
-            for (String sampleId : sampleIdsValidations.keySet()) {
+            for (var entry : sampleIdsValidations.entrySet()) {
```
romanstrobl, line 486: Is there any benefit to using var? In my opinion this usage hides the type and makes code less portable, so unless we plan to use var everywhere, I would not use it.

romanstrobl (approved):
Looks OK, I have just one concern related to usage of `var` keyword.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/docverify/zenid/provider/ZenidDocumentVerificationProvider.java
banterCZ: We are changing `keySet` to `entrySet`, so it should be `Map.Entry<String, List<ZenidWebInvestigationValidatorResponse>>` which is less readable than `var`.

romanstrobl: Ok.

PR #439, opened 2022-10-12
Fix #414: A separate error code when iProov is down

### LINKED ISSUE #414: A separate error code when iProov is down
When iProov is down and returning 503, we should get a clue clear enough. Let's introduce a separate error code.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #442, opened 2022-10-12
Fix #415: Implement locking mechanism for DB rows in enrollment-onboarding

### LINKED ISSUE #415: Implement locking mechanism for DB rows in enrollment-onboarding
We should identify potential conflicts when DB rows get updated by various processes in enrollment-server-onboarding and lock DB rows in case there are potential conflicts in the logic which updates the data (onboarding processes, identity verifications, OTPs, and documents - wherever the status can change).

romanstrobl:
This pull request brings two data integrity improvements:

- Decrease probability of an accidental data overwrite due to scheduling of jobs which work on the same database tables at different time.
- Fix #415

The locking should eliminate main reasons for data overwrites:
- App vs. backend jobs - process DB rows are locked both in frontend and backend services.
- State machine vs. cleaning jobs - the schedule is no longer conflicting for most of the jobs, furthermore DB rows are locked.

We cannot lock data for the whole processing in following cases which continue to run in separate transactions as defined using `Propagation.REQUIRES_NEW`:
- state machine - data is read, then written later on and the lock is applied for the state transition
- document verification tasks - communication with the provider may take long, lock is applied after results are available

I will analyze these two scenarios to check if there are any pending issues, but we should be in much better shape than when no locking was done at all.

Performance is out of scope of this work right now, the main problem is with the state machine `ProcessIdentifierGuard`. This will need to be addressed, probably separately, ideally with some tests which measure the performance.

Note that the scheduling change does not try to eliminate all possible conflicts because this would mean basically using just 1 global lock which would be quite bad for performance. However this change should decrease significantly probability of a conflict on same data. Rare conflicts are not a problem because beside the scheduling change the DB row locks are applied, too. So current implementation is a compromise between performance and conflict avoidance.

### DIFF
--- a/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
+++ b/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
@@ -57,6 +57,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
      * Find an existing process using identification data and status. Lock this process using PESSIMISTIC_WRITE lock
      * until the end of the transaction.
      *
+     * @see #findByActivationIdAndStatus(String, OnboardingStatus) (String, OnboardingStatus) - alternative method without lock
      * @param identificationData Identification data.
      * @param status Process status.
      * @return Optional onboarding process.
@@ -68,6 +69,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
     /**
      * Find an existing process by activation identifier and process status. The process is not locked.
      *
+     * @see #findByActivationIdAndStatusWithLock(String, OnboardingStatus) - alternative method with lock
      * @param activationId Activation identifier.
      * @param status Onboarding process status.
      * @return Optional onboarding process.
--- a/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
+++ b/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
@@ -51,7 +51,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
      */
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.id = :processId")
-    Optional<OnboardingProcessEntity> findProcessByIdWithLock(String processId);
+    Optional<OnboardingProcessEntity> findByIdWithLock(String processId);
 
     /**
      * Find an existing process using identification data and status. Lock this process using PESSIMISTIC_WRITE lock
@@ -63,7 +63,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
      */
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.status = :status AND p.identificationData = :identificationData")
-    Optional<OnboardingProcessEntity> findExistingProcessWithLock(String identificationData, OnboardingStatus status);
+    Optional<OnboardingProcessEntity> findByIdentificationDataAndStatusWithLock(String identificationData, OnboardingStatus status);
 
     /**
      * Find an existing process by activation identifier and process status. The process is not locked.
@@ -75,11 +75,11 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
     @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.status = :status " +
             "AND p.activationId = :activationId " +
             "ORDER BY p.timestampCreated DESC")
-    Optional<OnboardingProcessEntity> findExistingProcessForActivation(String activationId, OnboardingStatus status);
+    Optional<OnboardingProcessEntity> findByActivationIdAndStatus(String activationId, OnboardingStatus status);
 
     /**
-     * Find an existing process by activation identifier and process status. Lock the onboarding process until the
-     * end of the transaction.
+     * Find an existing process by activation identifier and process status. Lock this process using PESSIMISTIC_WRITE
+     * lock until the end of the transaction.
      *
      * @param activationId Activation identifier.
      * @param status Onboarding process status.
@@ -89,7 +89,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
     @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.status = :status " +
             "AND p.activationId = :activationId " +
             "ORDER BY p.timestampCreated DESC")
-    Optional<OnboardingProcessEntity> findExistingProcessForActivationWithLock(String activationId, OnboardingStatus status);
+    Optional<OnboardingProcessEntity> findByActivationIdAndStatusWithLock(String activationId, OnboardingStatus status);
 
     /**
      * Find process by activation identifier. Do not lock the process.
@@ -100,7 +100,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
     @Query("SELECT p FROM OnboardingProcessEntity p " +
             "WHERE p.activationId = :activationId " +
             "ORDER BY p.timestampCreated DESC")
-    Optional<OnboardingProcessEntity> findProcessByActivationId(String activationId);
+    Optional<OnboardingProcessEntity> findByActivationId(String activationId);
 
     /**
      * Count onboarding processes for user with timestamp created after given timestamp.
@@ -110,7 +110,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
      * @return Count of onboarding processes satisfying query criteria.
      */
     @Query("SELECT count(p) FROM OnboardingProcessEntity p WHERE p.userId = :userId AND p.timestampCreated > :dateAfter")
-    int countProcessesAfterTimestamp(String userId, Date dateAfter);
+    int countByUserIdAndTimestamp(String userId, Date dateAfter);
 
     /**
      * Return onboarding process IDs by the given timestamp and status. Lock these processes using PESSIMISTIC_WRITE
@@ -124,7 +124,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
     @Query("SELECT p.id FROM OnboardingProcessEntity p " +
             "WHERE p.status = :status " +
             "AND p.timestampCreated < :dateCreatedBefore")
-    List<String> findExpiredProcessIdsByStatusWithLock(Date dateCreatedBefore, OnboardingStatus status);
+    List<String> findByTimestampAndStatusWithLock(Date dateCreatedBefore, OnboardingStatus status);
 
     /**
      * Return onboarding process IDs by the given timestamp. Include only not yet finished entities. Lock these
@@ -138,7 +138,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
             "WHERE p.status <> com.wultra.app.enrollmentserver.model.enumeration.OnboardingStatus.FINISHED " +
             "AND p.status <> com.wultra.app.enrollmentserver.model.enumeration.OnboardingStatus.FAILED " +
             "AND p.timestampCreated < :dateCreatedBefore")
-    List<String> findExpiredProcessIdsWithLock(Date dateCreatedBefore);
+    List<String> findActiveByTimestampWithLock(Date dateCreatedBefore);
 
     /**
      * Return onboarding processes to remove activation. Lock these processes using PESSIMISTIC_WRITE lock until
@@ -151,7 +151,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
             "WHERE p.status = com.wultra.app.enrollmentserver.model.enumeration.OnboardingStatus.FAILED " +
             "AND p.activationId IS NOT NULL " +
             "AND p.activationRemoved = false")
-    Stream<OnboardingProcessEntity> findProcessesToRemoveActivationWithLock();
+    Stream<OnboardingProcessEntity> findAllToRemoveActivationWithLock();
 
     /**
      * Mark the given onboarding processes as failed.
--- a/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
+++ b/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
@@ -92,6 +92,17 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
     Optional<OnboardingProcessEntity> findExistingProcessForActivationWithLock(String activationId, OnboardingStatus status);
 
     /**
+     * Find process by activation identifier. Do not lock the process.
+     *
+     * @param activationId Activation identifier.
+     * @return Optional onboarding process.
+     */
+    @Query("SELECT p FROM OnboardingProcessEntity p " +
+            "WHERE p.activationId = :activationId " +
+            "ORDER BY p.timestampCreated DESC")
+    Optional<OnboardingProcessEntity> findProcessByActivationId(String activationId);
+
+    /**
      * Count onboarding processes for user with timestamp created after given timestamp.
      *
      * @param userId User identifier.
--- a/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
+++ b/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
@@ -50,7 +50,7 @@ public interface OnboardingProcessRepository extends CrudRepository<OnboardingPr
      * @return Optional onboarding process.
      */
     @Lock(LockModeType.PESSIMISTIC_WRITE)
-    @Query("SELECT p FROM OnboardingOtpEntity p WHERE p.id = :processId")
+    @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.id = :processId")
     Optional<OnboardingProcessEntity> findProcessByIdWithLock(String processId);
 
     /**
--- a/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
+++ b/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
@@ -21,11 +21,13 @@ package com.wultra.app.onboardingserver.common.database;
 import com.wultra.app.enrollmentserver.model.enumeration.ErrorOrigin;
 import com.wultra.app.enrollmentserver.model.enumeration.OnboardingStatus;
 import com.wultra.app.onboardingserver.common.database.entity.OnboardingProcessEntity;
+import org.springframework.data.jpa.repository.Lock;
 import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.stereotype.Repository;
 
+import javax.persistence.LockModeType;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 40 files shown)


enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
```diff
@@ -40,72 +42,133 @@
...
+     * until the end of the transaction.
+     *
+     * @param identificationData Identification data.
+     * @param status Process status.
+     * @return Optional onboarding process.
+     */
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
     @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.status = :status AND p.identificationData = :identificationData")
-    Optional<OnboardingProcessEntity> findExistingProcessByIdentificationData(String identificationData, OnboardingStatus status);
+    Optional<OnboardingProcessEntity> findExistingProcessWithLock(String identificationData, OnboardingStatus status);
```
banterCZ, line 66: Just not sure about naming (I know, it's not the crucial part of this PR) , what about

- `findProcessByIdWithLock` -> `findByIdWithLock` (we are in process repository, so process is obvious)
- `findExistingProcessWithLock` -> `findByIdentificationDataWithLock` (it's obvious that we are looking for an existing process)

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
```diff
@@ -40,72 +42,133 @@
...
      *
      * @return onboarding processes
      */
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
     @Query("SELECT p FROM OnboardingProcessEntity p " +
             "WHERE p.status = com.wultra.app.enrollmentserver.model.enumeration.OnboardingStatus.FAILED " +
             "AND p.activationId IS NOT NULL " +
             "AND p.activationRemoved = false")
-    Stream<OnboardingProcessEntity> findProcessesToRemoveActivation();
+    Stream<OnboardingProcessEntity> findProcessesToRemoveActivationWithLock();
```
banterCZ, line 154: 🤔 Do you know how does lock work for streams? Are all entries locker or just one at the time?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/IdentityVerificationController.java
```diff
@@ -169,6 +174,9 @@ public ResponseEntity<Response> initializeIdentityVerification(@EncryptedRequest
         final OwnerId ownerId = PowerAuthUtil.getOwnerId(apiAuthentication);
         final String processId = request.getRequestObject().getProcessId();
 
+        // Lock onboarding process
```
banterCZ, line 177: I suggest comment reasons, not what is going on. This is obvious, since the method `verifyProcessIdAndLock` is called. Or if you want to, let's log e.g. `logger.debug("Locking process ID: {}", processId);`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/IdentityVerificationController.java
```diff
@@ -367,7 +380,8 @@ public ObjectResponse<DocumentVerificationSdkInitResponse> initVerificationSdk(
         final OwnerId ownerId = PowerAuthUtil.getOwnerId(apiAuthentication);
         final String processId = request.getRequestObject().getProcessId();
 
-        onboardingService.verifyProcessId(ownerId, processId);
+        // Onboarding process is not locked
```
banterCZ, line 383: What about to transfer these comments to DEBUG or TRACE logger?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
```diff
@@ -141,8 +136,7 @@ private Consumer<EvaluateClientResponse> createSuccessConsumer(final IdentityVer
             final IdentityVerificationPhase phase = identityVerification.getPhase();
             if (response.isAccepted()) {
                 logger.info("Client evaluation accepted for {}", identityVerification);
-                saveInANewTransaction(status ->
-                        identityVerificationService.moveToPhaseAndStatus(identityVerification, phase, ACCEPTED, ownerId));
+                identityVerificationService.moveToPhaseAndStatus(identityVerification, phase, ACCEPTED, ownerId);
```
banterCZ, line 139: Why not, but does it work? The reason for this mojo was, that the transaction processing this batch used to be `read-only`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -247,14 +250,32 @@ public Response performCleanup(OnboardingCleanupRequest request) throws Onboardi
...
+     * @param onboardingStatus Expected onboarding process status.
+     * @throws OnboardingProcessException Thrown in case process identifier is invalid.
+     */
+    public void verifyProcessIdAndLock(OwnerId ownerId, String processId, OnboardingStatus onboardingStatus) throws OnboardingProcessException {
+        final OnboardingProcessEntity process = onboardingProcessRepository.findExistingProcessForActivationWithLock(ownerId.getActivationId(), onboardingStatus)
+                .orElseThrow(() -> new OnboardingProcessException("Onboarding process not found, activation ID: " + ownerId.getActivationId()));
+        String expectedProcessId = process.getId();
+
+        if (!expectedProcessId.equals(processId)) {
+            logger.warn("Invalid process ID received in request: {}, {}", processId, ownerId);
```
banterCZ, line 266: Let's use exceptions with context and transform logger to:
```suggestion
            throw new OnboardingProcessException(String.format("Invalid process ID received in request: %s, %s", processId, ownerId));
```

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/DocumentsVerificationSyncTask.java
```diff
@@ -40,7 +40,7 @@ public DocumentsVerificationSyncTask(VerificationProcessingBatchService verifica
      * Scheduled task to check documents verifications at the target provider
      */
     @Scheduled(cron = "${enrollment-server-onboarding.document-verification.checkDocumentsVerifications.cron:0/5 * * * * *}", zone = "UTC")
-    @SchedulerLock(name = "checkDocumentsVerifications", lockAtLeastFor = "1s", lockAtMostFor = "5m")
+    @SchedulerLock(name = "documentVerificationLock", lockAtLeastFor = "1s", lockAtMostFor = "5m")
```
banterCZ, line 43: I am not a big fan of constants, but thinking about that, because this literal is important and used at several places.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/OtpTransitionsTest.java
```diff
@@ -64,12 +66,10 @@ class OtpTransitionsTest extends AbstractStateMachineTest {
     private VerificationProcessResultAction verificationProcessResultAction;
 
     @Test
+    @Transactional
```
banterCZ, line 69: You may annotate the whole test class.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
romanstrobl: Usage of streams should not matter, the query still translates to `SELECT ... FOR UPDATE` which will lock all entries at once. I will double check this by tracing Hibernate log messages.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/DocumentsVerificationSyncTask.java
romanstrobl: Good idea.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
romanstrobl: Ok.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
romanstrobl: I will double check this, thanks for noticing.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/CommonOnboardingService.java
```diff
@@ -62,28 +62,26 @@ public OnboardingProcessEntity findProcess(String processId) throws OnboardingPr
...
-        return processOptional.get();
+    public OnboardingProcessEntity findProcessWithLock(String processId) throws OnboardingProcessException {
+        return onboardingProcessRepository.findProcessByIdWithLock(processId).orElseThrow(() ->
+                new OnboardingProcessException("Onboarding process not found, process ID: " + processId));
     }
 
     @Override
     public String findUserIdByProcessId(final String processId) throws OnboardingProcessException {
-        return findProcess(processId).getUserId();
+        // The onboarding process is locked until the end of the transaction
```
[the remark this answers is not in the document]
romanstrobl: Sure.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/IdentityVerificationController.java
romanstrobl: Yes, I will switch to debug logging. The reason for the comments is that I wanted to make sure in each controller it is clear whether locking is on or off.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/IdentityVerificationController.java
romanstrobl: I will move them.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/CommonOnboardingService.java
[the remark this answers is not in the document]
romanstrobl: Yes, I will move it.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/IdentityVerificationController.java
romanstrobl: Sure, I will use logging.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -260,7 +260,9 @@ private void configureDocumentVerificationTransitions(StateMachineTransitionConf
                 .source(OnboardingState.DOCUMENT_UPLOAD_VERIFICATION_PENDING)
                 .event(OnboardingEvent.EVENT_NEXT_STATE)
                 .action(verificationDocumentStartAction)
-                .guard(documentsVerificationPendingGuard)
+                .guard(
```
banterCZ, line 263: May be improved after https://github.com/wultra/enrollment-server/pull/444

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
romanstrobl: It should work, because the client evaluation action is triggered by state machine which starts a transaction on background using:

```
                transactionTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
                transactionTemplate.executeWithoutResult(status -> { ... });
```

Or from REST API call using:
```
    @Transactional
    public StateMachine<OnboardingState, OnboardingEvent> processStateMachineEvent(OwnerId ownerId, String processId, OnboardingEvent event) { ... }
```

We can review if we need additional transactions in the future, however such behaviour goes unfortunately against the locks which span only a single transaction.

romanstrobl:
All tests pass including when run in parallel with similar or slightly better execution time than in `develop` branch. There is some leftover work for the future (e.g. improved locking in state machine), but I think this can be done later.

petrdvorak (approved):
Looks OK

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
banterCZ: Thanks for the verification. I have realized that most probably it is a relic of client evaluation batch processing https://github.com/wultra/enrollment-server/commit/0c70f964ab0531788cd9bdee98e744a84b6bc463

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
```diff
@@ -40,72 +42,133 @@
...
     @Query("SELECT p FROM OnboardingProcessEntity p WHERE p.status = :status AND p.identificationData = :identificationData")
-    Optional<OnboardingProcessEntity> findExistingProcessByIdentificationData(String identificationData, OnboardingStatus status);
+    Optional<OnboardingProcessEntity> findByIdentificationDataAndStatusWithLock(String identificationData, OnboardingStatus status);
 
+    /**
+     * Find an existing process by activation identifier and process status. The process is not locked.
+     *
+     * @param activationId Activation identifier.
+     * @param status Onboarding process status.
+     * @return Optional onboarding process.
```
banterCZ, line 73: Nice naming. If you want to be super perfect, you may cross reverence non/locking methods via `@see`.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/CommonOnboardingService.java
```diff
@@ -97,7 +91,8 @@ public OnboardingProcessEntity updateProcess(OnboardingProcessEntity process) {
 
     @Override
     public void updateProcess(final UpdateProcessRequest request) throws OnboardingProcessException {
-        final OnboardingProcessEntity process = findProcess(request.getProcessId());
+        // The onboarding process is locked until the end of the transaction
```
banterCZ, line 94: Just wondering about moving this sentence to javadoc, but not sure.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/OnboardingController.java
```diff
@@ -98,6 +98,7 @@ public ObjectResponse<OnboardingStartResponse> startOnboarding(@EncryptedRequest
             throw new PowerAuthEncryptionException("Invalid request received during onboarding");
         }
 
+        // Onboarding process is locked inside the service
```
banterCZ, line 101: dtto I suggest to move it to the to the javadoc.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
```diff
@@ -0,0 +1,653 @@
...
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@ConditionalOnProperty(
+        value = "enrollment-server-onboarding.identity-verification.enabled",
+        havingValue = "true"
+)
+@Service
+public class IdentityVerificationRestService {
+
+    private static final Logger logger = LoggerFactory.getLogger(IdentityVerificationRestService.class);
```
banterCZ, line 80: nitpick: do we prefer `@Slf4j` now or not?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
```diff
@@ -0,0 +1,653 @@
...
+
+        final String operationDescription = "initializing document verification SDK";
+        checkApiAuthentication(apiAuthentication, operationDescription);
+        checkEciesContext(eciesContext, operationDescription);
+        checkRequestObject(request, operationDescription);
+
+        final OwnerId ownerId = PowerAuthUtil.getOwnerId(apiAuthentication);
+        final String processId = request.getRequestObject().getProcessId();
+
+        logger.debug("Onboarding process will not be locked, {}", processId);
```
banterCZ, line 341: I know, it's debug only, but probably highlight to distinguish from `Onboarding process will be locked`
```suggestion
        logger.debug("Onboarding process will NOT be locked, {}", processId);
```

banterCZ (approved):
Great work!
Lest deploy and test it. I have only test-case to verify: delay in client evaluation.
Other are just little comments to consider.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/CommonOnboardingService.java
romanstrobl: I will use the debug logging.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/OnboardingController.java
romanstrobl: I will use the debug logging.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
romanstrobl: Yes, my bad.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
romanstrobl: I don't like all-caps in log messages. In case we need to use these debug messages I am sure we will be able to differentiate the locked and not-locked cases.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/DocumentSubmitSyncTask.java
```diff
@@ -40,8 +44,10 @@ public DocumentSubmitSyncTask(DocumentProcessingBatchService documentProcessingB
      * Scheduled task to check in progress document submits at the target provider
      */
     @Scheduled(cron = "${enrollment-server-onboarding.document-verification.checkInProgressDocumentSubmits.cron:0/5 * * * * *}", zone = "UTC")
-    @SchedulerLock(name = "checkInProgressDocumentSubmits", lockAtLeastFor = "1s", lockAtMostFor = "5m")
+    @SchedulerLock(name = SchedulerLockNames.DOCUMENT_SUBMIT_SYNC_LOCK, lockAtLeastFor = "100ms", lockAtMostFor = "5m")
```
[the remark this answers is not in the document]
romanstrobl: Good point, I will switch to the default, I think there's no point in having multiple different values.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
romanstrobl: Ok, good idea.

banterCZ (approved):

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/rest/RestOnboardingProvider.java
```diff
@@ -139,32 +139,22 @@ public ApproveConsentResponse approveConsent(final ApproveConsentRequest request
...
-            sink.tryEmitError(new OnboardingProviderException("Unable to evaluate client for " + request, t));
-        };
+                    .build();
 
-        try {
-            final ParameterizedTypeReference<ClientEvaluateResponseDto> type = ParameterizedTypeReference.forType(ClientEvaluateResponseDto.class);
-            restClient.postNonBlocking("/client/evaluate", requestDto, null, createHeaders(), type, onSuccess, onError);
         } catch (RestClientException e) {
-            sink.tryEmitError(new OnboardingProviderException("Unable to evaluate client for " + request, e));
+            throw new OnboardingProviderException("Unable to evaluate client for " + request);
```
banterCZ, line 156: Let's propagate cause
```suggestion
            throw new OnboardingProviderException("Unable to evaluate client for " + request, e);
```

romanstrobl: Sure.

PR #445, opened 2022-10-13
Fix #355: Add precondition check before finishing onboarding

### LINKED ISSUE #355: Add precondition check before finishing onboarding
We should check that all conditions were met before onboarding is finished, e.g.:
- user identity verification status in DB 
   - two ID documents accepted
- successful presence check
- successful OTP verification
- activation in correct state

This will make it harder for attackers to try to hack the onboarding process.

banterCZ:


enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/enumeration/ErrorOrigin.java
```diff
@@ -34,6 +34,7 @@ public enum ErrorOrigin {
 
     PROCESS_LIMIT_CHECK,
 
-    USER_REQUEST
+    USER_REQUEST,
 
+    FINAL_VALIDATION
```
banterCZ, line 39: I guess that this should be propagated to documentation because of reporting.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -106,7 +103,7 @@ public class IdentityVerificationService {
      * @param auditService Audit service.
      */
     @Autowired
-    public IdentityVerificationService(
+    IdentityVerificationService(
```
banterCZ, line 106: Let's keep the service _public_ because of API, but hide the constructor little bit, because `IdentityVerificationPrecompleteGuard` is not public on purpose.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
```diff
@@ -0,0 +1,173 @@
...
+ * <p>
+ * This should never happen for the state machine.
+ * It works as a safety stop.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+// TODO (racansky, 2022-10-14) consider make it Guard for Spring State Machine
+@Component
+@Slf4j
+class IdentityVerificationPrecompleteGuard {
```
romanstrobl, line 52: I am not sure we should call this class a `guard` when it is not an actual guard for state machine.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
```diff
@@ -0,0 +1,173 @@
...
+            logger.debug("Not all required documents are present for verification ID: {}", idVerification.getProcessId());
+            return Result.failed("Required documents not present");
+        }
+
+        if (!isPrefinalPhaseAndStateValid(idVerification)) {
+            logger.debug("Not valid phase and state for verification ID: {}", idVerification.getProcessId());
+            return Result.failed("Not valid phase and state");
+        }
+
+        if (!isOtpValid(idVerification)) {
```
romanstrobl, line 101: It would be nice to validate both activation and user verification OTPs. I assume it's not too difficult, can we add this?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
```diff
@@ -0,0 +1,173 @@
...
+
+    private boolean isPresenceCheckValid(final IdentityVerificationEntity identityVerification) {
+        if (!identityVerificationConfig.isPresenceCheckEnabled()) {
+            logger.trace("Presence check is disabled");
+            return true;
+        }
+        final RejectOrigin rejectOrigin = identityVerification.getRejectOrigin();
+        final ErrorOrigin errorOrigin = identityVerification.getErrorOrigin();
+
+        return errorOrigin != ErrorOrigin.PRESENCE_CHECK && rejectOrigin != RejectOrigin.PRESENCE_CHECK;
```
romanstrobl, line 143: This seems strange. Why would we allow any error origins or reject origins when the identity verification succeeds? Maybe this is a flaw in current logic, that the error is not reset when the step succeeds?

Ideally, there should rather be a check that there are no errors or rejections present.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    private boolean isPrefinalPhaseAndStateValid(final IdentityVerificationEntity idVerification) {
+        final IdentityVerificationPhase phase = idVerification.getPhase();
+        final IdentityVerificationStatus status = idVerification.getStatus();
+        return (phase == OTP_VERIFICATION && status == VERIFICATION_PENDING) ||
+                (phase == PRESENCE_CHECK && status == ACCEPTED && !identityVerificationConfig.isVerificationOtpEnabled()) ||
+                (phase == CLIENT_EVALUATION && status == ACCEPTED && !identityVerificationConfig.isVerificationOtpEnabled() &&!identityVerificationConfig.isPresenceCheckEnabled());
```
romanstrobl, line 151: Strange formatting `&&!identityVerificationConfig.isPresenceCheckEnabled()`.

romanstrobl (changes requested):
I have several comments which we should at least discuss before approving the pull request.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
```diff
@@ -0,0 +1,173 @@
...
+     *
+     * @param idVerification identity verification to evaluate
+     * @return evaluation result
+     */
+    Result evaluate(final IdentityVerificationEntity idVerification) throws RemoteCommunicationException {
+        final List<DocumentVerificationEntity> documentVerifications = idVerification.getDocumentVerifications().stream()
+                .filter(it -> DocumentStatus.ALL_PROCESSED.contains(it.getStatus()))
+                .collect(Collectors.toList());
+
+        if (!documentVerifications.stream()
```
romanstrobl, line 84: I am not sure this check is correct. A document can be `REJECTED` or `FAILED` and later `ACCEPTED` after reupload as another DB row.

We should rather check that 2 required documents were `ACCEPTED` and these documents have `usedForVerification = true`, no matter the status of the remaining documents.

romanstrobl (changes requested):
I found one more issue which will block the process in case of re-upload of documents which previously failed verification.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
romanstrobl: Ok.

enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/enumeration/ErrorOrigin.java
banterCZ: Prepared change of the documentation https://github.com/wultra/powerauth-product/pull/86

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
banterCZ: I agree that it may be confusing a little bit but the state machine is not the only one using the keyword `guard`.  IMHO `validator` is something much simpler. What about `check`? There is the same issue with `RequiredDocumentTypesGuard`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
banterCZ: 👀  Good catch. Logic with `REJECTED`, `FAILED`, and `ACCEPTED` is correct. But I have optimized it too much. There used to be
```java
List<DocumentVerificationEntity> processedDocVerifications =
                documentVerificationRepository.findAllDocumentVerifications(idVerification, DOCUMENT_STATUSES_PROCESSED);
```
which I replaced by
```java
idVerification.getDocumentVerifications().stream()
                .filter(it -> DocumentStatus.ALL_PROCESSED.contains(it.getStatus()))
                .collect(Collectors.toList());
```
It gives different result e.g. for `DISPOSED` representing resubmitted documents.

romanstrobl: 👍

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
romanstrobl: `Check` or `Validator` is fine, let's just not confuse reader with state machine terminology.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteGuard.java
banterCZ: Ok, let's remove it until #453

romanstrobl (approved):
Nice job.

PR #450, opened 2022-10-14
new reusable builds

zcgandcomp:
new reusable workflows part of wultra/tasklist#23


.github/workflows/maven-deploy.yml
```diff
@@ -0,0 +1,50 @@
+name: Deploy with Maven
+
+on:
+  workflow_dispatch:
+    branches:
+      - 'develop'
+      - 'master'
+      - 'release/*'
```
romanstrobl, line 8: Should be `releases/*`.

romanstrobl (changes requested):
The release branch name is incorrect, otherwise I do not see any issues.

romanstrobl (approved):
Looks OK.

PR #456, opened 2022-10-19
Fix #449: Cleaning task may not expire all records

### LINKED ISSUE #449: Cleaning task may not expire all records
We use in DB queries following condition: `WHERE *.timestampLastUpdated < :timestamp`.

Such condition is not reliable, because `timestampLastUpdated` is nullable. Furthermore, this may extend the timeout window, which is undesirable. We should use `timestampCreated` for expiration queries which is not null.

banterCZ:
Replace nullable timestampLastUpdated by timestampCreated in the where condition for expiration queries


enrollment-server-onboarding/src/test/resources/com/wultra/app/onboardingserver/task/cleaning/CleaningServiceTest.testTerminateExpiredDocumentVerifications.sql
```diff
@@ -1,9 +1,9 @@
 INSERT INTO es_identity_verification(id, activation_id, user_id, process_id, status, phase, timestamp_created, timestamp_last_updated) VALUES
     ('a6055e8b-4ac0-45dd-b68e-29f4cd991a5c', 'a1', 'u1', 'p1', 'IN_PROGRESS', 'PRESENCE_CHECK', now(), now()),
-    ('8d036a18-f51f-4a30-92cd-04876172ebca', 'a2', 'u2', 'p2', 'IN_PROGRESS', 'PRESENCE_CHECK', now() - 1, now() - 1), -- to be terminated
-    ('c918e1c4-5ca7-47da-8765-afc92082f717', 'a3', 'u3', 'p3', 'ACCEPTED', 'COMPLETED', now() - 1, now() - 1);
+    ('8d036a18-f51f-4a30-92cd-04876172ebca', 'a2', 'u2', 'p2', 'IN_PROGRESS', 'PRESENCE_CHECK', now() - interval '1' day, now()), -- to be terminated
```
banterCZ, line 3: - `now() - 1` is equivalent of `now() - interval '1' day` but is more legible
- set `timestamp_last_updated` to `now()` to fail tests first

romanstrobl (approved):
Looks OK.

PR #461, opened 2022-10-20
Fix #460: Configuration startup check

### LINKED ISSUE #460: Configuration startup check
Add safety break of OTP verification and presence check configuration.

banterCZ:
Presence check and OTP verification have to be both disabled or both enabled.


romanstrobl (approved):
Nice improvement.

Issue #453, opened 2022-10-18
Improve working with error states


banterCZ:
Improve working with error states. Identity verification status and phases are rewritten during the process. E.g., for two-step SCA we abuse _error origin_ and _reject origin_ #302
Possible solution is to historize data or introduce extra column to keep some flags.

PR #462, opened 2022-10-20
Add TODO to PresenceCheckLimitService for #453

banterCZ:
OTP verification could be turned off, logic to count presence check limit should be based on another data.


romanstrobl (approved):
Good idea to add the TODO.

PR #471, opened 2022-10-21
Fix #470: Activation flag at verification status

### LINKED ISSUE #470: Stop looking for activation flag at verification status

banterCZ:


romanstrobl (approved):
OK, but we'll need to test the impact on the mobile app.

romanstrobl (changes requested):
Let's switch to using `VERIFICATION_PENDING` flag instead of `VERIFICATION_IN_PROGRESS`. 

We still need to return the `NOT_INITIALIZED` state because otherwise the app would not call the initialization endpoint and thus a new DB row would not be created for the next identity verification round.

romanstrobl (approved):
Excellent.

PR #473, opened 2022-10-23
wultra/tasklist#23 do not skip war deploy in case of jfrog

zcgandcomp:


romanstrobl (approved):
Looks OK.

enrollment-server-onboarding/pom.xml
```diff
@@ -278,6 +270,27 @@
...
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-deploy-plugin</artifactId>
+                        <configuration>
+                            <skip>true</skip>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
```
banterCZ, line 292: It is possible to configure plugin just via properties, it does not have to be explicitly declared. https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html#skip
```suggestion
            <properties>
                <maven.deploy.skip>true</maven.deploy.skip>
            </properties>
```

banterCZ: Of course, consistency is important. This was just a hint, how it is possible to write it differently.

PR #501, opened 2022-11-07
Fix #494: Check that only one verification ID is present for accepted documents

### LINKED ISSUE #494: Check that only one verification ID is present for accepted documents
In method `getVerificationId` we should improve the check of consistency of extracted verification ID:

https://github.com/wultra/enrollment-server/blob/36ab899215fbf2d79a3df068b7bfb5caf4530292/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java#L126 

This check is still suboptimal. We rather want to execute `findAll()`, put all results into a `Set` and make sure this set has only one item which is the ID of verification we want to use. In case of multiple IDs found, this is a severe error in business logic and we want to abort the processing.

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
```diff
@@ -103,33 +98,49 @@ public void initClientEvaluation(final OwnerId ownerId, final IdentityVerificati
...
                 .build();
 
         final int maxFailedAttempts = config.getClientEvaluationMaxFailedAttempts();
         for (int i = 0; i < maxFailedAttempts; i++) {
             try {
                 final EvaluateClientResponse response = onboardingProvider.evaluateClient(request);
                 processEvaluationSuccess(identityVerification, ownerId, response);
-                break;
-            } catch (Throwable t) {
-                processEvaluationError(identityVerification, ownerId, t);
```
banterCZ, line 120: When we really want to allow more attempts, we must not fail the identity verification after the first error. Covered by test since now.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
```diff
@@ -103,33 +98,49 @@ public void initClientEvaluation(final OwnerId ownerId, final IdentityVerificati
...
+                .verificationId(verificationId)
                 .build();
 
         final int maxFailedAttempts = config.getClientEvaluationMaxFailedAttempts();
         for (int i = 0; i < maxFailedAttempts; i++) {
             try {
                 final EvaluateClientResponse response = onboardingProvider.evaluateClient(request);
                 processEvaluationSuccess(identityVerification, ownerId, response);
-                break;
-            } catch (Throwable t) {
```
banterCZ, line 119: > Catching either Throwable or Error will also catch OutOfMemoryError and InternalError, from which an application should not attempt to recover.

https://rules.sonarsource.com/java/RSPEC-1181

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
```diff
@@ -156,24 +165,26 @@ private void processEvaluationSuccess(final IdentityVerificationEntity identityV
...
         identityVerification.setErrorOrigin(ErrorOrigin.PROCESS_LIMIT_CHECK);
         identityVerification.setTimestampFailed(ownerId.getTimestamp());
         final IdentityVerificationPhase phase = identityVerification.getPhase();
-        saveInExistingTransaction(status ->
-                identityVerificationService.moveToPhaseAndStatus(identityVerification, phase, IdentityVerificationStatus.FAILED, ownerId));
+        identityVerificationService.moveToPhaseAndStatus(identityVerification, phase, IdentityVerificationStatus.FAILED, ownerId);
     }
 
-    private void saveInExistingTransaction(final Consumer<TransactionStatus> consumer) {
-        transactionTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_MANDATORY);
```
banterCZ, line 176: Default propagation level is `REQUIRED`. I do not see any reason yet to be more strict.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
```diff
@@ -103,33 +98,49 @@ public void initClientEvaluation(final OwnerId ownerId, final IdentityVerificati
...
-        return identityVerification.getDocumentVerifications().stream()
+        final Set<String> verificationIds = identityVerification.getDocumentVerifications().stream()
                 .filter(DocumentVerificationEntity::isUsedForVerification)
                 .filter(it -> it.getStatus() == DocumentStatus.ACCEPTED)
-                .findAny()
                 .map(DocumentVerificationEntity::getVerificationId)
-                .orElseThrow(() -> new IllegalStateException("No accepted document verification for " + identityVerification));
+                .collect(toSet());
+
+        if (verificationIds.size() == 1) {
```
banterCZ, line 138: The main goal of the issue was about this change, but it turned out that the service deserves larger cleanup.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
romanstrobl: Ok.

romanstrobl (approved):
Looks OK.

PR #505, opened 2022-11-08
Fix #504 correct distribution repo for maven

### LINKED ISSUE #504: different repository for maven distribution
s01.* should be used for distribution management instead of oss

zcgandcomp:


romanstrobl (approved):
Approved.

banterCZ (approved):

PR #517, opened 2022-11-09
Fix #516: Document may be verified twice

### LINKED ISSUE #516: Document may be verified twice
See:
https://github.com/wultra/enrollment-server/blob/3ee6eba168a1b1be6bc88c60d75a83058dd41056/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/verification/VerificationProcessingService.java#L168

This logic seems obsolete, because partial document verifications are not used anymore.

romanstrobl:


banterCZ (approved):
Ok, let's simplify the logic.

PR #519, opened 2022-11-11
Kill verify on submit

romanstrobl:
I am done with parameter `documentVerificationOnSubmitEnabled`, this is causing neverending confusion and invalid configuration issues.

I added a TODO for synchronous processing which we can fix in another issue. The document state does not currently reflect exactly what happened on server (document is already uploaded). I added similar TODO into tests.


banterCZ:
👏 Well done!

PR #523, opened 2022-11-11
Fix #522: Maximum number of presence check attempts is off by 1

### LINKED ISSUE #522: Maximum number of presence check attempts is off by 1
When maximum number is set to five, six attempts are allowed.

romanstrobl:
The issue is caused by the OTP being generated at the end, however the check is in the beginning, so 1 OTP attempt needs to be subtracted.


banterCZ (approved):
Thanks for the fix!

PR #526, opened 2022-11-15
Fix #493: Publish Inbox API for Mobile Token

### LINKED ISSUE #493: Publish Inbox API for Mobile Token
Publish an Inbox API for Mobile Token as a facade for Push Server REST API with following endpoints:
- POST `/api/inbox/count` - get count of unread messages in inbox
- POST `/api/inbox/message/list` - get a page with messages with given page number and size
- POST `/api/inbox/message/detail` - get detail of a message with given message identifier
- POST `/api/inbox/message/read` - mark a message with given message identifier as read
- POST `/api/inbox/message/read-all` - mark all messages as read for user

The reason of using `POST` for all these endpoints is that we do not want to expose any sensitive details in the URL and for consistency with e.g. POST `/api/auth/token/app/operation/list`.

romanstrobl:


enrollment-server-api-model/pom.xml
```diff
@@ -33,6 +33,24 @@
         <version>1.4.0-SNAPSHOT</version>
     </parent>
 
+    <dependencies>
+        <dependency>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+            <version>${jakarta-validation-api.version}</version>
+            <scope>compile</scope>
```
banterCZ, line 41: Default. Any reason to declare it explicitly?

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/InboxController.java
```diff
@@ -0,0 +1,187 @@
...
+ * Controller with facade for Inbox services in Push server.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@ConditionalOnExpression("!'${powerauth.push.service.url}'.empty and ${enrollment-server.inbox.enabled}")
+@RestController
+@RequestMapping(value = "api/inbox")
+public class InboxController {
+
+    private static final Logger logger = LoggerFactory.getLogger(InboxController.class);
```
banterCZ, line 56: `@Slf4j`?

banterCZ (approved):
Looks good.

enrollment-server-api-model/pom.xml
romanstrobl: Just a copy-paste reason.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/InboxController.java
romanstrobl: Yes.

PR #534, opened 2022-11-17
Fix #529: Required documents should be configurable

### LINKED ISSUE #529: Required documents should be configurable
Required documents should be configurable for `RequiredDocumentTypesCheck`. By default ID and one another should be required. Right now, not needed to expose that at configuration endpoint.

Follow-up to #354

banterCZ:


romanstrobl (approved):
Nice improvement.

PR #535, opened 2022-11-17
Fix #524: Do not return document status from submit

### LINKED ISSUE #524: Do not return document status from submit
Do not return document status from `document/submit`. Mobile app should call `document/status`

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/controller/api/IdentityVerificationController.java
```diff
@@ -138,7 +138,7 @@ public ObjectResponse<IdentityVerificationStatusResponse> checkIdentityVerificat
     @PowerAuthToken(signatureType = {
             PowerAuthSignatureTypes.POSSESSION
     })
-    public ObjectResponse<DocumentSubmitResponse> submitDocuments(@EncryptedRequestBody ObjectRequest<DocumentSubmitRequest> request,
+    public Response submitDocuments(@EncryptedRequestBody ObjectRequest<DocumentSubmitRequest> request,
```
banterCZ, line 141: 👀 @kober32

romanstrobl (approved):
Agreed, returned value is not necessary for the mobile app.

PR #536, opened 2022-11-20
Fix #513: Result investigateSamples is ignored

### LINKED ISSUE #513: Result investigateSamples is ignored
It seems to me, that when `investigateSample` is called in sync mode, the result is ignored, and `investigation/{id}` is called unnecessarily. Investigate it little bit deeper.

https://github.com/wultra/enrollment-server/blob/979c4614338e5876feff7514f76f2d140f2e864a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java#L246

banterCZ:


romanstrobl (changes requested):
I think this change is OK, however I would expect that the background which checks investigation results would be disabled as well.

romanstrobl (approved):
Excellent!

PR #540, opened 2022-11-22
Fix #539: Polish iProov provider

### LINKED ISSUE #539: Polish iProov provider

banterCZ:
- Polish IProovRestApiService
- Add ApproveConsentResponse#toString
- Remove empty IProovPresenceCheckProvider#cleanupIdentityData
- Log iProov token and userId

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/service/IProovRestApiService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/service/IProovRestApiService.java
@@ -108,7 +108,7 @@ public class IProovRestApiService {
     public ResponseEntity<String> generateEnrolToken(OwnerId id) throws RestClientException {
         final ServerClaimRequest request = createServerClaimRequest(id);
 
-        logger.info("Calling /claim/enrol/token userId={}, {}", request.getUserId(), id);
+        logger.debug("Calling /claim/enrol/token userId={}, {}", request.getUserId(), id);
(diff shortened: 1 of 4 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/service/IProovRestApiService.java
```diff
@@ -162,45 +164,30 @@ public ResponseEntity<String> generateVerificationToken(OwnerId id) throws RestC
...
         request.setClient("Wultra Enrollment Server, activationId: " + id.getActivationId()); // TODO value from the device
         request.setIp("192.168.1.1"); // TODO deprecated but still required
         request.setRiskProfile(configProps.getRiskProfile());
         request.setToken(token);
 
-        String userId = getUserId(id);
+        final String userId = getUserId(id);
         request.setUserId(userId);
 
+        logger.info("Calling /claim/verify/validate userId={}, token={}, {}", userId, token, id);
```
banterCZ, line 178: It seems insecure, but recommended by iProov, see https://docs.iproov.com/docs/Content/ImplementationGuide/api/api-logging.htm

romanstrobl, line 178: This does not seem to be a good idea. We store the value in DB. I suggest to switch to `debug`, in case there will be any issues we can always enable `debug` on individual classes.

banterCZ: Changed to `debug`.

PR #541, opened 2022-11-22
Fix #511: Investigate license issue with Zen ID provider

### LINKED ISSUE #511: Investigate license issue with Zen ID provider
In case Zen ID runs out of licenses, all onboarding processes will get stuck on verification. We should investigate if there is a way how to handle this issue in a slightly more customer friendly way (e.g. fail process). Log errors in high level, potentially `FATAL` or at least `ERROR`, because this state breaks all functionality.

banterCZ:


romanstrobl (approved):
Useful improvement.

PR #548, opened 2022-11-29
Fix #547: Allow MFA token for inbox services

### LINKED ISSUE #547: Allow MFA token for inbox services

petrdvorak:


banterCZ (approved):
Looks good.

romanstrobl (approved):
Good idea. Not sure we need `POSSESSION_KNOWLEDGE_BIOMETRY`, this is a bit much work for the user to use both PIN and biometry.

PR #550, opened 2022-12-01
Fix #549: Switch jakarta.validation imports to javax.validation

### LINKED ISSUE #549: Switch jakarta.validation imports to javax.validation
The `jakarta.validation` imports should be switched to `javax.validation`, so that the whole stack uses the same validation annotations.

banterCZ:
- downgrade jakarta bean validation to 2.0.2

### DIFF
--- a/enrollment-server-api-model/pom.xml
+++ b/enrollment-server-api-model/pom.xml
@@ -37,7 +37,6 @@
         <dependency>
             <groupId>jakarta.validation</groupId>
             <artifactId>jakarta.validation-api</artifactId>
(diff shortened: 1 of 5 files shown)


enrollment-server-api-model/pom.xml
```diff
@@ -37,7 +37,6 @@
         <dependency>
             <groupId>jakarta.validation</groupId>
             <artifactId>jakarta.validation-api</artifactId>
-            <version>${jakarta-validation-api.version}</version>
```
banterCZ, line 40: Default to Spring Boot versioning.

PR #556, opened 2022-12-06
Fix #555: Remove mandatory request object for services that do not need it

### LINKED ISSUE #555: Remove mandatory request object for services that do not need it

petrdvorak:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/InboxController.java
```diff
@@ -68,7 +68,7 @@ public InboxController(PushServerClient pushClient) {
             PowerAuthSignatureTypes.POSSESSION_KNOWLEDGE,
             PowerAuthSignatureTypes.POSSESSION_KNOWLEDGE_BIOMETRY
     })
-    public ObjectResponse<GetInboxCountResponse> countUnreadMessages(@RequestBody ObjectRequest<GetInboxCountRequest> objectRequest, @Parameter(hidden = true) PowerAuthApiAuthentication apiAuthentication) throws InboxException {
+    public ObjectResponse<GetInboxCountResponse> countUnreadMessages(@Parameter(hidden = true) PowerAuthApiAuthentication apiAuthentication) throws InboxException {
```
hvge, line 71: We should also delete the unused class `GetInboxCountRequest`, but I don't know whether it's correct to declare method as "POST" and then use no post payload.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/InboxController.java
```diff
@@ -160,7 +161,7 @@ public Response readMessage(@RequestBody ObjectRequest<InboxReadRequest> objectR
             PowerAuthSignatureTypes.POSSESSION_KNOWLEDGE,
             PowerAuthSignatureTypes.POSSESSION_KNOWLEDGE_BIOMETRY
     })
-    public Response readAllMessages(@RequestBody ObjectRequest<InboxReadAllRequest> objectRequest, @Parameter(hidden = true) PowerAuthApiAuthentication apiAuthentication) throws InboxException {
+    public Response readAllMessages(@Parameter(hidden = true) PowerAuthApiAuthentication apiAuthentication) throws InboxException {
```
hvge, line 164: The same with unused class as for `countUnreadMessages()`

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/InboxController.java
petrdvorak: @hvge The method definition says that the incoming request body is not processed. We do not need to process the request body since there is no usable payload. So yes - this approach is correct. Good point about deleting unused class, though!

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/InboxController.java
petrdvorak: dtto, good point.

zcgandcomp (approved):

hvge (commented):
Looks OK

PR #557, opened 2022-12-07
propagate secrets for maven test

zcgandcomp:
propagate secrets for maven test
correct name of deploy job

### DIFF
--- a/.github/workflows/maven-deploy.yml
+++ b/.github/workflows/maven-deploy.yml
@@ -29,7 +29,7 @@ on:
 jobs:
   maven-deploy-internal:
     if: ${{ github.event_name == 'push' }}
-    name: Deploy to ${{ inputs.environment }}
+    name: Deploy to jfrog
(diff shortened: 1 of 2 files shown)


petrdvorak (approved):
Looks OK

.github/workflows/maven-deploy.yml
```diff
@@ -40,7 +40,7 @@ jobs:
 
   maven-deploy-public:
     if: ${{ github.event_name == 'workflow_dispatch' }}
-    name: Deploy to ${{ inputs.environment }}
+    name: Deploy by parameter
```
banterCZ, line 43: Really? I would expect something like `oss.sonatype`

banterCZ (approved):
Looks good

.github/workflows/maven-deploy.yml
zcgandcomp: the jobs are for automated push - it goes to jfrog all the time and the second job is for manual deployment, so you can specify the target. I change name of jobs to be more suitable..

banterCZ (approved):

PR #559, opened 2022-12-08
Fix #558: SMS does not respect accept-language

### LINKED ISSUE #558: SMS does not respect accept-language
Locale is taken from `LocaleContextHolder.getLocale()`, which is a problem, because it is initialized by the backend process.

The first SMS is fine, because it is called directly from `OnboardingController`.

banterCZ:
Update script for existing environments
```sql
alter table es_onboarding_process add column custom_data VARCHAR(1024) NOT NULL default '{}';
```


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationOtpService.java
```diff
@@ -73,6 +76,8 @@ public class IdentityVerificationOtpService {
 
     private final AuditService auditService;
 
+    private final ObjectMapper mapper = new ObjectMapper();
```
petrdvorak, line 79: Lets review what `ObjectMapper` instances we use.

banterCZ: I agree, this was quick and working fix.

romanstrobl:
What happens when the user switches language during onboarding process in app settings, will the next SMS arrive in correct language?

banterCZ:
> What happens when the user switches language during onboarding process in app settings, will the next SMS arrive in correct language?

Good point! We have discussed it with @kober32 and accepted that scenario. Do you expect that it will happen often? If so, we may update custom_data more often, but is it too big overhead?

romanstrobl:
I suggest to raise a separate ticket. The overhead is minimal, we can just check whether the language has changed and update the locale.

PR #560, opened 2022-12-09
Add github action for dependency check

banterCZ:
See https://github.com/marketplace/actions/dependency-check


banterCZ:
Will be checked at Snyk.

PR #567, opened 2022-12-12
Fix #566: Change H2 dependency scope to test

### LINKED ISSUE #566: Change H2 dependency scope to test
Right not, it is `runtime`, which is not needed and reported by vulnerability scans.

banterCZ:


romanstrobl (approved):
Good idea.

PR #574, opened 2022-12-13
Fix #573: Coverity - Dereference null return value in MobileTokenConverter

### LINKED ISSUE #573: Coverity - Dereference null return value in MobileTokenConverter

banterCZ:


romanstrobl (approved):
Looks OK.

PR #591, opened 2022-12-23
Fix #590: Add PostApprovalScreen to UI extension payload

### LINKED ISSUE #590: Add PostApprovalScreen to UI extension payload
```json
"postApprovalScreen": {
  "type": "MERCHANT_REDIRECT",
  "heading": "...",
  "message": "...",
  "payload": {
     "redirectText": "...",
     "redirectUrl": "https://...",
     "countdown": 5
    }
}
```

Relates to #306

banterCZ:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -95,46 +101,61 @@ public Operation convert(OperationDetailResponse operationDetail, OperationTempl
...
 
-    private UiExtensions convertUiExtension(final OperationDetailResponse operationDetail, final OperationTemplateEntity operationTemplate) throws JsonProcessingException {
+    private UiExtensions convertUiExtension(
+            final OperationDetailResponse operationDetail,
+            final OperationTemplateEntity operationTemplate,
+            final StringSubstitutor substitutor) throws JsonProcessingException {
+
         if (StringUtils.hasText(operationTemplate.getUi())) {
-            final String uiJsonString = operationTemplate.getUi();
+            final String uiJsonString = substitutor == null ? operationTemplate.getUi() : substitutor.replace(operationTemplate.getUi());
```
banterCZ, line 158: It is possible to substitute the whole UI.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
```diff
@@ -159,4 +148,114 @@ void testConvertUiRiskFlagsXCF() throws Exception {
...
         assertEquals(PreApprovalScreen.ScreenType.WARNING, preApprovalScreen.getType());
     }
+
+    @Test
+    void testConvertUiPostApprovalMerchantRedirect() throws Exception {
+        final OperationDetailResponse operationDetail = createOperationDetailResponse();
+
+        final OperationTemplateEntity operationTemplate = new OperationTemplateEntity();
+        operationTemplate.setUi("{\n" +
+                "  \"postApprovalScreen\": {\n" +
```
banterCZ, line 158: I realize that the original proposal was to have polymorphism even for `postApprovalScreen` itself, but it does not make sense to me yet. We may extend it later. More important is to have generic payload, which you may see below.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
```diff
@@ -159,4 +148,114 @@ void testConvertUiRiskFlagsXCF() throws Exception {
...
+                .isEqualTo(expectedPayload);
+    }
+
+    @Test
+    void testConvertUiPostApprovalInfoMessage() throws Exception {
+        final OperationDetailResponse operationDetail = createOperationDetailResponse();
+
+        final OperationTemplateEntity operationTemplate = new OperationTemplateEntity();
+        operationTemplate.setUi("{\n" +
+                "  \"postApprovalScreen\": {\n" +
```
banterCZ, line 235: This is an example of generic _post approval screen_. Payload is missing on purpose. We may cover such scenario without discriminator.

romanstrobl (changes requested):
As discussed, in the first version we will also need the `REVIEW` payload subtype. The original suggestion was:
```
"postApprovalScreen": {
   "type": "REVIEW",
   "heading": "...",
   "message": "...",
   "attributes": [
      {
         "type": "AMOUNT",
         // atd.
      }
   ]
}
```
@petrdvorak can provide more details about this requirement. 

I think we should also include a `GENERIC` payload subtype where the payload is fully customizable using the `${payload}` parameterization to make sure this generic subtype won't require some changes later on.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
romanstrobl: This is not necessary right now as far as I can tell.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
romanstrobl: I am not sure this is a good approach.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
banterCZ: @romanstrobl In fact, I was trying to say that I had not implemented what you mentioned at https://github.com/wultra/enrollment-server/pull/591#pullrequestreview-1229232076

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
banterCZ: As far as I remember, that was an agreement from the last meeting, but unfortunately is not written down :(

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
banterCZ: @kober32 Do you need a discriminator even for `postApprovalScreen` or is enough only for `payload`?
```json
"postApprovalScreen": {
   "type": "REVIEW",
```

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
romanstrobl: My suggestion is to allow substitution of the payload for the `GENERIC` subtype. AFAIK there is no other need for substitution at the moment, but we can of course discuss it. @kober32

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
romanstrobl: My understanding is that the payload for the `GENERIC` subtype should not be missing, but it should be customizable using a subtitutor. This way the backend can support generic functionality which is added to the app before next backend release, using parameter substitution. @kober32 do we agree on this use case?

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
romanstrobl: AFAIK both `heading` and `message` should be always required.

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/attributes/Attribute.java
```diff
@@ -61,6 +71,8 @@ public enum Type {
     /**
      * Type of the attribute.
      */
+    // JsonIgnore added, otherwise type was serialized twice
+    @JsonIgnore
```
banterCZ, line 75: Do we really need the field `type`? I am about to remove it completely. Could we rely on concrete subtype?

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/PostApprovalScreen.java
```diff
@@ -0,0 +1,62 @@
...
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@Data
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = MerchantRedirectPostApprovalScreen.class, name = "MERCHANT_REDIRECT"),
+        @JsonSubTypes.Type(value = GenericPostApprovalScreen.class, name = "GENERIC"),
+        @JsonSubTypes.Type(value = ReviewPostApprovalScreen.class, name = "REVIEW")
+})
+public abstract class PostApprovalScreen {
```
romanstrobl, line 38: Perfect, exactly what I had in mind.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
banterCZ: I updated the comment to be more obvious that it is only a snippet.

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/PostApprovalScreen.java
romanstrobl: Btw, I would either use alphabetic ordering of subtypes or put `GENERIC` subtype on top.

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/GenericPostApprovalScreen.java
```diff
@@ -0,0 +1,63 @@
...
+    public void setPayload(GenericPayload payload) {
+        this.payload = payload;
+    }
+
+    /**
+     * Specialization of {@link Payload} for generic usage.
+     */
+    public static class GenericPayload implements Payload {
+
+        private final Map<String, Object> properties = new HashMap<>();
```
romanstrobl, line 51: I prefer `LinkedHashMap` to keep the order when used twice.

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/attributes/Attribute.java
romanstrobl: We should not modify an existing model class.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
banterCZ: As we agreed on the call, the whole UI will be substituted. 

- [ ] Add test case for dangerous json characters

docs/Mobile-Tokens.md
```diff
@@ -0,0 +1,80 @@
+# Mobile Tokens
```
petrdvorak, line 1: Let's call this "Operation Extensions" in the "Mobile Token" section.

petrdvorak (changes requested):
My main comment goes to the documentation, I am not sure if we gain anything with the switch from explicit `type` attribute to JSON annotation - the new code does not improve readability, since we serialize a field that is not visibly present in the code.

petrdvorak (approved):
One minor thing to fix in documentation - let's add the section to sidebar, and we can keep L2 headings in Home file.

romanstrobl (approved):
Looks OK.

PR #598, opened 2023-01-05
Fix #595: Onboarding adapter POST /process/event

### LINKED ISSUE #595: Onboarding adapter POST /process/event
Call `POST /process/event` at ATS adapter. It makes possible to send SMS, call FDS etc. from ATS.

[adapter-api.txt](https://github.com/wultra/enrollment-server/files/10336679/adapter-api.txt)

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/model/request/ProcessEventRequest.java
```diff
@@ -0,0 +1,56 @@
...
+    private String userId;
+
+    @NonNull
+    private String identityVerificationId;
+
+    /**
+     * Event type, e.g. FINISHED.
+     */
+    @NonNull
+    private String type;
```
romanstrobl, line 52: I suggest to introduce an enumeration, similar to `OtpType`. We know upfront the event types, so no need for a generic `String`.

romanstrobl (approved):
Looks good, I have a suggestion for improvement of type safety.

PR #602, opened 2023-01-09
Fix #600: Store IP address and user-agent

### LINKED ISSUE #600: Store IP address and user-agent
Store IP address and user-agent (e.g. `PowerAuthNetworking/1.1.6 (cs; wifi) com.wultra.app.MobileToken.wuat/2.0.0 (Apple; iOS/15.4.1; iphone9,3)`) to process custom data.

See

https://github.com/wultra/enrollment-server/blob/develop/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java#L232

https://github.com/wultra/enrollment-server/blob/2fc1087d59551c648b2f7df929c2137650c21196/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/RequestContextConverter.java#L31

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -414,30 +418,40 @@ private String lookupUser(final OnboardingProcessEntity process, final Map<Strin
...
-    private OnboardingProcessEntity resumeExistingProcess(final OnboardingProcessEntity process, final Map<String, Object> identification) {
+    private OnboardingProcessEntity resumeExistingProcess(
+            final OnboardingProcessEntity process,
+            final Map<String, Object> identification,
+            final RequestContext requestContext) {
+
         logger.debug("Resuming process ID: {}", process.getId());
         process.setTimestampLastUpdated(new Date());
-        new OnboardingProcessEntityWrapper(process).setLocale(LocaleContextHolder.getLocale());
+        setProcessCustomData(process, requestContext);
```
banterCZ, line 454: Right now, we are rewriting `ipAddress` and `user-agent` when resuming the process. So the last attempt wins. I can imagine to keep only origin values or both.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/RequestContextConverter.java
```diff
@@ -0,0 +1,88 @@
...
+import javax.servlet.http.HttpServletRequest;
+import java.util.List;
+
+/**
+ * Converter for HTTP request context information.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+// TODO (racansky, 2023-01-09) duplicates com.wultra.app.enrollmentserver.impl.service.converter.RequestContextConverter, extract to lime-java-core
```
banterCZ, line 32: Enrollment server without onboarding features does not depend on `enrollment-server-onboarding-common`, so we do not have any other reasonable _common_ module. `lime-java-core` has been already released. Let's keep duplicity here for a while and extract it later on.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
romanstrobl: Yes, this is OK.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/RequestContextConverter.java
romanstrobl: OK.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/RequestContextConverter.java
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param source HttpServletRequest instance.
+     * @return Request context data.
+     */
+    public RequestContext convert(HttpServletRequest source) {
+        if (source == null) {
+            return null;
+        }
+        final RequestContext destination = new RequestContext();
+        destination.setUserAgent(source.getHeader("User-Agent"));
```
romanstrobl, line 65: Can we use a constant for the "User-Agent" string?

romanstrobl (approved):
Nice job.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/RequestContextConverter.java
banterCZ: See https://github.com/wultra/lime-java-core/issues/151

PR #603, opened 2023-01-10
Add javadoc to OperationTemplateEntity#placeholder

banterCZ:


romanstrobl (approved):
Can we add link to this URL for further explanation?

https://github.com/wultra/powerauth-cloud/blob/develop/docs/Configuration.md#mobile-token-operation-localization

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/database/entity/OperationTemplateEntity.java
```diff
@@ -45,6 +45,9 @@ public class OperationTemplateEntity implements Serializable {
     @Column(name = "id", nullable = false)
     private Long id;
 
+    /**
+     * Operation type of operation at PowerAuth server.
```
romanstrobl, line 49: `Operation type of operation` - is this intended?

banterCZ:
> Can we add link to this URL for further explanation?
> 
> https://github.com/wultra/powerauth-cloud/blob/develop/docs/Configuration.md#mobile-token-operation-localization

It's link from a public repo to a private repo. Do we want it?

PR #615, opened 2023-01-11
Fix #614: Do not include FDS specifics in process finish even

### LINKED ISSUE #614: Do not include FDS specifics in process finish event
We should implement FDS product specifics in the downstream, because we do not want to include FDS product specifics in upstream.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #617, opened 2023-01-11
Fix #616: Remove endpoint for soft update and forced update from REST API

### LINKED ISSUE #616: Remove endpoint for soft update and forced update from REST API
The initial implementation of soft update and forced update support backend is not finished yet, we will implement this functionality in next release.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #620, opened 2023-01-12
Fix #619: Use RequestContextConverter from lime-core

### LINKED ISSUE #619: Use RequestContextConverter from lime-core

banterCZ:


romanstrobl (approved):
Looks OK.

PR #651, opened 2023-02-02
Fix #650: Improve MobileTokenConverter for attributes

### LINKED ISSUE #650: Improve MobileTokenConverter for attributes
Fix static code analysis and add tests.

banterCZ:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -184,89 +184,71 @@ private UiExtensions convertUiExtension(
         }
     }
 
-    private Attribute buildAttribute(OperationTemplateParam templateParam, Map<String, String> params) {
+    private static Optional<Attribute> buildAttribute(final OperationTemplateParam templateParam, final Map<String, String> params) {
```
banterCZ, line 187: Change of `null` to `Optional` is in a separate commit, if we do not like it.

romanstrobl (approved):
Nice cleanup!

PR #659, opened 2023-02-07
Fix #655: Use API and SPI document annotations from lime-core

### LINKED ISSUE #655: Use API and SPI document annotations from lime-core
Replace API and SPI annotations introduced in #221 by common ones from lime-core https://github.com/wultra/lime-java-core/issues/160

banterCZ:


pom.xml
```diff
@@ -99,9 +99,7 @@
         <spring-statemachine.version>3.2.0</spring-statemachine.version>
         <swagger-annotations.version>2.2.8</swagger-annotations.version>
         <springdoc-openapi.version>1.6.14</springdoc-openapi.version>
-        <wultra-rest-client.version>1.6.0</wultra-rest-client.version>
```
banterCZ, line 102: @romanstrobl May we use only single version? Lime core modules are not released separately.

romanstrobl: Yes we can.

romanstrobl (approved):
Great.

PR #662, opened 2023-02-07
Issues/661 missing package lock.json

zcgandcomp:


romanstrobl (changes requested):
No, we don't want to do this. We should remove the JavaScript code instead.

romanstrobl (approved):
Approved.

banterCZ (approved):
Yes, let's delete it. I guess this was important in the beginning as a PoC but not used anymore.

PR #668, opened 2023-02-10
Fix #666: Return the current server timestamp in the operation list (backport)

### LINKED ISSUE #666: Return the current server timestamp in the operation list

petrdvorak:


romanstrobl (approved):
Looks OK.

PR #703, opened 2023-04-11
Fix #699: Use one property for selecting userinfo bean

### LINKED ISSUE #699: Use one property for selecting userinfo bean
We use two boolean properties to pick the bean here:

https://github.com/wultra/enrollment-server/blob/develop/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/configuration/UserInfoProviderConfiguration.java#L50

As a result, we can have a bean clash in case the following two properties are set to `true`:

```
enrollment-server.user-info.minimal-provider.enabled
enrollment-server.user-info.rest-provider.enabled
```

We should replace this property with an enum property, i.e.:

```
enrollment-server.user-info.provider=MINIMAL/REST
```

banterCZ:


zcgandcomp:
A question - originally we spoke about 3 levels

- No user-info
- Internal user-info
- proxied user-info

It would make more sense to me to have the following:

- one property (boolean) to turn the user-info - implicitly on the internal user-info level, default false,
- one property (string) to configure proxied call - explicitly switching to the proxied user-info.

This will be IMHO a more fool-proof configuration setup. 

The reviewed change will do a similar job. After the last config issues, I believe this will crash - at least in Azure without having a set-up default for `enrollment-server.user-info.provider`

docker/enrollment-server.xml
```diff
@@ -40,8 +40,7 @@
     <Parameter name="enrollment-server.activation-spawn.enabled" value="${ENROLLMENT_SERVER_ACTIVATION_SPAWN_ENABLED}"/>
 
     <!--  User-info configuration  -->
-    <Parameter name="enrollment-server.user-info.minimal-provider.enabled" value="${ENROLLMENT_SERVER_USER_INFO_MINIMAL_ENABLED:false}"/>
-    <Parameter name="enrollment-server.user-info.rest-provider.enabled" value="${ENROLLMENT_SERVER_USER_INFO_REST_ENABLED:false}"/>
+    <Parameter name="enrollment-server.user-info.provider" value="${ENROLLMENT_SERVER_USER_INFO_PROVIDER:}"/>
```
zcgandcomp, line 43: Last time the value without default crash the app during the startup. Please test it or put a default here.

zcgandcomp (changes requested):

docker/enrollment-server.xml
banterCZ: I would like to see the error message. This is legit and working. It means no user-info provider at all. I have tested that. `${ENROLLMENT_SERVER_USER_INFO_PROVIDER:}` has a default value, which is empty.

romanstrobl (approved):
Looks OK.

zcgandcomp (approved):

PR #731, opened 2023-04-20
Fix #728: Missing template for operation results in 500 for mobile

### LINKED ISSUE #728: Missing template for operation results in 500 for mobile
The mobile operation list and history API will return 500 in cases when the operation template is no longer available.

Steps to reproduce (for history):

1. Create an operation with a specific template
2. Remove the template
3. Call history API (from mobile)
4. Get 500

The same can be done for the operation list

1. Create an operation with a specific template
2. Remove the template
3. Call operation list API (from mobile)
4. Get 500

banterCZ:


romanstrobl (approved):
Good to have this check.

PR #733, opened 2023-04-20
Fix #732: Replace collect(Collectors.toList()) by toList()

### LINKED ISSUE #732: Replace collect(Collectors.toList()) by toList()
Possible after update to JDK 17.

banterCZ:


romanstrobl (approved):
Nice simplification.

PR #735, opened 2023-04-21
Fix #734: Ignore missing / null URL value in params when building image template

### LINKED ISSUE #734: Ignore missing / null URL value in params when building image template
For

```
POST /v2/operations    

"parameters": {
        "MMI_icon": "https://famousfoxes.com/hd/5586.png"       
}
```

We should get (without `originalUrl`).

```json
{
  "type":"IMAGE",
  "id":"operation.image",
  "label":"",
  "thumbnailUrl":"https://famousfoxes.com/hd/5586.png"
}
```

banterCZ:


romanstrobl (approved):
Looks OK.

PR #737, opened 2023-04-21
Fix #736: Fewer log arguments than placeholders in OnboardingProcessEntityWrapper

### LINKED ISSUE #736: Fewer log arguments than placeholders in OnboardingProcessEntityWrapper

banterCZ:


romanstrobl (approved):
Good catch.

PR #739, opened 2023-04-21
Fix #738: Review Java warnings

### LINKED ISSUE #738: Review Java warnings
Review Java warnings specifically after upgrading to JDK 17.

banterCZ:


enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentResultEntity.java
```diff
@@ -113,8 +115,7 @@ public class DocumentResultEntity implements Serializable {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (!(o instanceof DocumentResultEntity)) return false;
-        DocumentResultEntity that = (DocumentResultEntity) o;
+        if (!(o instanceof DocumentResultEntity that)) return false;
```
romanstrobl, line 118: We can add `final` keyword here, I think.

romanstrobl (approved):
Nice improvements, I have one recommendation related to use of `instanceof`.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentResultEntity.java
banterCZ: Ok, why not.

PR #754, opened 2023-05-15
Read real image at WultraMockDocumentVerificationProvider to make ImageProcessor working

banterCZ:
Follow-up to #527


romanstrobl (approved):
Looks OK.

PR #756, opened 2023-05-16
Fix #755: Coverity - Dereference null return value by ImageProcessor

### LINKED ISSUE #755: Coverity - Dereference null return value by ImageProcessor
Calling a method on null object `bufferedSourceImage`.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #758, opened 2023-05-17
Fix #745: Cancel OTP marks VERIFIED entries as FAILED

### LINKED ISSUE #745: Cancel OTP marks VERIFIED entries as FAILED
`OtpServiceImpl#cancelOtp` (called by `cleanup`) marks entries as FAILED, even verified. In case of a successful process, data looks strange.

**Hold this issue**, reporting may be based on this behavior.

https://github.com/wultra/enrollment-server/blob/424d60b276ac941bfad4a2ced53ea222bdf309ba/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OtpServiceImpl.java#L132

romanstrobl:


petrdvorak (approved):
Looks good

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OtpServiceImpl.java
```diff
@@ -131,17 +131,17 @@ public String createOtpCodeForResend(OnboardingProcessEntity process, OtpType ot
      */
     public void cancelOtp(OnboardingProcessEntity process, OtpType otpType) {
         final String processId = process.getId();
-        // Fail current OTP, if it is present
+        // Fail current OTP, if it is present and in ACTIVE state
         onboardingOtpRepository.findNewestByProcessIdAndType(processId, otpType).ifPresent(otp -> {
             final Date now = new Date();
-            if (otp.getStatus() != OtpStatus.FAILED) {
+            if (otp.getStatus() == OtpStatus.ACTIVE) {
```
banterCZ, line 137: Ok, I have double-checked that we have just three states and just one _in-progress_, so we do not have to introduce a constant list of statuses yet.

banterCZ (approved):
My only objection is about customers' expectation and their data consistency, otherwise perfect.

PR #763, opened 2023-05-18
Fix #760: Add error_detail for OTP resend

### LINKED ISSUE #760: Add error_detail for SMS OTP canceled due to new SMS OTP request
When a SMS OTP is canceled due to new SMS OTP request from user, the `error_detail` is null. We should add a new `error_detail` value for this case.

romanstrobl:


banterCZ (approved):
Thanks for the fix.

PR #767, opened 2023-05-23
Fix #766: Reconsider level of logging at IProovRestApiService, ZenidRestApiService

### LINKED ISSUE #766: Reconsider level of logging at IProovRestApiService, ZenidRestApiService
Sensitive data should be moved to the TRACE level.

banterCZ:


romanstrobl (approved):
Thanks for the fix.

PR #770, opened 2023-05-30
Fix #752: Optimize iProov calls

### LINKED ISSUE #752: Optimize iProov calls
According to the statistics, iProov `enroll` is called quite more often than `verify`

One possible reason
https://github.com/wultra/enrollment-server/blob/4b45f675b18b95d79a9a82875591333762220762/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/provider/IProovPresenceCheckProvider.java#L82

banterCZ:


enrollment-server-onboarding/src/main/resources/api/api-iproov.json
```diff
@@ -1032,21 +1027,6 @@
...
-          "description": "The URL to redirect to on failure. Note: This field has been deprecated and will be omitted in the next release.",
-          "type": "string",
-          "deprecated": true
-        },
-        "abort_url": {
-          "description": "The URL to redirect to on user abort. Note: This field has been deprecated and will be omitted in the next release.",
-          "type": "string",
-          "deprecated": true
-        },
         "user_id": {
```
banterCZ, line 1030: Manually edited OpenAPI, when download https://eu.rp.secure.iproov.me/swagger.json you get this. But confirmed by iProov, that is not data validation but only recommendation.

```json
"type": "string",
"format": "uuid"
```

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/provider/IProovPresenceCheckProvider.java
```diff
@@ -77,31 +79,26 @@ public IProovPresenceCheckProvider(
...
+        final ResponseEntity<String> responseEntityToken = callGenerateEnrolToken(id);
 
-        if (responseEntityToken.getBody() == null) {
+        final String body = responseEntityToken.getBody();
+        if (body == null) {
             throw new RemoteCommunicationException("Missing response body when generating an enrol token in iProov, " + id);
         }
 
+        // Deleting the iProov enrollment properly is not implemented yet
+        if (body.contains("already_enrolled")) {
```
zcgandcomp, line 91: Simply raising the exception will deadlock the process. 
We need to:
- If the photo was enrolled in this process, we could move to the verification cycle. (if photo enroll fails on our side to properly update the process)
- If the photo was enrolled in this process, but the whole cycle is repeated (i.e. new documents)
- If this is a new process, we need to get the enrolled token from previous attempt (or use alternated userId)

There could be controls on other parts of the process to ensure this will not happened, but I thing currently there are not.

zcgandcomp (changes requested):

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/provider/IProovPresenceCheckProvider.java
banterCZ: This should never happen, just a handbrake. The real check is in the layer above https://github.com/wultra/enrollment-server/pull/770/files#diff-ee0512b663943b0bbb18ad44c1551c55261a422493b7736049cd0ada225027f7R230

banterCZ: 🤔 But yes, I have to double-check `userId`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
```diff
@@ -224,6 +214,38 @@ public void cleanup(OwnerId ownerId) throws PresenceCheckException, RemoteCommun
...
+     * @param idVerification Verification identity.
+     * @throws DocumentVerificationException When not able to find documet image.
+     * @throws PresenceCheckException In case of business logic error.
+     * @throws RemoteCommunicationException In case of remote communication error.
+     */
+    private void initPresentCheckWithImage(final OwnerId ownerId, final IdentityVerificationEntity idVerification)
+            throws PresenceCheckException, DocumentVerificationException, RemoteCommunicationException {
+
+        if (!idVerification.isPresenceCheckInitialized()) {
+            if (StringUtils.hasText(idVerification.getSessionInfo())) {
```
romanstrobl, line 230: We should check the content of the `sessionInfo`, not only this presence, because this record is shared among all parts of the identity verification process. Currently we only use it for iProov, but that could change in the future.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/provider/IProovPresenceCheckProvider.java
banterCZ: - [ ] Improve handling `already_enrolled` based on the answer from iProov support.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
```diff
@@ -224,6 +217,38 @@ public void cleanup(OwnerId ownerId) throws PresenceCheckException, RemoteCommun
...
+     * @param idVerification Verification identity.
+     * @throws DocumentVerificationException When not able to find documet image.
+     * @throws PresenceCheckException In case of business logic error.
+     * @throws RemoteCommunicationException In case of remote communication error.
+     */
+    private void initPresentCheckWithImage(final OwnerId ownerId, final IdentityVerificationEntity idVerification)
+            throws PresenceCheckException, DocumentVerificationException, RemoteCommunicationException {
+
+        if (!idVerification.isPresenceCheckInitialized()) {
+            if (wasImageAlreadyUploaded(idVerification)) {
```
romanstrobl, line 233: The method name seems strange, perhaps `imageAlreadyUploaded` is better?

enrollment-server-onboarding/src/main/resources/application.properties
```diff
@@ -149,6 +149,8 @@ enrollment-server-onboarding.presence-check.iproov.ensureUserIdValueEnabled=${IP
 enrollment-server-onboarding.presence-check.iproov.riskProfile=${IPROOV_RISK_PROFILE:}
 enrollment-server-onboarding.presence-check.iproov.serviceBaseUrl=${IPROOV_SERVICE_BASE_URL}
 enrollment-server-onboarding.presence-check.iproov.serviceUserAgent=Wultra/OnboardingServer
+enrollment-server-onboarding.presence-check.iproov.oAuthUsername=${IPROOV_OAUTH_USERNAME:}
```
romanstrobl, line 152: I would prefer `oauthClientUsername` and `oauthClientPassword` to make sure it is clear this is used for client authentication during OAuth.

romanstrobl (approved):
Good job, I have just two comments related to naming.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
```diff
@@ -341,9 +366,27 @@ private IdentityVerificationEntity fetchIdVerification(OwnerId ownerId) throws P
...
+     * @return true if image was already uploaded
+     */
+    private boolean wasImageAlreadyUploaded(final IdentityVerificationEntity identityVerification) {
+        final String sessionInfoString = identityVerification.getSessionInfo();
+        if (!StringUtils.hasText(sessionInfoString)) {
+            return false;
+        }
+        final SessionInfo sessionInfo = jsonSerializationService.deserialize(sessionInfoString, SessionInfo.class);
+        // TODO (racansky, 2023-05-31) improve logic to something more robust than presence of the attribute `timestampLastUsed`
+        return sessionInfo != null && !CollectionUtils.isEmpty(sessionInfo.getSessionAttributes()) && sessionInfo.getSessionAttributes().containsKey(SESSION_ATTRIBUTE_TIMESTAMP_LAST_USED);
```
romanstrobl, line 390: Is there any reason why we just don't use some new property e.g. `imageUploaded`:`true` for this?

banterCZ: 😞 Indeed, I have a TODO there. Let's improve it later.

zcgandcomp: We need to improve it now. It is not clear how the session use attribute is propagated and if this is relevant for the image upload at all.

zcgandcomp (changes requested):

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
banterCZ: Ok, I have refactored the way how we work with the `SessionInfo`.

zcgandcomp (commented):
I assume the identityVerification is newly created for each verification loop (i.e. in the very first and with each SCA fail).

In this verification loop, I believe the verification of user existence can be done by calling enrolling endpoint - the whole OAUTH authorization can be invoked only if we need to delete user - we will reduce the number of calls to iProov by 2 for image upload.

banterCZ:
> I assume the identityVerification is newly created for each verification loop (i.e. in the very first and with each SCA fail).
> 
> In this verification loop, I believe the verification of user existence can be done by calling enrolling endpoint - the whole OAUTH authorization can be invoked only if we need to delete user - we will reduce the number of calls to iProov by 2 for image upload.

Your assumption is correct but existence check and deletion at management API are done just once, see the link to the logic. OAuth authorization is done after 50 calls or 1 hour, but this is done out of the box by the filter.

https://github.com/wultra/enrollment-server/pull/770/files#diff-ee0512b663943b0bbb18ad44c1551c55261a422493b7736049cd0ada225027f7R236

banterCZ:
iProov have declared that the TPS rate limit is only applied to `POST /claim/enrol/token` and `POST /claim/verify/token`, not the management API.

zcgandcomp (approved):
Satisfactory.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/config/IProovConfig.java
```diff
@@ -67,4 +93,68 @@ public RestClient restClientIProov(IProovConfigProps configProps) throws RestCli
         return new DefaultRestClient(restClientConfiguration);
     }
 
+
+    @Bean
```
petrdvorak, line 97: Maybe too soon but we should qualify the bean in case we have more than one provider using OAuth 2.0...

petrdvorak (approved):
Looks OK for now, I added one minor comment that is irrelevant now.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/config/IProovConfig.java
banterCZ: Qualifier is already used, `@Qualifier("iproovManagemenentWebClient") WebClient webClient) {` https://github.com/wultra/enrollment-server/pull/770/files#diff-2aa3940816a2a6154066d5fa9310d3f5e4e04b5dcb874f225992f34e1d5e4124R108

petrdvorak: I see, resolved then.

PR #786, opened 2023-06-14
Fix #761: Invalid presence check limit check

### LINKED ISSUE #761: Invalid presence check limit check
The presence check limit is checked using SMS OTP count (default limit: 5). This is problematic, because the user can resend a SMS OTP, and then the number of presence check attempts does not correspond to the SMS OTP count. This manifests in the data that some users have up to 8 failed SMS OTPs.

We should consider implementing a different strategy, e.g. using a separate table which stores all presence check attempts and their results, including the SMS OTP result because it belongs to the same SCA verification step.

banterCZ:


docs/db/changelog/changesets/enrollment-server-onboarding/1.5.x/20230614-add-sca-result.xml
```diff
@@ -0,0 +1,52 @@
...
+            <not>
+                <tableExists tableName="es_sca_result"/>
+            </not>
+        </preConditions>
+        <comment>Create a new table es_onboarding_process</comment>
+        <createTable tableName="es_sca_result">
+            <column name="id" type="bigint">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="identity_verification_id" type="varchar(36)">
```
romanstrobl, line 27: I suggest we also add `process_id` for simpler joins.

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-1.5.0.md
```diff
@@ -75,6 +75,53 @@ ALTER TABLE es_onboarding_process
...
 
+### SCA Result
+
+A new table `es_sca_result` has been created.
+
+
+#### PostgreSQL
+
+```sql
+CREATE SEQUENCE es_sca_result_seq INCREMENT BY 50 START WITH 1 CACHE 2;
```
romanstrobl, line 86: What is the point of caching 2 sequence IDs?

romanstrobl (approved):
Looks nice, I found two minor issues.

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-1.5.0.md
banterCZ: Let's drop the explicit CACHE size and use the default.

docs/db/changelog/changesets/enrollment-server-onboarding/1.5.x/20230614-add-sca-result.xml
banterCZ: Ok, why not.

romanstrobl (approved):
Great, thanks!

PR #790, opened 2023-06-19
Fix #789: Improve logging of calling iProov

### LINKED ISSUE #789: Improve logging of calling iProov

banterCZ:
- Log error even if it is not instance of WebClientResponseException
- Log access token at TRACE level

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/config/IProovConfig.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/config/IProovConfig.java
@@ -49,6 +49,7 @@ import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.web.reactive.function.client.ClientResponse;
 import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
 import org.springframework.web.reactive.function.client.WebClient;
+import org.springframework.web.reactive.function.client.WebClientResponseException;
 import org.springframework.web.util.UriComponentsBuilder;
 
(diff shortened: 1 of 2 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/config/IProovConfig.java
```diff
@@ -136,16 +137,24 @@ private static AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager auth
...
-                    logger.debug("Removing scope because does not comply with RFC and not needed anyway");
-                    map.remove(AuthTokenResponse.JSON_PROPERTY_SCOPE);
-                    return builder.body(JSONObject.toJSONString(map)).build();
-                } else {
-                    return builder.build();
-                }
-            });
+        final ClientResponse.Builder builder = response.mutate();
+        return response.bodyToMono(Map.class).map(map -> {
+                    logger.trace("Got access token={}", map.get(AuthTokenResponse.JSON_PROPERTY_ACCESS_TOKEN));
```
banterCZ, line 142: 👀 Consider this log from the security point of view. It is the `TRACE` level.

romanstrobl (changes requested):
Let's not log whole access tokens, even in `TRACE` level.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/presencecheck/iproov/config/IProovConfig.java
banterCZ: I have removed the value completely.

PR #792, opened 2023-06-21
Fix #791: OnboardingProcessRepository may cause follow-on locking

### LINKED ISSUE #791: OnboardingProcessRepository may cause follow-on locking
`ORDER BY` is not needed https://github.com/wultra/enrollment-server/blob/7f9cb620a5776c3156912ba4baa551d16ad1e1fb/enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java#L94

banterCZ:


romanstrobl (approved):
Good catch!

PR #795, opened 2023-06-22
Fix #794: Set up timeout for iProov OAuth2 and user-management

### LINKED ISSUE #794: Set up timeout for iProov OAuth2 and user-management
Follow-up to #752

banterCZ:


romanstrobl (approved):
Perfect, the timeout handling was missing.

PR #804, opened 2023-07-12
Fix #353: Add request validation for new request / response classes

### LINKED ISSUE #353: Add request validation for new request / response classes
Use bean validation for simple validations, implement validators for complex logic.

jandusil:
- Init draft commit
- Add simple bean validators


petrdvorak (commented):
Not sure why this is required... The classes you modified are actually already validated (see `ActivationCodeRequestValidator` and `PushRegisterRequestValidator`) so the change does not add any value.

Looking at the code, we should also add a handler method for validation exceptions as with the current format, we might miss some opportunities for good error handling. ;-)

cc @romanstrobl

jandusil:
@petrdvorak Sorry this one is still a draft

PR #808, opened 2023-07-20
Fix #807: fds_data should be nullable in postgresql/create-schema.sql

### LINKED ISSUE #807: fds_data should be nullable in postgresql/create-schema.sql
Remove `NOT NULL` from
```sql
fds_data TEXT NOT NULL
```

banterCZ:


romanstrobl (approved):
Looks OK.

PR #813, opened 2023-07-25
Fix #812: Revert downgrade of enforcer plugin

### LINKED ISSUE #812: Revert downgrade of enforcer plugin
Make the version of the enforcer plugin implicit again (from the Spring Boot parent).

banterCZ:


romanstrobl (approved):
Looks OK.

PR #818, opened 2023-08-01
Fix #817: Configure proxy for iProov user management

### LINKED ISSUE #817: Configure proxy for iProov user management
A follow-up to #752

banterCZ:


petrdvorak (approved):
Looks OK

PR #828, opened 2023-08-10
Fix #827: Expose OTP resent period

### LINKED ISSUE #827: Expose OTP resent period
OTP resent period is configurable and may be changed. Mobile application needs to configure it as well. But these two values may be out of sync. Expose the value at REST API, where the mobile application may take it.

See `enrollment-server-onboarding.onboarding-process.otp.resend-period:30s`

jandusil:
- Update ConfigurationDataDto to expose otpResendPeriodSeconds
- Expose the Dto to responses of api/onboarding/start and api/onboarding/status endpoints
- Fill the config in services

### DIFF
--- a/enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/OnboardingStartResponse.java
+++ b/enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/OnboardingStartResponse.java
@@ -31,6 +31,6 @@ public class OnboardingStartResponse {
 
     private String processId;
     private OnboardingStatus onboardingStatus;
-    private ConfigurationDataDto configurationDataDto;
+    private ConfigurationDataDto config;
 
 }
--- a/enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/OnboardingStartResponse.java
+++ b/enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/OnboardingStartResponse.java
@@ -17,6 +17,7 @@
  */
 package com.wultra.app.enrollmentserver.api.model.onboarding.response;
 
+import com.wultra.app.enrollmentserver.api.model.onboarding.response.data.ConfigurationDataDto;
 import com.wultra.app.enrollmentserver.model.enumeration.OnboardingStatus;
 import lombok.Data;
 
@@ -30,5 +31,6 @@ public class OnboardingStartResponse {
 
     private String processId;
     private OnboardingStatus onboardingStatus;
+    private ConfigurationDataDto configurationDataDto;
 
 }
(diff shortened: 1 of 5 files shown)


enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/data/ConfigurationDataDto.java
```diff
@@ -31,4 +31,9 @@ public class ConfigurationDataDto {
      */
     private String otpResendPeriod;
```
banterCZ, line 32: We may mark this property as deprecated, create an issue for removal and link it from here something like `// TODO (dusil, 2023-08-10, #xxx) some explanation`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -121,6 +127,9 @@ public OnboardingServiceImpl(
         this.otpService = otpService;
         this.activationService = activationService;
         this.onboardingProvider = onboardingProvider;
+        this.integrationConfigDto = new ConfigurationDataDto();
```
banterCZ, line 130: You may introduce a builder if you want so.

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/OnboardingStartResponse.java
```diff
@@ -30,5 +31,6 @@ public class OnboardingStartResponse {
 
     private String processId;
     private OnboardingStatus onboardingStatus;
+    private ConfigurationDataDto configurationDataDto;
```
banterCZ, line 34: The property name is not consistent with `IdentityVerificationStatusResponse`, where it is only `config`.

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/OnboardingStatusResponse.java
```diff
@@ -30,5 +31,6 @@ public class OnboardingStatusResponse {
 
     private String processId;
     private OnboardingStatus onboardingStatus;
+    private ConfigurationDataDto configurationDataDto;
```
banterCZ, line 34: dtto

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -84,6 +85,11 @@ public class OnboardingServiceImpl extends CommonOnboardingService {
 
     private final ActivationService activationService;
 
+    /**
+     * Configuration data for client integration
+     */
+    private final ConfigurationDataDto integrationConfigDto;
```
banterCZ, line 91: Ok, IMHO it is quite unusual but why not. In theory, this makes it impossible to reload properties in runtime, which we do not use anyway. Just a note to realize.

banterCZ (commented):
Looks good from the backend point of view, except consistency of property naming. Also, let's wait for an opinion from mobile developers.

kober32 (approved):
OK with me.

PR #830, opened 2023-08-14
Fix #773: Upload presence check photo for failed authentication attempts

### LINKED ISSUE #773: Upload presence check photo for failed authentication attempts
When the presence check fails, upload the photo even for failed authentication attempts due to the requirements for fraud verification.

The issue is related to: https://github.com/wultra/enrollment-server/issues/761, we should store SCA results correctly, because the photo upload is now used for checking that SCA step succeeded.

jandusil:
- Modified the checkPresenceVerification method to handle photo uploads irrespective of the presence check result.

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
@@ -134,13 +134,14 @@ public class PresenceCheckService {
         final PresenceCheckResult result = presenceCheckProvider.getResult(ownerId, sessionInfo);
         auditService.auditPresenceCheckProvider(idVerification, "Got presence check result: {} for user: {}", result.getStatus(), ownerId.getUserId());
 
+        // Evaluate the result
         if (result.getStatus() != PresenceCheckStatus.ACCEPTED) {
             logger.info("Not accepted presence check, status: {}, {}", result.getStatus(), ownerId);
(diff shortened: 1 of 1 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
```diff
@@ -134,13 +134,14 @@ public void checkPresenceVerification(
         final PresenceCheckResult result = presenceCheckProvider.getResult(ownerId, sessionInfo);
         auditService.auditPresenceCheckProvider(idVerification, "Got presence check result: {} for user: {}", result.getStatus(), ownerId.getUserId());
 
+        // Evaluate the result
         if (result.getStatus() != PresenceCheckStatus.ACCEPTED) {
             logger.info("Not accepted presence check, status: {}, {}", result.getStatus(), ownerId);
-            evaluatePresenceCheckResult(ownerId, idVerification, result);
-            return;
+        } else {
```
petrdvorak, line 140: Does the code need to drop the `evaluatePresenceCheckResult(ownerId, idVerification, result);` line? I understand we do not want to return and want to continue uploading the photo, but we can still proceed with the business logic (which we would miss in the case the code throws on `if (photo == null)` line)... or alternatively, we should add the `evaluatePresenceCheckResult` call there...

jandusil: Absolutely true, thanks for the comment

petrdvorak (approved):
Looks OK

PR #856, opened 2023-09-07
Fix #773: Upload presence check photo for failed authentication attempts

### LINKED ISSUE #773: Upload presence check photo for failed authentication attempts

banterCZ:
- Parse frame even if the presence check failed


romanstrobl (approved):
Looks OK.

PR #866, opened 2023-09-19
Fix #777: Add column for fake OTP delivery

### LINKED ISSUE #777: Add column for fake OTP delivery
In case a fake OTP record is created, it cannot be differentiated from regular OTP records. We should add a column which identifies such OTP rows.

banterCZ:


romanstrobl (approved):
Looks good.

PR #874, opened 2023-09-26
Fix #873: Enrollment server fails on NPE when operation parameter is …

### LINKED ISSUE #873: Enrollment server fails on NPE when operation parameter is null
PAC allows 
Steps to replicate
1) create operation with parameter set to null
2) call endpoint for pending operation

Expected result:
list is returned, null parameter is omited.

Actual result:
Error

Technical notes:
Examle of operation parameters:
```
{
  "amount" : "100.00",
  "currency" : "EUR",
  "iban" : null
}
```

stacktrace:

```
2023-09-20T10:50:07.989953252Z java.lang.NullPointerException: null
2023-09-20T10:50:07.989958152Z  at java.base/java.util.Objects.requireNonNull(Unknown Source) ~[na:na]
2023-09-20T10:50:07.989963052Z  at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Unknown Source) ~[na:na]
2023-09-20T10:50:07.989967752Z  at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(Unknown Source) ~[na:na]
2023-09-20T10:50:07.989972252Z  at java.base/java.util.Iterator.forEachRemaining(Unknown Source) ~[na:na]
2023-09-20T10:50:07.989977052Z  at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Unknown Source) ~[na:na]
2023-09-20T10:50:07.989981552Z  at java.base/java.util.stream.AbstractPipeline.copyInto(Unknown Source) ~[na:na]
2023-09-20T10:50:07.989986052Z  at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source) ~[na:na]
2023-09-20T10:50:07.989990452Z  at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source) ~[na:na]
2023-09-20T10:50:07.989995052Z  at java.base/java.util.stream.AbstractPipeline.evaluate(Unknown Source) ~[na:na]
2023-09-20T10:50:07.990012153Z  at java.base/java.util.stream.ReferencePipeline.collect(Unknown Source) ~[na:na]
2023-09-20T10:50:07.990016953Z  at com.wultra.app.enrollmentserver.impl.service.converter.MobileTokenConverter.createStringSubstitutor(MobileTokenConverter.java:152) ~[classes/:1.4.2]
```

jandusil:
…null

- Fix omitting nulls

### DIFF
--- a/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
+++ b/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
@@ -142,11 +142,12 @@ public class MobileTokenConverter {
         return formData;
     }
 
-    private static StringSubstitutor createStringSubstitutor(final Map<String, String> parameters) {
+    static StringSubstitutor createStringSubstitutor(final Map<String, String> parameters) {
         if (CollectionUtils.isEmpty(parameters)) {
             return null;
         } else {
             final Map<String, String> escapedParameters = parameters.entrySet().stream()
+                    .filter(entry -> entry.getValue() != null)
                     .collect(toMap(Map.Entry::getKey, it -> StringEscapeUtils.escapeJson(it.getValue())));
             return new StringSubstitutor(escapedParameters);
         }
--- a/enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
+++ b/enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
@@ -20,15 +20,18 @@ package com.wultra.app.enrollmentserver.impl.service.converter;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableMap;
 import com.wultra.app.enrollmentserver.database.entity.OperationTemplateEntity;
+import com.wultra.app.enrollmentserver.errorhandling.MobileTokenConfigurationException;
 import com.wultra.security.powerauth.client.model.enumeration.OperationStatus;
 import com.wultra.security.powerauth.client.model.enumeration.SignatureType;
 import com.wultra.security.powerauth.client.model.response.OperationDetailResponse;
 import com.wultra.security.powerauth.lib.mtoken.model.entity.*;
 import com.wultra.security.powerauth.lib.mtoken.model.entity.attributes.*;
+import org.apache.commons.text.StringSubstitutor;
 import org.junit.jupiter.api.Test;
 import org.springframework.context.i18n.LocaleContextHolder;
 
 import java.math.BigDecimal;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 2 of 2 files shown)


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -142,11 +142,12 @@ private FormData prepareFormData(
         return formData;
     }
 
-    private static StringSubstitutor createStringSubstitutor(final Map<String, String> parameters) {
+    static StringSubstitutor createStringSubstitutor(final Map<String, String> parameters) {
```
banterCZ, line 145: May we test it without making it _package private_? According to the test coverage, current tests go through both branches.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -142,11 +142,12 @@ private FormData prepareFormData(
...
         if (CollectionUtils.isEmpty(parameters)) {
             return null;
         } else {
             final Map<String, String> escapedParameters = parameters.entrySet().stream()
+                    .filter(entry -> entry.getValue() != null)
```
banterCZ, line 150: @zcgandcomp I am not sure what is the goal of this issue. Should we filter out `null` values as suggested in this PR or fix NPE and include it? If so, @jandusil see https://stackoverflow.com/a/24634007/204950

banterCZ (changes requested):
We should clarify first, what to do with `null` values whether to skip or include them.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
zcgandcomp: The default behavior is that if the value is omitted the key is not sent to the mobile phone. So you can "skip" the null values. 
Reason:
- They are not possible to display as of now anyway.
- The mobile app can have a problem with processing another null.

banterCZ: Ok, this implementation skips `null` values.

banterCZ:
Ok, we have clarified, that we may skip `null` values. But still, I would like to keep `createStringSubstitutor` _private_, if possible.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
```diff
@@ -592,6 +594,47 @@ void testConvertUiPreapprovalScanQrSuppressed() throws Exception {
...
+                    }
+                  }
+                ]""");
+
+        final Operation operation = tested.convert(operationDetail, operationTemplate);
+        final List<Attribute> attributes = operation.getFormData().getAttributes();
+
+
+        assertNotNull(operation);
+        assertNotNull(attributes);
```
banterCZ, line 633: To avoid NPE, it is needed to reorder.
```suggestion
        assertNotNull(operation);
        final List<Attribute> attributes = operation.getFormData().getAttributes();

        assertNotNull(attributes);
```

banterCZ (approved):
Looks good. One suggestion for the test.

romanstrobl (commented):
Good solution.

PR #875, opened 2023-09-26
Revert "Fix #777: Add column for fake OTP delivery"

### LINKED ISSUE #777: Add column for fake OTP delivery

banterCZ:
This reverts commit 3e3c4f404c1df526057632199c3a97c9ba04689a.


zcgandcomp (approved):
Looks OK

PR #883, opened 2023-10-05
Fix #882: Filter jboss descriptor with bouncycastle version from pom

### LINKED ISSUE #882: Filter jboss descriptor with bouncycastle version from pom

banterCZ:


romanstrobl (approved):
Looks OK.

Issue #891, opened 2023-10-20
Change defaulting of locales for template finding


zcgandcomp:
Currently, if a template is not found for a locale, the English version is used as a default.

This causes an issue when the English version is not set.

Modify the logic - if the template is not found for the requested locale, return any existing template.

banterCZ:
Let's make the order in this way
1) Given language
2) English
3) Any other found language

OliverJanstaWultra:
QA: Create templates that are not in English and try them.

kopovan:
QA OK, the criteria have been met as described above. Tested od PACE, screenshots attached:
<img src="https://github.com/wultra/enrollment-server/assets/135111051/93c78d91-956e-49e0-bb6b-ef7177695b06" width="300" /> <img src="https://github.com/wultra/enrollment-server/assets/135111051/a461fda5-a94d-44bd-81e3-59b555548387" width="300" /> <img src="https://github.com/wultra/enrollment-server/assets/135111051/627c36d6-f330-4954-acbb-a2a4bd74b2d8" width="300" />

PR #895, opened 2023-10-25
Optimize defaulting of locales for template finding

banterCZ:
A follow-up to #891


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/OperationTemplateService.java
```diff
@@ -44,20 +46,36 @@ public OperationTemplateService(OperationTemplateRepository operationTemplateRep
...
-                return operationTemplateRepository.findFirstByPlaceholder(operationType);
-            });
-        });
+            return findEnglishTemplate(operationType);
+        } else {
+            return findAnyTemplate(operationType);
+        }
+    }
+
+    private Optional<OperationTemplateEntity> findEnglishTemplate(final String operationType) {
```
romanstrobl, line 71: Can we call this method `findDefaultTemplate` instead of `findEnglishTemplate`? The variable name is `DEFAULT_LANGUAGE`, so it will correspond nicely.

romanstrobl (approved):
Looks OK, just one small suggestion for naming.

PR #901, opened 2023-10-30
Fix #896: Use different error when PAS operation approve fails

### LINKED ISSUE #896: Use different error when PAS operation approve fails
Discovered during the development by @Hopsaheysa 

Right now, when a client sends a Proximity OTP that is expired or just wrong, the server returns:

```json
{"status":"ERROR","responseObject":{"code":"POWERAUTH_AUTH_FAIL","message":"Authentication failed"}}
```

This is the same error as WRONG PASSWORD, so as of now, the frontend cannot differentiate these cases which leads to a very misleading UX in the app.

We need to be able to know when the PAS _operation approval_ fails. For this use case let's introduce a new error code `OPERATION_FAILED`. 

An invalid signature is already covered by 

```java
@PostMapping("/operation/authorize")
@PowerAuth(resourceId = "/operation/authorize", signatureType = {
        PowerAuthSignatureTypes.POSSESSION,
        PowerAuthSignatureTypes.POSSESSION_KNOWLEDGE,
        PowerAuthSignatureTypes.POSSESSION_BIOMETRY
})
public Response operationApprove(
```

banterCZ:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/MobileTokenService.java
```diff
@@ -181,8 +181,8 @@ public Response operationApprove(@NotNull final OperationApproveParameterObject
             return new Response();
         } else {
             final OperationDetailResponse operation = approveResponse.getOperation();
-            handleStatus(operation.getStatus());
-            throw new MobileTokenAuthException();
+            handleStatus(operation);
+            throw new MobileTokenAuthException(ErrorCode.OPERATION_FAILED, "PowerAuth server operation approval fails");
```
banterCZ, line 185: 👀

docs/Mobile-Token-API.md
```diff
@@ -75,15 +75,16 @@ Mobile token API provides access to operations.
...
+| `OPERATION_ALREADY_CANCELED` | Operation is already canceled                                                                                                                                                                                                                                       | 400              |
+| `OPERATION_EXPIRED`          | Operation is expired                                                                                                                                                                                                                                                | 400              |
+| `OPERATION_FAILED`           | PowerAuth server operation approval fails.                                                                                                                                                                                                                          | 401              |
```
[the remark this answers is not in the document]
kober32: Now when I'm looking at it, I'm not sure if 401 is a good fit. Does it really mean `Unauthorized`? I think that 400 would be more suitable 🤔

kober32 (approved):
OK with me

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/MobileTokenService.java
```diff
@@ -262,7 +262,7 @@ public Response operationReject(
             return new Response();
         } else {
             final OperationDetailResponse operation = rejectResponse.getOperation();
-            handleStatus(operation.getStatus());
+            handleStatus(operation);
```
zcgandcomp, line 265: and should not we be consistent with raising OPERATION_FAILED as default when an operation failed?

zcgandcomp (approved):
Looks OK.

PR #924, opened 2023-11-21
Fix #923: Introduce maven modules enrollment-onboarding-api, -zenid, and -iproov

### LINKED ISSUE #923: Introduce maven modules enrollment-onboarding-api, -zenid, and -iproov
Introduce maven modules `enrollment-onboarding-api`, `enrollment-onboarding-zenid`, and `enrollment-onboarding-iproov`. It is important to keep order when adding other providers such as Innovatrics and not keep all in i single pile.

Innovatrics

banterCZ:


banterCZ:
> There are problems with running tests from `.providers.*` modules.

You mean for example `enrollment-server-onboarding-providers-iproov/src/test/java/com/wultra/app/onboardingserver/providers/iproov/IProovPresenceCheckProviderTest.java`? It is a historical integration test to interact with a real running service. We do not want to commit secrets and URLs. The exclusion from the suite for maven is done this way.
```xml
<plugin>
     <groupId>org.apache.maven.plugins</groupId>
     <artifactId>maven-surefire-plugin</artifactId>
     <configuration>
     <excludedGroups>external-service</excludedGroups>
     </configuration>
</plugin>
```

pom.xml
```diff
@@ -114,7 +117,25 @@
...
+
+            <dependency>
+                <groupId>com.wultra.security</groupId>
+                <artifactId>enrollment-server-onboarding-providers-iproov</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.wultra.security</groupId>
+                <artifactId>enrollment-server-onboarding-providers-zenid</artifactId>
```
banterCZ, line 138: ❓ Should be module and package singular or plural `provider` or `providers`.

jnpsk (approved):
Nice solution for having different providers.

romanstrobl (approved):
I agree.

pom.xml
banterCZ: Ok, let's make it singular.

PR #941, opened 2023-12-06
Fix #911: Expose API to upload presence check data

### LINKED ISSUE #911: Expose API to upload presence check data
When using the iProov provider for presence check (liveness), the mobile application connects directly to the public iProov API. In the case of Innovatrics, the Onboarding server should proxy the private API. Design and implement the endpoint.

Relates to #908

banterCZ:


romanstrobl (approved):
I did not notice any issues, looks good.

PR #957, opened 2023-12-15
Fix #956: Respect provider name in queries

### LINKED ISSUE #956: Respect provider name in queries
When querying `DocumentVerificationRepository` the `providerName` is totally ignored. This is not an issue when using a single provider at a time. But it causes unexpected behaviour when using multiple providers on the same database.

`DocumentVerificationRepository`
- `failVerifications`: by activationID, triggered only from an API call. Does not need to specify provider.
- `findExpiredVerifications`: does not cause errors, but better to specify provider, so a server instance cleans only his documents.
- `setOtherDocumentSide`: update by an ID. Does not need to specify provider.
- 🔴  `findAllDocumentVerifications`
- 🔴  `findAllUsedForVerification`
- `findAllUploadIds`: select by verificationID which identifies a group of documents submitted to specific provider. No need to specify provider.
- `findAllWithPhoto`: should be fine, the queried identityVerification is specified by OwnerId from an API call.
-  `findDocumentVerifications`: does not cause errors, but better to specify provider, so a server instance cleans only his documents.
- `terminate`: uses ids fetched from `findExpiredVerifications` or `findDocumentVerifications` from above.

`DocumentResultRepository`
- ⚠️ `streamAllInProgressDocumentSubmits`: Only relevant for async submit. Specify provider.
- ⚠️ `streamAllInProgressDocumentSubmitVerifications`: Only relevant for async verify. Specify provider.
- `findLatestResults`: Select by verificationID, which are gathered from `streamAllInProgressDocumentSubmitVerifications` or `findAllDocumentVerifications`.

**EDIT**: The problem turned out to be in the scheduled tasks. For async processing there are two task -- to check if document submit is completed and to check if document verification is completed. For that, two queries in `DocumentResultRepository` are used -- `streamAllInProgressDocumentSubmits` and `streamAllInProgressDocumentSubmitVerifications`. Async is actually not used.

The main problem is in scheduled task that triggers `NEXT_STEP` event every 3s. That one just gets all IdentityVerificationEntities and tries to move them to a next state. As the query to get the entities is not conditioned on the provider name, each deployed components tries to move with all those entities, creating mismatch of providers.

jnpsk:
Fix #956
A State Machine instance is created during `/identity/init` call.
The main problem was in the State Machine task, that periodically triggers `NEXT_STEP` event. The task works by taking all relevant IdentityVerificationEntities from the database. Based on these entities, a State Machine instance is retrieved **or created if it doesn't exist** yet. The associated IdentityVerificationEntity is then inserted as a context into the obtained State Machine.  
Relevant SQL queries are called by actions of State Machine and are parameterised by IdentityVerificationEntity that is retrieved from context of the current State Machine. _So there could be duplicated State Machines for each onboarding instance racing between each other._

The other queries from the DocumentVerificationRepository are mainly for periodic cleanup. These have not been modified because it can be expected those will not cause errors during testing and the modification does not add value to the production run.

Provider name has been added to queries to provide async runs.

> The modification of the State Machine task slightly changes its behaviour. Previously it gets all relevant IdentityVerificationEntity regardless of whether they have documents assigned. Now it considers only IdentityVerificationEntity that has at least one document submitted to the configured provider. But ultimately it doesn't affect the run, because to pass the defined Guard there has to be at least one document correctly submitted to the configured provider.


enrollment-server-onboarding-common/src/test/java/com/wultra/app/onboardingserver/common/database/DocumentResultRepositoryTest.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @author Jan Pesek, jan.pesek@wultra.com
+ */
+@DataJpaTest
+@ActiveProfiles("test")
+@Transactional
+class DocumentResultRepositoryTest {
+
+    @Autowired
+    DocumentResultRepository tested;
```
banterCZ, line 43:
```suggestion
    private DocumentResultRepository tested;
```

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentProcessingBatchService.java
```diff
@@ -45,17 +46,22 @@ public class DocumentProcessingBatchService {
...
      * Service constructor.
      * @param documentResultRepository Document verification result repository.
      * @param documentProcessingService Document processing service.
+     * @param identityVerificationConfig Configuration properties.
      */
     @Autowired
     public DocumentProcessingBatchService(
             DocumentResultRepository documentResultRepository,
-            DocumentProcessingService documentProcessingService) {
+            DocumentProcessingService documentProcessingService,
```
banterCZ, line 60: You may replace it with `@AllArgsConstructor` if you want to.

banterCZ (approved):
Looks good.

PR #969, opened 2023-12-21
Fix #968: MonetaryConverter throws an exception for currency '...'

### LINKED ISSUE #968: MonetaryConverter throws an exception for currency '...'
When currency `...` is passed to `MonetaryConverter` it throws an unhandled exception which leads to an internal server error.

Relates to #657

banterCZ:


romanstrobl (approved):
Looks OK.

kober32 (approved):
👍

PR #971, opened 2023-12-21
Fix #970: AmountAttribute with invalid amount

### LINKED ISSUE #970: AmountAttribute with invalid amount
The input validation should be improved. Currently, when not a number is passed into `AmountAttribute`, the operation is returned successfully in the list but without that attribute at all (a warning is logged `logger.warn("Invalid number format: {}, skipping the AMOUNT attribute!", amount);`). This is caused by wrong integration, not by user interaction. The integrator has to fix it but should get a clue.

There are two options to handle that:

- Do not return such an operation at all. Unfortunately, the push notification has already been sent.
- Do not fill in the formatted field but the raw one, at least.

A similar situation is for `AmountConversionAttribute` as well.

banterCZ:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -326,25 +335,37 @@ private static Optional<Attribute> buildAmountConversionAttribute(final Operatio
...
+        String targetAmountFormatted;
+        String sourceValueFormatted;
+        String targetValueFormatted;
+        final Locale locale = LocaleContextHolder.getLocale();
+        final String sourceCurrencyRaw = sourceCurrency.get();
+        final String targetCurrencyRaw = targetCurrency.get();
+
         try {
             sourceAmountRaw = new BigDecimal(sourceAmount.get());
             targetAmountRaw = new BigDecimal(targetAmount.get());
```
zcgandcomp, line 350: In this case, you need to do it per partes - as when the source amount is not a number but the target amount is Ok, you will put null to the target amount.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverterTest.java
```diff
@@ -515,6 +515,98 @@ void testConvertAttributes() throws Exception {
...
+                .sourceCurrencyFormatted("ETH")
+                .sourceValueFormatted("source not a number ETH")
+                .targetAmount(null)
+                .targetAmountFormatted("target not a number")
+                .targetCurrency("USD")
+                .targetCurrencyFormatted("$")
+                .targetValueFormatted("target not a number USD")
+                .build(), atributesIterator.next());
+    }
+
```
zcgandcomp, line 609: test for single amount attribute (target/source) 
```suggestion
    @Test
    void testConvertAmountConversionSource_notANumber() throws Exception {
        final OperationDetailResponse operationDetail = createOperationDetailResponse();
        operationDetail.setParameters(ImmutableMap.<String, String>builder()
                .put("sourceAmount", "source not a number")
                .put("sourceCurrency", "ETH")
                .put("targetAmount", "10")
                .put("targetCurrency", "USD")
                .put("dynamic", "true")
                .build());

        final OperationTemplateEntity operationTemplate = new OperationTemplateEntity();
        operationTemplate.setAttributes("""
                [
                  {
                    "id": "operation.amountConversion",
                    "type": "AMOUNT_CONVERSION",
                    "text": "Amount Conversion",
                    "params": {
                      "dynamic": "dynamic",
                      "sourceAmount": "sourceAmount",
                      "sourceCurrency": "sourceCurrency",
                      "targetAmount": "targetAmount",
                      "targetCurrency": "targetCurrency"
                    }
                  }
                ]""");

        LocaleContextHolder.setLocale(new Locale("en"));
        final Operation result = tested.convert(operationDetail, operationTemplate);

        final List<Attribute> attributes = result.getFormData().getAttributes();

        assertEquals(1, attributes.size());
        final var atributesIterator = attributes.iterator();
        assertEquals(AmountConversionAttribute.builder()
                .id("operation.amountConversion")
                .label("Amount Conversion")
                .dynamic(true)
                .sourceAmount(null)
                .sourceAmountFormatted("source not a number")
                .sourceCurrency("ETH")
                .sourceCurrencyFormatted("ETH")
                .sourceValueFormatted("source not a number ETH")
                .targetAmount(BigDecimal.TEN)
                .targetAmountFormatted("10.00")
                .targetCurrency("USD")
                .targetCurrencyFormatted("$")
                .targetValueFormatted("target not a number USD")
                .build(), atributesIterator.next());
    }

    @Test
    void testConvertAmountConversionTarget_notANumber() throws Exception {
        final OperationDetailResponse operationDetail = createOperationDetailResponse();
        operationDetail.setParameters(ImmutableMap.<String, String>builder()
                .put("sourceAmount", "10")
                .put("sourceCurrency", "ETH")
                .put("targetAmount", "target not a number")
                .put("targetCurrency", "USD")
                .put("dynamic", "true")
                .build());

        final OperationTemplateEntity operationTemplate = new OperationTemplateEntity();
        operationTemplate.setAttributes("""
                [
                  {
                    "id": "operation.amountConversion",
                    "type": "AMOUNT_CONVERSION",
                    "text": "Amount Conversion",
                    "params": {
                      "dynamic": "dynamic",
                      "sourceAmount": "sourceAmount",
                      "sourceCurrency": "sourceCurrency",
                      "targetAmount": "targetAmount",
                      "targetCurrency": "targetCurrency"
                    }
                  }
                ]""");

        LocaleContextHolder.setLocale(new Locale("en"));
        final Operation result = tested.convert(operationDetail, operationTemplate);

        final List<Attribute> attributes = result.getFormData().getAttributes();

        assertEquals(1, attributes.size());
        final var atributesIterator = attributes.iterator();
        assertEquals(AmountConversionAttribute.builder()
                .id("operation.amountConversion")
                .label("Amount Conversion")
                .dynamic(true)
                .sourceAmount(BigDecimal.TEN)
                .sourceAmountFormatted("10.00")
                .sourceCurrency("ETH")
                .sourceCurrencyFormatted("ETH")
                .sourceValueFormatted("source not a number ETH")
                .targetAmount(null)
                .targetAmountFormatted("target not a number")
                .targetCurrency("USD")
                .targetCurrencyFormatted("$")
                .targetValueFormatted("target not a number USD")
                .build(), atributesIterator.next());
    }

```

zcgandcomp (changes requested):

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
banterCZ: A nice catch, thanks.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -285,18 +285,27 @@ private static Optional<Attribute> buildAmountAttribute(final OperationTemplateP
...
+        String valueFormatted;
+        final String currencyRaw = currency.get();
+        final Locale locale = LocaleContextHolder.getLocale();
+
         try {
             amountRaw = new BigDecimal(amount.get());
-        } catch (NumberFormatException ex) {
-            logger.warn("Invalid number format: {}, skipping the AMOUNT attribute!", amount);
-            return Optional.empty();
+            amountFormatted = MonetaryConverter.formatAmount(amountRaw, currencyRaw, locale);
```
romanstrobl, line 297: Perhaps we can extract the logic which repeats in this pull request into a method? I realize this will require handling of inputs and inputs, however I am corned of future fixes which may update the logic only in one place. If this is not possible, leave it as is.

romanstrobl (approved):
Approved, however I would take a look if we can extract the repeated logic into a method.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
banterCZ: Thanks for the comment, yes, it is possible.

zcgandcomp (approved):
Looks OK.

PR #979, opened 2024-01-04
Fix #922: EvaluateClientRequest should contain extracted data

### LINKED ISSUE #922: EvaluateClientRequest should contain extracted data
The current implementation of `OnboardingProvider#evaluateClient(EvaluateClientRequest)` is suboptimal and contains only `verificationId`. It should be improved to contain the extracted data to be able for example to perform cross-document verification.

jnpsk:
Fix #922:

Evaluate Client Request now contains list of serialised extracted data. Each element of the list corresponds to a document/page uploaded and its format depends on the document provider used. The list of extracted data is filled only when `include-extracted-data` property is set to true. Default is false, so the default behaviour is not changed.

Verification result data are not included in the request. It does not seem beneficial as those were already checked during the onboarding process and the verification strictness is partially configurable.

**In case of Innovatrics** the extracted data also contains links to document photos or portrait photo, hence can be revisited using Innovatrics API. It does not contain reference to a selfie photo, which does not seem as a problem as previous document portrait to selfie matching was already done during the onboarding process.

**In case of ZenID** the extracted data also contains image hash which can be used to identify a portrait photo in ZenID API. The verificationId still available to revisit the onboarding process in Zen ID.

It does not seem beneficial to also include portrait or selfie directly as a payload. It would significantly increase the request size and the recipient would need a dedicated service to compare the faces anyway. Reference to provider resources seems sufficient.

Finally the document provider name in the request was never set before, although the request contained the property. As of now, we are preparing our service for another provider, and so it could be beneficial to actually include the name in the request.

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
@@ -21,7 +21,9 @@ import com.wultra.app.enrollmentserver.model.enumeration.DocumentStatus;
 import com.wultra.app.enrollmentserver.model.enumeration.ErrorOrigin;
 import com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationPhase;
 import com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus;
+import com.wultra.app.enrollmentserver.model.integration.DocumentSubmitResult;
 import com.wultra.app.enrollmentserver.model.integration.OwnerId;
+import com.wultra.app.onboardingserver.common.database.entity.DocumentResultEntity;
 import com.wultra.app.onboardingserver.common.database.entity.DocumentVerificationEntity;
 import com.wultra.app.onboardingserver.common.database.entity.IdentityVerificationEntity;
 import com.wultra.app.onboardingserver.common.service.AuditService;
@@ -33,6 +35,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.util.List;
 import java.util.Set;
 
 import static com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus.ACCEPTED;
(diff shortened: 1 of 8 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
```diff
@@ -117,10 +127,29 @@ public void processClientEvaluation(final IdentityVerificationEntity identityVer
         processTooManyEvaluationError(identityVerification, ownerId);
     }
 
-    private static String getVerificationId(final IdentityVerificationEntity identityVerification) {
-        final Set<String> verificationIds = identityVerification.getDocumentVerifications().stream()
+    private static Set<DocumentVerificationEntity> getAcceptedDocuments(final IdentityVerificationEntity identityVerification) {
```
banterCZ, line 130: Nitpick: `get*` may be confusing when we want to distinguish ordinary _getter_.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/model/request/EvaluateClientRequest.java
```diff
@@ -22,13 +22,16 @@
...
+import java.util.List;
+
 /**
  * Request object for {@link OnboardingProvider#evaluateClient(EvaluateClientRequest)}.
  *
  * @author Lubos Racansky, lubos.racansky@wultra.com
  */
 @Builder
 @Getter
+@Setter
```
banterCZ, line 34: Do we really need the setter? Until now, the class has been immutable. (Or wannabe at least).

banterCZ (commented):
In general, it looks good. I suggest merging it after the 1.6.0 release.

romanstrobl (approved):
I have nothing to add, however I agree with @banterCZ comments.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/model/request/EvaluateClientRequest.java
jnpsk: Ok, setter is not used now. I wanted to keep the condition with `config.isSendingExtractedDataEnabled()` in the main method so the logic when extracted data are included is clear.

banterCZ (approved):

PR #980, opened 2024-01-04
Fix #978: Remove Guava dependency

### LINKED ISSUE #978: Remove Guava dependency
Guava dependency is transitively introduced from `powerauth-crypto`. Since we are removing this dependency in https://github.com/wultra/powerauth-crypto/issues/554 we should also revisit the usage of Guava library in `enrollment-server`.

banterCZ:


pom.xml
```diff
@@ -336,26 +329,12 @@
                                         <exclude>org.apache.tomcat.embed:*:*:*:compile</exclude>
                                         <exclude>org.bouncycastle:bcpkix-jdk15on:*:*:compile</exclude>
                                         <exclude>org.bouncycastle:bcprov-jdk15on:*:*:compile</exclude>
+                                        <exclude>com.google.guava:guava*:*:*:compile</exclude>
```
banterCZ, line 332: - [ ] Depends on https://github.com/wultra/powerauth-crypto/pull/569

romanstrobl (approved):
Looks OK.

PR #992, opened 2024-01-12
Fix #913: Improve error handling in MobileTokenController

### LINKED ISSUE #913: Improve error handling in MobileTokenController
`MobileTokenController` error handling is too strict. E.g. `/operation/history` catches `PowerAuthClientException` and remaps it to `MobileTokenAuthException`. The result is `401 Unauthorized` without any detail. Actually, the detail is only in the enrollment-server log.
We should provide some hints. The mobile application must be able to handle it.

❗ For missing authorization, 401 **must** remain.

Consider other methods in this controller as well.

jnpsk:
Fix #913 
by resolving the error code in the exception.

### DIFF
--- a/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java
+++ b/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java
@@ -182,7 +182,7 @@ public class MobileTokenController {
                 case OPERATION_NOT_FOUND -> {
                     logger.info("Operation ID: {} not found: {}", request.getRequestObject().getId(), e.getMessage());
                     logger.debug("Operation ID: {} not found.", request.getRequestObject().getId(), e);
-                    throw new MobileTokenException(ErrorCode.OPERATION_FAILED, "No operation was found with the provided identifier.");
+                    throw new MobileTokenException(ErrorCode.INVALID_OPERATION, "No operation was found with the provided identifier.");
                 }
                 case INVALID_REQUEST -> {
                     logger.info("Request validation error: {}", e.getMessage());
@@ -234,7 +234,7 @@ public class MobileTokenController {
                 case OPERATION_NOT_FOUND -> {
                     logger.info("Operation ID: {} not found: {}", request.getRequestObject().getId(), e.getMessage());
                     logger.debug("Operation ID: {} not found.", request.getRequestObject().getId(), e);
-                    throw new MobileTokenException(ErrorCode.OPERATION_FAILED, "No operation was found with the provided identifier.");
+                    throw new MobileTokenException(ErrorCode.INVALID_OPERATION, "No operation was found with the provided identifier.");
                 }
                 case INVALID_REQUEST -> {
                     logger.info("Request validation error: {}", e.getMessage());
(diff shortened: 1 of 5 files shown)


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java
```diff
@@ -68,6 +71,13 @@
 @RequestMapping("api/auth/token/app")
 public class MobileTokenController {
 
+    public static final String APPLICATION_NOT_FOUND = "ERR0015";
```
banterCZ, line 74: May we stick with private constants?

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/MobileTokenController.java
```diff
@@ -115,8 +125,24 @@ public ObjectResponse<OperationListResponse> operationList(@Parameter(hidden = t
...
+                }
+                case INVALID_REQUEST -> {
+                    logger.info("Request validation error: {}", e.getMessage());
+                    logger.debug("Request validation error.", e);
+                    throw new MobileTokenException(ErrorCode.INVALID_REQUEST, "Request validation error: %s".formatted(e.getMessage()));
+                }
+                default -> {
+                    logger.warn("Calling PowerAuth service failed: {}", e.getMessage());
+                    logger.debug("Calling PowerAuth service failed.", e);
+                    throw new InternalServiceException("Unable to call upstream service.");
```
banterCZ, line 143: 🤔 Is it really `InternalServiceException` or rather `RemoteCommunicationException`? Moreover, we provide details for  the `ERR0015` and `ERR0024` only, other error codes cause `InternalServiceException`, correct? But we map `RemoteCommunicationException` to 409 status code.

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/enumeration/ErrorCode.java
```diff
@@ -42,6 +42,18 @@ public class ErrorCode {
      */
     public static final String INVALID_ACTIVATION           = "INVALID_ACTIVATION";
 
+    /**
+     * Error code for situation when an invalid application identifier is
+     * attempted for operation manipulation.
+     */
+    public static final String INVALID_APPLICATION          = "INVALID_APPLICATION";
```
banterCZ, line 49: I suppose that these error codes should be discussed with @kober32 and @Hopsaheysa

kober32: It would be nice to give us examples of when those 2 errors can happen so we can make a hint for the mobile devs what to show to the user.

romanstrobl (approved):
I have nothing to add, however I agree with @banterCZ comments, especially reusing the `RemoteCommunicationException` for consistency with other endpoints. It would be probably confusing to introduce a new exception class for the same purpose.

banterCZ (approved):
Ok from BE point of view. Let's wait for approval from mobile developers.

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/enumeration/ErrorCode.java
jnpsk: This controller handles endpoints on the path `/api/auth/token/app`

**INVALID_APPLICATION** (status 400) happens if the application specified by `applicationId` in a request does not exist. Relevant to:
```
/operation/list
/operation/history
/operation/authorize
/operation/cancel
```

**INVALID_OPERATION** (status 400) may happen in any of the three cases:  
1. Operation specified by `operationId` in a request does not exist. Relevant to:
```
/operation/detail
/operation/detail/claim
/operation/authorize
/operation/cancel
```
2. Operation is not claimed to the user specified by `userId` in a request. Relevant to:
```
/operation/authorize
/operation/cancel 
```
3. Operation specified by `operationId` in a request is not in the `PENDING` state. Relevant to:
```
/operation/authorize
/operation/cancel 
```

**INVALID_REQUEST** (status 400) may happen in any of the six endpoints in case a required field of a request is missing.

**REMOTE_COMMUNICATION_ERROR** (status 409) may happen in any of the six endpoints in case an unexpected error caused by remote server occurs.

kober32 (approved):
OK

PR #1003, opened 2024-01-30
Fix #995: Filter out errors from document verification in onboarding

### LINKED ISSUE #995: Filter out errors from document verification in onboarding
Filter out specific reasons for document rejection in onboarding process - level only generic messages in errors.

i.e. in the errors list return a single error with a generic meaning.

```
"status": "OK",
"responseObject":
{
"status": "IN_PROGRESS",
"documents":
[
{
"filename": "id_card_front.jpg",
"id": "19c84663-b3bb-48f6-b2f4-079ebc6f9910",
"type": "ID_CARD",
"side": "FRONT",
"status": "REJECTED",
"errors": [
"Error verifying the document."

```

jnpsk:
Fix #995 by modifying the process of mapping `DocumentVerificationEntity` to `DocumentMetadataResponseDto` (`/document/status` endpoint response). Now, if an entity contains some specific error or rejection reasons, they are mapped to single generic reason "Error verifying the document."

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
@@ -534,12 +534,12 @@ public class IdentityVerificationService {
         entities.forEach(entity -> {
             DocumentMetadataResponseDto docMetadata = toDocumentMetadata(entity);
 
-            if (DocumentStatus.REJECTED.equals(entity.getStatus())) {
-                List<String> errors = collectRejectionErrors(entity);
-                if (docMetadata.getErrors() == null) {
-                    docMetadata.setErrors(new ArrayList<>());
+            if (DocumentStatus.REJECTED == entity.getStatus()) {
+                final List<String> errors = collectRejectionErrors(entity);
+                // Hide specific reasons for rejection if any.
+                if (!errors.isEmpty()) {
+                    docMetadata.setErrors(List.of("Error verifying the document."));
                 }
-                docMetadata.getErrors().addAll(errors);
             }
 
             docsMetadata.add(docMetadata);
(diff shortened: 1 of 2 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -534,12 +534,12 @@ public List<DocumentMetadataResponseDto> createDocsMetadata(List<DocumentVerific
         entities.forEach(entity -> {
             DocumentMetadataResponseDto docMetadata = toDocumentMetadata(entity);
 
-            if (DocumentStatus.REJECTED.equals(entity.getStatus())) {
-                List<String> errors = collectRejectionErrors(entity);
-                if (docMetadata.getErrors() == null) {
-                    docMetadata.setErrors(new ArrayList<>());
+            if (DocumentStatus.REJECTED == entity.getStatus()) {
+                final List<String> errors = collectRejectionErrors(entity);
```
banterCZ, line 538: Just thinking. Does it make sense to return from `collectRejectionErrors`? Should not we refactor it into something like `auditRejectionErrors`?

banterCZ (approved):
Great, error detailss are not propagated anymore.

zcgandcomp (commented):
Looks fine, are we logging those errors? Or store them? We will need it for internal handling. But as a return to user it is fine.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
jnpsk: We can do that. The `collectRejectionErrors` is called only in this case and so it is sufficient to return boolean from the method, indicating whether an error exists for a document.

jnpsk:
@zcgandcomp 
> Looks fine, are we logging those errors? Or store them?

The error / rejection reasons are produced by `DocumentVerificationProvider`s (submit or verify function) and stored in `es_document_result` table.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
jnpsk: Actually we do not need to call the method `collectRejectionErrors` at all. It is quite complex function that visits database twice and does deserialisation, just for getting a messages that are not used anyway.

1. Result of a verification is logged and stored right after the verification process is finished by a document verification provider. So we won't lose any information about a document.
2. A document that is failed has error detail, a document that is rejected has rejection reason for sure.
3. The method `collectRejectionErrors` did not set any other attribute than the errors, which should not leak any details now.
4. The method is called within `/document/status` endpoint and we do not want to include specific error details anymore. Using this method just to log rejection details does not seem as expected behaviour to me.
5. The audit function only logs access to the document provider.

zcgandcomp (approved):
Looks OK.

PR #1029, opened 2024-03-04
Fix #1028: Duplicate maven-war-plugin declaration

### LINKED ISSUE #1028: Duplicate maven-war-plugin declaration
```console
Some problems were encountered while building the effective model for com.wultra.security:enrollment-server-onboarding:war:1.7.0-SNAPSHOT
'build.plugins.plugin.(groupId:artifactId)' must be unique but found duplicate declaration of plugin org.apache.maven.plugins:maven-war-plugin @ line 231, column 21
It is highly recommended to fix these problems because they threaten the stability of your build.
For this reason, future Maven versions might no longer support building such malformed projects.

```

banterCZ:


jnpsk (approved):
nice catch!

Issue #1033, opened 2024-03-07
Template detail endpoint


banterCZ:
Expose an endpoint providing template details `/api/admin/template`.
The endpoint should not be exported into the swagger and should be authenticated.

PR #1037, opened 2024-03-20
Make admin API to return template attributes in structured format

banterCZ:
A follow-up to #1033

### DIFF
--- a/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/admin/AdminController.java
+++ b/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/admin/AdminController.java
@@ -17,6 +17,9 @@
  */
 package com.wultra.app.enrollmentserver.controller.api.admin;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.wultra.app.enrollmentserver.api.model.enrollment.response.TemplateDetailResponse;
 import com.wultra.app.enrollmentserver.database.entity.OperationTemplateEntity;
 import com.wultra.app.enrollmentserver.impl.service.OperationTemplateService;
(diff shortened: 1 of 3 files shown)


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/admin/AdminController.java
```diff
@@ -45,25 +49,40 @@ public class AdminController {
 
     private final OperationTemplateService operationTemplateService;
 
+    private final ObjectMapper objectMapper;
+
     @GetMapping("/template")
     public ObjectResponse<List<TemplateDetailResponse>> templates() {
```
petrdvorak, line 55: Not an ideal format of response - you cannot add any attributes to the response object. I recommend creating one more wrapper object and keeping the list under the `templates` key in it.

petrdvorak (changes requested):
There is one recommendation in the response structure, but I did not get the returned format.

We need grouping by template name, then grouping by language, and then grouping by content type. I do not see it here.

banterCZ:
> There is one recommendation in the response structure, but I did not get the returned format.
> 
> We need grouping by template name, then grouping by language, and then grouping by content type. I do not see it here.

I prepared grouping in PAC, not in ES

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/controller/api/admin/AdminController.java
zcgandcomp: I guess you mean (as used in others APIs)
API should return `ObjectResponse<TemplateListResponse>`
and create an object
`TemplateListResponse extends ArrayList<TemplateDetailResponse> `

banterCZ: Ok, I see.

petrdvorak (approved):
Looks OK

romanstrobl (approved):
Nice refactoring.

PR #1080, opened 2024-06-19
Fix #1079: Refactor MobileTokenConverter#prepareFormData

### LINKED ISSUE #1079: Refactor MobileTokenConverter#prepareFormData
Refactor `forEach` with a side-effect into `collect` in https://github.com/wultra/enrollment-server/blob/ceca096d84c27f92b952ff52d402e315aca1aa98/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java#L139

Safe to fix because it is covered by `MobileTokenConverterTest`.

banterCZ:
- Change `forEach` with a side-effect into `collect`

### DIFF
--- a/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
+++ b/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
@@ -139,8 +139,7 @@ public class MobileTokenConverter {
         if (operationTemplateParams != null) {
             final List<Attribute> formDataAttributes = operationTemplateParams.stream()
                     .map(templateParam -> buildAttribute(templateParam, parameters))
-                    .filter(Optional::isPresent)
-                    .map(Optional::get)
+                    .flatMap(Optional::stream)
                     .toList();
             formData.setAttributes(formDataAttributes);
         }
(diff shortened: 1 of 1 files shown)


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -130,15 +131,18 @@ private FormData prepareFormData(
...
+        if (attributes == null) {
+            return formData;
+        }
+
+        final List<OperationTemplateParam> operationTemplateParams = objectMapper.readValue(attributes, new TypeReference<>(){});
+        if (operationTemplateParams != null) {
+            final List<Attribute> formDataAttributes = operationTemplateParams.stream()
+                    .map(templateParam -> buildAttribute(templateParam, parameters))
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
```
jnpsk, line 143: This could be one line shorter when using `.flatMap(Optional::stream)`.

jnpsk (approved):
Looks good

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
banterCZ: 🤔 Nice!

romanstrobl (approved):
Good idea.

PR #1085, opened 2024-06-19
Fix #1077: Add optional texts to the operation template for - success - failure - reject

### LINKED ISSUE #1077: Add optional texts to the operation template for - success - failure - reject
Add optional texts to the operation template for
- success
- failure
- reject

The Operation `ui` object could be extended to contain something like `texts` or more specifically `resultTexts` with all texts optional (for customers to be able to configure only the success text for example).

Maybe the text should be part of the operation (in the `formData` where `title` and `message` is) as this is not "what the operation looks like" but more of the "what the operation is and shows".

When the text is missing in the template, default from the app will be used.

banterCZ:


docs/Operation-Form-Data.md
```diff
@@ -204,3 +205,24 @@ Remarks:
...
+    "reject": "reject-message-value-may-be-changed"
+  }
+}
+```
+
+Remarks:
+
+- Always use `RESULT_TEXTS` as `type`
+- Parameters `success`, `failure`, and `reject` are optional. If not present (single property or the whole object), default messages are used.
+- The attribute `params` specifies values (not keys to retrieve values from the operation details parameters).
```
banterCZ, line 228: 👀 Are we ok with the params being used as values and not keys?

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/attributes/Attribute.java
```diff
@@ -86,7 +87,12 @@ public enum Type {
         /**
          * Image attribute type - represents an image as a URL to the thumbnail and the higher resolution.
          */
-        IMAGE
+        IMAGE,
+
+        /**
+         * Customization texts to show for {@code success}, {@code failure}, or {@code reject} operation.
+         */
+        RESULT_TEXTS
```
banterCZ, line 95: ℹ️ It should not be needed to model the discriminator, but let's refactor that in a separate issue.

docs/Operation-Form-Data.md
```diff
@@ -21,6 +21,7 @@ Following structured custom form data attributes are available:
 - `HEADING` - heading with label 
 - `ALERT` - alert with various alert types
 - `PARTY_INFO` - information about party
+- `RESULT_TEXTS` - Customization texts to show for `success`, `failure`, or `reject` operation. If not present (single property or the whole object), default messages are used.
```
romanstrobl, line 24: Fixed grammar:

`Customized texts to display for success, failure, or reject operations. If not provided (either as individual properties or for the entire object), default messages will be used.`

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
```diff
@@ -216,6 +216,8 @@ private static Optional<Attribute> buildAttribute(final OperationTemplateParam t
                 buildImageAttribute(templateParam, params);
             case "PARTY_INFO" ->
                 buildPartyInfoAttribute(templateParam, params);
+            case "RESULT_TEXTS" ->
+                buildResultTextsAttribute(templateParam);
```
romanstrobl, line 220: I think we should allow same customization as for other types, so the `params` should be used as a parameter here.

romanstrobl (changes requested):
I think we should follow the customization of operations in the same way as we customize all other form data attributes, e.g. use the `params` argument.

The other form data attributes also control what is displayed, so why should result texts behave differently?

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/converter/MobileTokenConverter.java
banterCZ: As @petrdvorak wrote, the format should be for example
```json
{
  "formData": {
    "title": "Confirm Payment",
    "message": "Hello,\nplease confirm following payment:",
    "resultTexts": {
      "success": "Payment of ${amount} ${currency} was confirmed",
      "reject": "Payment was rejected",
      "failure": "Payment approval failed"
    },
    "attributes": [
      {
        "type": "HEADING",
        "id": "operation.heading",
        "label": "Utility Payment"
      }
    ]
  }
}
```

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/controller/api/admin/AdminControllerTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    void testGetTemplates() throws Exception {
+        mvc.perform(get("/api/admin/template")
+                        .accept(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
+                .andExpect(content()
+                        .contentTypeCompatibleWith(MediaType.APPLICATION_JSON))
+                .andExpect(jsonPath("$.status", is("OK")));
+        // TODO Lubos assert resultTexts
```
banterCZ, line 52: - [ ] Add `resultTexts` even to admin API.

kober32 (commented):
Looks OK with me, but I let BackEnders to approve...

docs/sql/postgresql/enrollment/migration_1.7.0_1.8.0.sql
```diff
@@ -0,0 +1,3 @@
+-- Changeset enrollment-server/1.8.x/20240620-add-resultTexts.xml::1::Lubos Racansky
+-- Add result_texts column
+ALTER TABLE es_operation_template ADD result_texts TEXT;
```
romanstrobl, line 3: Do we really need a `TEXT` data type? The texts are messages displayed on a mobile phone, so a long varchar should be sufficient.

romanstrobl (approved):
Looks good, just one concern related to the `TEXT` data type for DB schema.

docs/sql/postgresql/enrollment/migration_1.7.0_1.8.0.sql
banterCZ: 🤔 A good point. Discussed with @petrdvorak. Since it is structured JSON, we prefer `TEXT`.

PR #1087, opened 2024-06-19
Fix #1086: Consider omitting Attribute discriminator

### LINKED ISSUE #1086: Consider omitting Attribute discriminator
Modeling of `Attribute` JSON discriminator should not be needed.
https://github.com/wultra/enrollment-server/blob/a57b00bf0d390330d447cc4658099450d8714a60/mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/attributes/Attribute.java#L97

banterCZ:


mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/attributes/AmountConversionAttribute.java
```diff
@@ -60,12 +60,11 @@ public class AmountConversionAttribute extends Attribute {
     /**
      * No-arg constructor.
      */
-    AmountConversionAttribute() {
-        super(Type.AMOUNT_CONVERSION);
+    private AmountConversionAttribute() {
```
banterCZ, line 63: The private no-arg constructor is enough, which is proved by the test.

romanstrobl (approved):
Let's test this change properly.

PR #1091, opened 2024-06-21
Fix #1090: Set proxyBeanMethods to false in Springs Configuration

### LINKED ISSUE #1090: Set proxyBeanMethods to false in Springs Configuration

banterCZ:


romanstrobl (approved):
Approved, even though benefits are marginal.

banterCZ:
Let's close it since it is not agreed upon by the team as the benefit is marginal and the risk is high.

PR #1129, opened 2024-10-24
Fix #1128: Confusing naming of method claimOperationInternal

### LINKED ISSUE #1128: Confusing naming of method claimOperationInternal
The method `claimOperationInternal` is used for two purposes:
- get operation detail
- claim operation

This is confusing because there are two scenarios for which the method is used. Let's clean up the code.

romanstrobl:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/MobileTokenService.java
```diff
@@ -312,7 +312,7 @@ public Operation claimOperation(String operationId, String language, String user
      * @throws PowerAuthClientException In case communication with PowerAuth Server fails.
      * @throws MobileTokenException When the operation is in incorrect state.
      */
-    private OperationDetailResponse getOperationDetailInternal(String operationId) throws PowerAuthClientException, MobileTokenException {
+    private OperationDetailResponse getOperationDetailInternal(@NotNull String operationId) throws PowerAuthClientException, MobileTokenException {
```
banterCZ, line 315: 🤔 If we are already refactoring the method names, how about renaming this to something like `fetchOpertionDetailFromPowerAuth`?

banterCZ (approved):
Thanks for removing the confusion!

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/MobileTokenService.java
romanstrobl: Yes, I will rename the `getOperationDetail` methods.

banterCZ (approved):

PR #1153, opened 2024-12-03
Fix #1150: RestOnboardingProviderAutoConfiguration could not be regis…

### LINKED ISSUE #1150: RestOnboardingProviderAutoConfiguration could not be registered
When the property `enrollment-server-onboarding.onboarding-adapter.url` is set, the enrollment server does not start with this error.

```console
The bean 'onboardingProvider', defined in class path resource [com/wultra/app/onboardingserver/provider/rest/RestOnboardingProviderAutoConfiguration.class], could not be registered. A bean with that name has already been defined in class path resource [com/wultra/app/onboardingserver/provider/OnboardingProviderAutoConfiguration.class] and overriding is disabled.

Action:

Consider renaming one of the beans or enabling overriding by setting spring.main.allow-bean-definition-overriding=true
```

Reported in 1.7.0

banterCZ:
…tered


romanstrobl (approved):
Nice fix.

PR #1167, opened 2025-01-28
Fix #1166: Duplicate package in component scan

### LINKED ISSUE #1166: Duplicate package in component scan
Duplicate packages after refactoring:

```
@ComponentScan(basePackages = {
        "com.wultra.security.powerauth",
        "com.wultra.security.powerauth",
})
```

_No response_

### Environment

_No response_

Incorrect package names after refactoring:

### Log output

_No response_

romanstrobl:


zcgandcomp (approved):
Looks OK.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/configuration/PowerAuthWebServiceConfiguration.java
```diff
@@ -41,8 +41,7 @@
 @Configuration
 @ConfigurationProperties("ext")
 @ComponentScan(basePackages = {
-        "com.wultra.security.powerauth",
-        "io.getlime.security.powerauth",
+        "com.wultra.security.powerauth"
```
banterCZ: Nitpick: It may be simplified even into `@ComponentScan("com.wultra.security.powerauth")`

banterCZ (approved):
As expected.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/configuration/PowerAuthWebServiceConfiguration.java
banterCZ: It is not crucial (just „figure skating“ 😉), but the point was that `value` is an alias for `basePackages`, so it is not needed at all.
```java
@AliasFor("value")
String[] basePackages() default {};
```

banterCZ (approved):

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/configuration/PowerAuthWebServiceConfiguration.java
romanstrobl: Ah, ok, sure.

banterCZ (approved):
Nice, thanks!

PR #1271, opened 2025-08-05
Fix #1270: status in Operation should be @NotNull

### LINKED ISSUE #1270: `status` in `Operation` should be `@NotNull`
The value is always filled anyway, and we expect it on the frontend

https://github.com/wultra/enrollment-server/blob/2dd7e4f300f4e402ae4bdf4ffa1fe8f956d7ef5a/mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/Operation.java#L41

banterCZ:


mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/Operation.java
```diff
@@ -38,6 +38,7 @@ public class Operation {
     private String name;
     @NotNull
     private String data;
+    @NotNull
```
banterCZ, line 41: 🤔 Should it be `@NotNull` only for `/operation/history` or also for `/operation/list`, `/operation/detail`, and `/operation/detail/claim`?

kober32 (commented):
OK with me

romanstrobl (changes requested):
The code still contains:

```
Optional details why the status has changed by backend services.
```

banterCZ:
> The code still contains:
> 
> ```
> Optional details why the status has changed by backend services.
> ```

But it relates to the `statusReason`, not the `status` itself, doesn't it?

romanstrobl:
> But it relates to the `statusReason`, not the `status` itself, doesn't it?

Ah yes, sorry.

romanstrobl (approved):

PR #1273, opened 2025-08-11
Fix #1272: Allow disabling Bouncy Castle initialization

### LINKED ISSUE #1272: Allow disabling Bouncy Castle initialization
### Task specification

Due to Bouncy Castle classloading issues on Tomcat we should allow just one `Security.addProvider()` call in case the same Tomcat instance is shared.

Thus, the `Security.addProvider()` should be optional.

romanstrobl:
This PR makes BC initialization configurable for setup with multiple applications on same application server.


hvge (approved):
Looks OK

PR #1291, opened 2025-09-23
Fix #1290: Adapt changes of PowerAuth Server model

### LINKED ISSUE #1290: Adapt changes of PowerAuth Server model
### Task specification

PowerAuth Server has been refactored into v3 and v4 models. Make a quick fix using v3. It may be migrated to v4 later on.

banterCZ:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/MobileTokenService.java
```diff
@@ -117,7 +121,7 @@ public OperationListResponse operationListForUser(
         request.setActivationId(activationId);
         final MultiValueMap<String, String> queryParams = httpCustomizationService.getQueryParams();
         final MultiValueMap<String, String> httpHeaders = httpCustomizationService.getHttpHeaders();
-        final com.wultra.security.powerauth.client.model.response.OperationListResponse operations =
+        final com.wultra.security.powerauth.client.model.response.v3.OperationListResponse operations =
```
banterCZ, line 124: This is a quick fix to make it compile. We may migrate to v4 later on.

banterCZ:
Let's prefer #1292 with v4.

PR #1292, opened 2025-09-23
Fix #1290: Adapt changes of PowerAuth Server model

### LINKED ISSUE #1290: Adapt changes of PowerAuth Server model

romanstrobl:


banterCZ (approved):
Great! Thanks for v4 implementation.

PR #1296, opened 2025-09-26
Fix #1295: Default DelegatingActivationCodeHandler

### LINKED ISSUE #1295: Default DelegatingActivationCodeHandler
### Task specification

Provide a default `DelegatingActivationCodeHandler` implementation. Custom beans in downstreams still have to be functional.

Retrieve the application configuration from PAS and use it for the logic.

banterCZ:


enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private ActivationCodeConfiguration convert(final Object value) {
+        try {
+            return objectMapper.convertValue(value, new TypeReference<>() {});
+        } catch (IllegalArgumentException e) {
+            logger.warn("Unable to convert {}", value, e);
+            return null;
+        }
+    }
```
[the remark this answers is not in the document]
banterCZ: No, explicit type argument `ActivationCodeConfiguration` can be replaced with `<>` and it is more legible.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+@AllArgsConstructor
+@Slf4j
+class DefaultDelegatingActivationCodeHandler implements DelegatingActivationCodeHandler {
+
+    private static final String ACTIVATION_CODE = "activation_code";
+
+    private final PowerAuthClient powerAuthClient;
+
+    private final ObjectMapper objectMapper = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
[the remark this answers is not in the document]
banterCZ: The handler is a singleton.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    private record ActivationCodeConfiguration(List<String> allowedTargetApplicationIds, ActivationType type) {
+        boolean isTypeOfCopy() {
+            return type == ActivationType.COPY;
+        }
+    }
+
+    private enum ActivationType {
```
romanstrobl: There is a clash with existing `ActivationType` in restful-integration.

I suggest `ActivationTransferType`.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandlerTest.java
```diff
@@ -0,0 +1,96 @@
...
+        assertNull(result);
+    }
+
+    private GetApplicationConfigResponse createResponse() throws JsonProcessingException {
+        final String json = """
+                {
+                  "applicationId": "source-1",
+                  "applicationConfigs": [
+                    {
+                      "key": "activation_code",
```
romanstrobl: This terminology is already used for activations using activation code.

I think we should call this configuration `activation_transfer`.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Specialization of {@link DelegatingActivationCodeHandler} using application configuration from PowerAuth Server.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@AllArgsConstructor
+@Slf4j
+class PowerAuthActivationCodeHandler implements DelegatingActivationCodeHandler {
+
+    private static final String ACTIVATION_CODE = "activation_code";
```
romanstrobl: I find this confusing, because this terminology conflicts with activation using activation code.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/DelegatingActivationCodeHandler.java
```diff
@@ -58,7 +60,9 @@ public interface DelegatingActivationCodeHandler {
      * @return List of new activation flags for the destination activation.
      * @throws ActivationCodeException Thrown in case activation flag processing fails.
      */
-    List<String> addActivationFlags(String sourceActivationId, List<String> sourceActivationFlags, String userId, String applicationId, String sourceAppId, List<String> sourceApplicationRoles, String destinationAppId, String destinationActivationId, String activationCode, String activationCodeSignature) throws ActivationCodeException;
+    default List<String> addActivationFlags(String sourceActivationId, List<String> sourceActivationFlags, String userId, String applicationId, String sourceAppId, List<String> sourceApplicationRoles, String destinationAppId, String destinationActivationId, String activationCode, String activationCodeSignature) throws ActivationCodeException {
+        return List.of();
```
romanstrobl, line 64: We are losing this functionality in the default implementation. Shouldn't it be configurable on PA server as well?

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/DelegatingActivationCodeHandler.java
```diff
@@ -73,6 +77,8 @@ public interface DelegatingActivationCodeHandler {
      * @param activationCodeSignature Activation code signature of the new activation code.
      * @throws ActivationCodeException Thrown in case activation code processing fails.
      */
-    void didReturnActivationCode(String sourceActivationId, String userId, String applicationId, String sourceAppId, String destinationAppId, String destinationActivationId, String activationCode, String activationCodeSignature) throws ActivationCodeException;
+    default void didReturnActivationCode(String sourceActivationId, String userId, String applicationId, String sourceAppId, String destinationAppId, String destinationActivationId, String activationCode, String activationCodeSignature) throws ActivationCodeException {
+        // Default implementation does nothing
```
romanstrobl, line 81: Isn't this a problem? What if a callback is required?

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
```diff
@@ -0,0 +1,103 @@
...
+
+        /**
+         * Keeps the original activation.
+         */
+        COPY,
+
+        /**
+         * The original activation is to be removed.
+         */
+        MOVE
```
romanstrobl, line 101: We were discussing that the original activation should be removed after the new activation is active and confirmed. The problem is that when the client does not finish the activation, neither of the activations is in `ACTIVE` state.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/ActivationCodeService.java
```diff
@@ -136,8 +100,9 @@ public ActivationCodeResponse requestActivationCode(ActivationCodeRequest reques
             final InitActivationRequest initRequest = new InitActivationRequest();
             initRequest.setUserId(sourceUserId);
             initRequest.setApplicationId(destinationAppId);
-            initRequest.setActivationOtpValidation(ActivationOtpValidation.ON_KEY_EXCHANGE);
+            initRequest.setCommitPhase(CommitPhase.ON_KEY_EXCHANGE);
             initRequest.setActivationOtp(otp);
+            initRequest.setAdditionalData(Map.of("sourceAppId", sourceAppId, "targetAppId", destinationAppId));
```
romanstrobl: We should probably include some additional data which explains that this activation was created by activation transfer.

romanstrobl (changes requested):
I think the PR is not ready to be merged yet, see my comments. We can resolve some of these issues separately, but let's discuss them first.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/ActivationCodeService.java
zcgandcomp: We do have the additional data, right?

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/DelegatingActivationCodeHandler.java
zcgandcomp: We can/should move the flag from the source activation in the case of move.
But yes, the new set of flags could be parameterized per source/target app pair.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/DelegatingActivationCodeHandler.java
zcgandcomp: Activation callback is going to be fired by PowerAuth Server, as configured. No additional callbacks are required.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
banterCZ: Sure, I will update the Javadoc. This is just an extension point to the next phase.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandlerTest.java
zcgandcomp: Agree, even though this is an activation by an activation code, we are configuring the transfer setting.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
banterCZ: Thanks for noticing. I will rename that.

enrollment-server/src/test/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandlerTest.java
banterCZ: A good idea!

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/DelegatingActivationCodeHandler.java
banterCZ: This is called from `ActivationCodeService` https://github.com/wultra/enrollment-server/blob/develop/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/ActivationCodeService.java#L157
```java
final List<String> flags = delegatingActivationCodeHandler.addActivationFlags(
        sourceActivationId, sourceActivationFlags, applicationId, sourceUserId, sourceAppId, sourceApplicationRoles, destinationAppId,
        iar.getActivationId(), iar.getActivationCode(), iar.getActivationSignature()
);
if (flags != null && !flags.isEmpty()) {
    logger.info("Calling PowerAuth Server to add activation flags to activation ID: {}, flags: {}.", iar.getActivationId(), flags.toArray());
    final AddActivationFlagsRequest addRequest = new AddActivationFlagsRequest();
    addRequest.setActivationId(iar.getActivationId());
    addRequest.getActivationFlags().addAll(flags);
    powerAuthClient.addActivationFlags(addRequest,
            httpCustomizationService.getQueryParams(),
            httpCustomizationService.getHttpHeaders()
    );
    logger.info("Successfully added flags to activation ID: {}.", iar.getActivationId());
} else {
    logger.info("Activation with ID: {} has no additional flags.", iar.getActivationId());
}
```

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/DelegatingActivationCodeHandler.java
banterCZ: It is called by `ActivationCodeService` https://github.com/wultra/enrollment-server/blob/develop/enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/ActivationCodeService.java#L150
This is kind of an adapter, not to force the implementation of the method. Anyway, RB downstream is also empty.
```java
delegatingActivationCodeHandler.didReturnActivationCode(
        sourceActivationId, sourceUserId, applicationId, sourceAppId, destinationAppId,
        iar.getActivationId(), iar.getActivationCode(), iar.getActivationSignature()
);
```

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/ActivationCodeService.java
banterCZ: Yes, additional data are present since https://github.com/wultra/powerauth-server/issues/1962

banterCZ: A good idea to add an explanation to the additional data.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
banterCZ: True, renamed.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
banterCZ: See that the current implementation works only with `COPY` as `.filter(ActivationCodeConfiguration::isTypeOfCopy)` https://github.com/wultra/enrollment-server/pull/1296/files#diff-6f1a9e7a2fa6f901de9498c60615c1ecee9c6df525f865b2859d3efe69326a11R71

banterCZ:
Documentation of the configuration suggested here https://github.com/wultra/powerauth-server/pull/2093

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/PowerAuthActivationCodeHandler.java
```diff
@@ -0,0 +1,103 @@
...
+            return type == ActivationTransferType.COPY;
+        }
+    }
+
+    private enum ActivationTransferType {
+
+        /**
+         * Keeps the original activation.
+         */
+        COPY,
```
banterCZ: `COPY` is misleading; let's rename it to `SPAWN`.

enrollment-server/src/main/java/com/wultra/app/enrollmentserver/impl/service/DelegatingActivationCodeHandler.java
banterCZ: Ok, I finally got it. We may introduce initial flags to the application configuration, but we agreed with @zcgandcomp that we may postpone it to version 2.0 with `MOVE`.

romanstrobl (approved):
I am fine with this change now.

PR #1304, opened 2025-10-02
Fix #1303: Disable Bouncy Castle dependency initialization by default

### LINKED ISSUE #1303: Disable Bouncy Castle dependency initialization by default
### Task specification

The Bouncy Castle dependency collides with PowerAuth server Bouncy Castle. Let's disable the initialization by default.

romanstrobl:
I disabled Bouncy Castle library initialization by default because it's inititalization clashes with PowerAuth server Bouncy Castle library when deployed on same Tomcat.


banterCZ (approved):
Ok, let's test it.

PR #1310, opened 2025-10-02
Add configuration property for Bouncy Castle provider initialization and enable it by default

romanstrobl:


banterCZ (approved):
Ok, for Tomcat, it could be disabled by default.

PR #1320, opened 2025-10-21
Fix #1319: Enable Bouncy Castle initialization by default

### LINKED ISSUE #1319: Enable Bouncy Castle initialization by default
### Task specification

Set property `bouncycastle.provider.initialization.enabled=true`.

A follow-up to #1272

banterCZ:


romanstrobl (approved):
I agree.

Issue #1289, opened 2025-09-22
New structure of preApprovalScreens


banterCZ:
### Task specification

Introduce a new `preApprovalScreens` (plural) in UI templates. Keep `preApprovalScreen` (singular) for compatibility reasons, but mark it as deprecated.

```json
{
  "flipButtons": true,
  "blockApprovalOnCall": false,
  "preApprovalScreens": [
    {
      "id": "custom_id", # This id will be used in 'mobileTokenData.preApprovalScreens' when calling /operation/cancel or /operation/authorize
      "type": "WARNING",
      "backButton": true,
      "image": "image-label", # Image is stored in-app
      "heading": "Watch out!",
      "message": "You may become a victim of an attack.", # Please add support for text formating
      "elements": [
        {
          "id": "custom_element_id", # It's optional for future use (styling)
          "type": "ALERT", # ENUM: LISTITEM, ALERT, BUTTON
          "style": "INFO", # ENUM: INFO, WARNING, DANGER
          "text": "Make sure the activation takes place on your device" # Please add support for text formating
        },
        {
          "id": "custom_element_id",
          "type": "BUTTON", # ENUM: LISTITEM, ALERT, BUTTON
          "action": "PHONE", # ENUM: LINK, MAIL, PHONE
          "text": "Call center",
          "href": "+42012345678"
        },
        {
          "id": "custom_element_id",
          "type": "LISTITEM" # ENUM: LISTITEM, ALERT, BUTTON
          "icon": "icon-label", # Icon is stored in-app
          "text": "You activate a new app and allow access to your accounts"  # Please add support for text formating
        }
      ],
      "controls": {
        "flip": true,
        "axis": "VERTICAL", # ENUM: HORIZONTAL, VERTICAL
        "decline": {
          "type": "REJECT", # ENUM: BACK, REJECT
          "text": "Reject Payment"
        },
        "approve": {
          "type": "BUTTON", # ENUM: SLIDER, BUTTON
          "text": "Approve Payment",
          "counter": 10
        }
      }
    },
    {
      "id": "custom_id",
      "type": "QR_SCAN",
      "heading": "Watch out!",
      "message": "You may become a victim of an attack."
    }
  ]
}
```

PR #1336, opened 2025-11-04
Fix #1335: Use polymorphism for elements of preApprovalScreens

### LINKED ISSUE #1335: Use polymorphism for elements of preApprovalScreens
### Task specification

Refactor elements of preApprovalScreens to use polymorphism with `one of`.

A follow-up to #1289

banterCZ:
Refactor elements of preApprovalScreens to use polymorphism with `one of`.

A follow-up to #1289

### DIFF
--- a/mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/PreApprovalScreenV2.java
+++ b/mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/PreApprovalScreenV2.java
@@ -17,9 +17,14 @@
  */
 package com.wultra.security.powerauth.lib.mtoken.model.entity;
 
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import lombok.Data;
+import lombok.Getter;
+import lombok.Setter;
 
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/PreApprovalScreenV2.java
```diff
@@ -65,28 +70,56 @@ public class PreApprovalScreenV2 {
...
+            @JsonSubTypes.Type(value = ButtonElement.class, name = Element.BUTTON)
+    })
+    public abstract static class Element {
+
+        public static final String LIST_ITEM = "LIST_ITEM";
+        public static final String ALERT = "ALERT";
+        public static final String BUTTON = "BUTTON";
+
+        @NotBlank
+        private String id;
```
Hopsaheysa, line 86: I thing that both id and text should be nullable

in this case ID is "helper" for additional styling - for example, LIST_ITEM with id="cursive" will change the style of the text to cursive

similar for the text - if it is empty or null we dont show it - for example ALERT without text will still show the cell with the warning icon

Hopsaheysa (changes requested):
I have a remark which is mainly for @tech-kratas so please look at it.
I think that id and text should be nullable?

mtoken-model/src/main/java/com/wultra/security/powerauth/lib/mtoken/model/entity/PreApprovalScreenV2.java
banterCZ: ok, reveted to make it nullable

Hopsaheysa (approved):
Don't know about the AI suggestions and how you normally do it with docs but from my perspective I like the nullability of the id and text so 👍

PR #1337, opened 2025-11-06
Fix #1325: Integrate BlinkID documents verification for Microblink provider

### LINKED ISSUE #1325: Integrate BlinkID documents verification
### Task specification

This endpoint should be integrated: https://blinkidverify.docs.microblink.com/docs/docver/api-docs/api-v2#tag/document-verification-v2/post/api/v2/docver

michal-rozehnal-w:
TODOs:
- test with cmd tool

Integration tests will be done in separate task https://github.com/wultra/enrollment-server/issues/1342


enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkCacheConfig.java
```diff
@@ -0,0 +1,48 @@
...
+    @Bean("microblinkDocumentsCache")
+    public Cache<String, MicroblinkVerificationData> microblinkDocumentsCache(final MicroblinkConfigProperties properties) {
+        return Caffeine.newBuilder()
+                .expireAfterWrite(properties.getExpireAfter())
+                .build();
+    }
+
+    @Bean("microblinkPhotoCache")
+    public Cache<String, String> microblinkPhotoCache(final MicroblinkConfigProperties properties) {
+        return Caffeine.newBuilder()
```
banterCZ: Have you considered using `CacheManager` instead?

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkCacheConfig.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Signer Cloud
```
banterCZ: different component

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkRestClientConfig.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Base64;
+
+/**
+ * Configuration for Microblink {@link RestClient}.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@ConditionalOnProperty(value = "enrollment-server-onboarding.document-verification.provider", havingValue = "microblink")
+@Configuration
+@ComponentScan("com.wultra.app.onboardingserver.provider.microblink")
```
banterCZ: Why are we turning the scan on? Isn't it working out of the box?

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkRestClientConfig.java
```diff
@@ -0,0 +1,63 @@
...
+
+        final var restClientConfig = properties.getRestClientConfig();
+        restClientConfig.setDefaultHttpHeaders(httpHeaders);
+
+        return new DefaultRestClient(restClientConfig);
+    }
+
+    private String buildBasicAuthHeader(MicroblinkConfigProperties properties) {
+        final var auth = properties.getUsername() + ":" + properties.getPassword();
+        final var authBytes = Base64.getEncoder().encode(auth.getBytes());
```
banterCZ: Have you noticed `RestClientConfiguration#httpBasicAuthUsername`?

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -17,66 +17,401 @@
...
+                .documents(documents)
+                .facePhotoId(UUID.randomUUID().toString())
+                .build();
+    }
+
+    private DocumentVerificationResponse sendApiRequest(final OwnerId ownerId, final MicroblinkVerificationData.Document frontDocument, final MicroblinkVerificationData.Document backDocument) throws DocumentVerificationException, RemoteCommunicationException {
+        try {
+            final var request = buildRequest(frontDocument, backDocument);
+
+            final var response = restClient.post("/api/v2/docver", request, new ParameterizedTypeReference<DocumentVerificationResponse>() {});
```
banterCZ: I do not insist, but for example, ZenId is split into a provider and a REST client, see `ZenidRestApiService`. Not everything there is perfect, just thinking about granularity and boundaries.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkCacheConfig.java
michal-rozehnal-w: Good idea, it is used now

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkCacheConfig.java
michal-rozehnal-w: Thank you for noticing this. I had only global setting setting for the Signer Cloud. I added local override for this repository.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkRestClientConfig.java
michal-rozehnal-w: I saw it in other providers, so I used it too. However you are right it works automatically and it is not needed. Removed

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkRestClientConfig.java
michal-rozehnal-w: I missed these, thank you. They are used now.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
michal-rozehnal-w: Yes, I was thinking about creating a separate service class, but I don’t see any significant benefit. There’s only a single REST API call, and the unit tests aren’t complex even with everything in one class. But I don't have strong opinion about it.

docs/onboarding/Configuration-Properties.md
```diff
@@ -155,6 +155,17 @@ The Onboarding Server uses the following public configuration properties:
...
+|---------------------------------------------------------------------------------------------------------|---------|----------------------------------------------------------------|
+| `enrollment-server-onboarding.document-verification.microblink.restClientConfig.baseUrl`                |         | Base REST service URL for Microblink.                          |
+| `enrollment-server-onboarding.document-verification.microblink.restClientConfig.httpBasicAuthEnabled`   | `true`  | Whether HTTP Basic authentication is enabled.                  |
+| `enrollment-server-onboarding.document-verification.microblink.restClientConfig.httpBasicAuthUsername`  |         | HTTP Basic authentication username.                            |
+| `enrollment-server-onboarding.document-verification.microblink.restClientConfig.httpBasicAuthPassword`  |         | HTTP Basic authentication password.                            |
+| `enrollment-server-onboarding.document-verification.microblink.cacheRecordTTL`                          | `1h`    | Time to live for cached Microblink documents for verification. |
```
michal-rozehnal-w, line 166: @tech-kratas what would be the most appropriate default value here?

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/api/DocumentVerificationResponse.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.List;
+
+/**
+ * Response from Microblink BlinkID REST API.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@Builder
+public record DocumentVerificationResponse(
```
michal-rozehnal-w: I created this custom response instead of using the generated one because I was struggling with the `extractions` field.
The `Result` type is essentially an abstract class, and there are subclasses such as `StringResult`, `DateResult`, etc.

I tried using `<useInheritance>true</useInheritance>` in the `pom.xml` and also modified the downloaded `blinkid-verify-api-v2.yaml` by adding:

```
discriminator:
  propertyName: type
```

However, this wasn’t ideal since the original definition shouldn’t be modified.

In the end, there was a conflict with the `originalResult` property I think - for some classes it’s a `String`, while for others it’s a `List<String>`, which caused issues for the generator.

Ultimately, it was easier to create this custom data structure than to try to fix the generator.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkCacheConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Cache config for storing Microblink documents.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@Configuration
+@EnableCaching
+@Slf4j
+public class MicroblinkCacheConfig {
```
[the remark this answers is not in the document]
michal-rozehnal-w: Right, I noticed that when other tests failed. Fixed now

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -17,66 +17,427 @@
...
+        var microblinkVerificationData = verificationDataCache.get(activationId, MicroblinkVerificationData.class);
+
+        final var addedDocuments = submittedDocuments.stream()
+                .map(MicroblinkDocumentVerificationProvider::buildMicroblinkVerificationDocument)
+                .toList();
+
+        if (microblinkVerificationData == null) {
+            microblinkVerificationData = MicroblinkVerificationData.builder()
+                    .documents(addedDocuments)
+                    .facePhotoId(UUID.randomUUID().toString())
```
michal-rozehnal-w, line 96: Documents are not uploaded, so document `uploadId` and `facePhotoId` are random ids. We just have to be able to find them in cache when they are by these ids requested.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -17,66 +17,427 @@
...
     @Override
     public boolean shouldStoreSelfie() {
-        throw new NotImplementedException();
+        return false;
     }
 
     @Override
-    public DocumentsVerificationResult verifyDocuments(OwnerId id, List<String> uploadIds) throws RemoteCommunicationException, DocumentVerificationException {
-        throw new NotImplementedException();
+    public DocumentsVerificationResult verifyDocuments(OwnerId ownerId, List<String> uploadIds) throws RemoteCommunicationException, DocumentVerificationException {
```
michal-rozehnal-w, line 138: @tech-kratas recap of performed checks:

1. ~~Both sides of each document must be uploaded. If any side is missing or more than two sides are provided → error.~~ this check together with few others is done in `RequiredDocumentTypesCheck` class
2. Order of documents used to extract the face photo: https://github.com/wultra/enrollment-server/blob/662b27b1d71fd40ac4877f6ea3d20cee64d034af/enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/enumeration/DocumentType.java#L74
3. If the identified document type differs from the one specified in the request → error.
4. Cross-check the fields FirstName, LastName, and DateOfBirth. If any of these values differ across documents → error.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
banterCZ: I am ok with that, it was just an idea.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/api/DocumentVerificationResponseParser.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private DocumentVerificationResponse response;
+
+    public DocumentVerificationResponseParser(String responseJson) throws JsonProcessingException {
+        this.responseJson = responseJson;
+
+        parseResponse(responseJson);
+    }
+
+    private void parseResponse(final String responseJson) throws JsonProcessingException {
```
banterCZ: 🤔  Couldn't be the parsen a singleton spring bean instead of recreating the mapper over again?

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkConfig.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+
+/**
+ * Configuration for Microblink {@link RestClient}.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@ConditionalOnProperty(value = "enrollment-server-onboarding.document-verification.provider", havingValue = "microblink")
```
banterCZ, line 42: Why not, but the key occurs in three different places. We could have just a single `MicroblinkConfig` with `@EnableProperties`.

enrollment-server-onboarding-provider-microblink/src/test/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProviderTest.java
```diff
@@ -0,0 +1,993 @@
...
+                new byte[] {4}
+        );
+
+        submittedDocumentIdCardFront = buildSubmittedDocument(verificationDocumentCardIdFront);
+        submittedDocumentIdCardBack = buildSubmittedDocument(verificationDocumentCardIdBack);
+
+        when(cacheManager.getCache("microblinkDocumentsCache")).thenReturn(verificationDataCache);
+        when(cacheManager.getCache("microblinkPhotoCache")).thenReturn(photoCache);
+
+        provider = new MicroblinkDocumentVerificationProvider(cacheManager, restClient);
```
banterCZ: Couldn't we use `@InjectMocks`?

banterCZ (approved):
Looks good.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -17,66 +17,427 @@
...
+                .filter(r -> "LastName".equals(r.field()))
+                .findFirst()
+                .map(com.wultra.app.onboardingserver.provider.microblink.api.DocumentVerificationResponse.Result::value)
+                .map(String::toLowerCase)
+                .orElseThrow(() -> new DocumentVerificationException("Field LastName not found in extracted data"));
+
+        final var dateOfBirth = documentExtractedData.stream()
+                .filter(r -> "DateOfBirth".equals(r.field()))
+                .findFirst()
+                .map(r -> LocalDate.of(r.year(), r.month(), r.day()))
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is valid. I wrapped the creation of `LocalDate` into separate method.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -17,66 +17,427 @@
...
+            final var overallExtraction = response.extraction().overall();
+            putDocumentCrosscheckData(overallExtraction, documentsCrosscheckData);
+
+            if (documentType == facePhotoExtractionDocumentType) {
+                final var faceImageBase64 = response.images()
+                        .stream()
+                        .filter(image -> image.name().equals("FaceImage"))
+                        .findFirst()
+                        .map(com.wultra.app.onboardingserver.provider.microblink.api.DocumentVerificationResponse.Image::base64)
+                        .orElseThrow(() -> new DocumentVerificationException("Face image not extracted from document of type %s".formatted(documentType)));
```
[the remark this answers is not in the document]
michal-rozehnal-w: Also valid, it is wrapped into `Optional` now

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkCacheConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Bean("microblinkCacheManager")
+    public CacheManager cacheManager(final MicroblinkConfigProperties properties) {
+        logger.info("Registering Microblink CacheManager");
+
+        final var caffeineBuilder = Caffeine.newBuilder()
+                .expireAfterWrite(properties.getCacheRecordTTL());
```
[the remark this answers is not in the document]
michal-rozehnal-w: Added default value directly into property

enrollment-server-onboarding-provider-microblink/src/test/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProviderTest.java
```diff
@@ -0,0 +1,993 @@
...
+                CardSide.BACK,
+                "driving_license_back.jpg",
+                new byte[] {4}
+        );
+
+        submittedDocumentIdCardFront = buildSubmittedDocument(verificationDocumentCardIdFront);
+        submittedDocumentIdCardBack = buildSubmittedDocument(verificationDocumentCardIdBack);
+
+        when(cacheManager.getCache("microblinkDocumentsCache")).thenReturn(verificationDataCache);
+        when(cacheManager.getCache("microblinkPhotoCache")).thenReturn(photoCache);
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is on purpose, I don't want to use values from class under the test.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -17,66 +17,427 @@
...
+                .map(com.wultra.app.onboardingserver.provider.microblink.api.DocumentVerificationResponse.Result::value)
+                .map(String::toLowerCase)
+                .orElseThrow(() -> new DocumentVerificationException("Field FirstName not found in extracted data"));
+
+        final var lastName = documentExtractedData.stream()
+                .filter(r -> "LastName".equals(r.field()))
+                .findFirst()
+                .map(com.wultra.app.onboardingserver.provider.microblink.api.DocumentVerificationResponse.Result::value)
+                .map(String::toLowerCase)
+                .orElseThrow(() -> new DocumentVerificationException("Field LastName not found in extracted data"));
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is not true. The `.findFirst()` returns `Optional` and any other call in chain is null safe. To be really sure I tried it locally and NPE is not thrown. Also IntelliJ shows warning for this suggestion, that `.filter(Objects::nonNull)` is always true.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/api/DocumentVerificationResponseParser.java
michal-rozehnal-w: Good idea, it is a bean now

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkConfig.java
michal-rozehnal-w: I agree, it is less error-prone to have it in a single place. Adjusted according to your suggestion. Thank you

enrollment-server-onboarding-provider-microblink/src/test/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProviderTest.java
michal-rozehnal-w: Unfortunately, the `@InjectMocks` doesn't work here. In the `@BeforeEach` there is stubbing for `cacheManager` and this method is executed after the annotation. This causes that each cache is set to `null` in constructor and each test throws NPE. The provider instance have to be created after the stubbing is in place, that's why it is in this method.

docs/onboarding/Configuration-Properties.md
tech-kratas: i think `1h` is fine. if it expires, user can reupload all documents again

PR #1361, opened 2025-11-19
Fix #1350: Store process type from onboarding start endpoint

### LINKED ISSUE #1350: Store process type from onboarding start endpoint
### Task specification

Pass _process type_ to onboarding start endpoint. The parameter must be optional for backward compatibility. If not filled, we expect a single process configuration. It was introduced in #1344
API change is defined by https://github.com/wultra/enrollment-server/issues/1358

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -438,29 +445,49 @@ private String lookupUser(final OnboardingProcessEntity process, final Map<Strin
...
         process.setTimestampCreated(new Date());
-        setProcessCustomData(process, fdsData, requestContext);
+        process.setProcessConfiguration(fetchProcessConfiguration(request.processType()));
+        setProcessCustomData(process, request.fdsData(), requestContext);
         return onboardingProcessRepository.save(process);
     }
 
+    private OnboardingProcessConfigurationEntity fetchProcessConfiguration(final String processType) throws OnboardingProcessException {
+        return onboardingProcessConfigurationRepository.findByProcessType(resolveProcessType(processType))
+                .orElseThrow(() -> new OnboardingProcessException("No configuration found for process type: " + processType));
```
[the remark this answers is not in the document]
banterCZ: It is acceptable.

enrollment-server-onboarding/src/main/resources/application-test.properties
```diff
@@ -1,18 +0,0 @@
-#
```
michal-rozehnal-w, line 1: Is the removal of this file intentional? Although all tests pass, it seems to be unrelated to the rest of the logic in this PR

banterCZ: Good catch, but yes, this removal is on purpose. It is shadowing the real test properties, and it is absolutely against the best practice in Spring Boot.

enrollment-server-onboarding/src/main/resources/application.properties
```diff
@@ -80,6 +80,7 @@ enrollment-server-onboarding.onboarding-process.activation.expiration=5m
 enrollment-server-onboarding.onboarding-process.verification.expiration=1h
 enrollment-server-onboarding.onboarding-process.max-processes-per-day=5
 enrollment-server-onboarding.onboarding-process.max-error-score=15
+#enrollment-server-onboarding.onboarding-process.default-type
```
michal-rozehnal-w: Here we can use some custom env variable since this is mandatory property. For example:
`enrollment-server-onboarding.onboarding-process.default-type=${ONBOARDING_PROCESS_DEFAULT_TYPE}`
However current implementation is also ok.

enrollment-server-onboarding/src/main/resources/application-test.properties
michal-rozehnal-w: This makes sense. Thank you.

michal-rozehnal-w (approved):

enrollment-server-onboarding/src/main/resources/application.properties
banterCZ: A good idea!

PR #1372, opened 2025-11-20
Fix #1367: Add Sonatype publishing configuration to pom.xml

### LINKED ISSUE #1367: Automatic publish in Sonatype
### Task specification

_No response_

filipbursik-wultra:


pom.xml
```diff
@@ -273,6 +274,11 @@
     <build>
         <pluginManagement>
             <plugins>
+                <plugin>
+                    <groupId>org.sonatype.central</groupId>
+                    <artifactId>central-publishing-maven-plugin</artifactId>
+                    <version>${central.publishing.plugin.version}</version>
```
banterCZ, line 280: ditto, I believe that we do not need this `pluginManagement` section, just define the version in the plugins section.

banterCZ (approved):

enrollment-server-onboarding/pom.xml
```diff
@@ -255,18 +255,8 @@
                 </property>
             </activation>
             <properties>
+                <skipPublishing>true</skipPublishing>
```
korbelm, line 258: i think we need to skip maven deploy plugin as well

filipbursik-wultra: Fixed

Issue #1385, opened 2025-11-24
Add Docker Image Build


filipbursik-wultra:
### Task specification

Set up Docker image building in the project as part of the CI process.

PR #1397, opened 2025-11-26
Add default values for datasource

banterCZ:
A follow-up to #1385


michal-rozehnal-w (approved):
Looks good. We can add env variables also for the `enrollment-server-onboarding.onboarding-adapter.*` properties. But it can be in separate PR

PR #1400, opened 2025-11-26
Issues/1387 add mock docker

korbelm:


.github/workflows/publish-docker-image.yml
```diff
@@ -25,6 +25,7 @@ on:
       - '**'
       - '!docs/**'
       - 'docs/db/**'
+      - '!enrollment-server-onboarding-adapter-mock/**'
```
zcgandcomp, line 28: We can actually set up only
'enrollment-server-onboarding-adapter-mock/**'

zcgandcomp (changes requested):
Optimize to build only if mock dir is changed

.github/workflows/publish-docker-image.yml
korbelm: mock is done in different pipeline

filipbursik-wultra (approved):
LGTM

zcgandcomp (approved):
The path setting is correct.

PR #1405, opened 2025-11-27
Fix #1404: Process consent endpoints in transaction

### LINKED ISSUE #1404: Get consent text endpoint returns 500

_No response_

### Environment

DEV

_No response_

### Log output

_No response_

michal-rozehnal-w:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
```diff
@@ -504,6 +504,7 @@ public Response cleanup(ObjectRequest<IdentityVerificationCleanupRequest> reques
      * @throws OnboardingProcessException Thrown when onboarding process is not found.
      * @throws PowerAuthEncryptionException Thrown when request decryption fails.
      */
+    @Transactional
```
michal-rozehnal-w, line 507: It causes problem [here](https://github.com/wultra/enrollment-server/blob/develop/enrollment-server-onboard[…]ra/app/onboardingserver/impl/service/OnboardingServiceImpl.java). The `process.getProcessConfiguration()` is lazy evaluation and outside of transaction it throws `LazyInitializationException`. Alternative way is to use eager loading

banterCZ (approved):
Thanks for the fix!

PR #1460, opened 2026-01-09
Fix #1459: Fix calling evaluate with correct param

### LINKED ISSUE #1459: Fix processId for RequiredDocumentTypesCheck evaluate method

_No response_

### Environment

_No response_

The `RequiredDocumentTypesCheck#evaluate` method should accept `processId` as one of parameters. But in code when method is called value of `identityVerificationId` which causes that record is not found in database.

### Log output

_No response_

michal-rozehnal-w:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteCheck.java
```diff
@@ -87,7 +87,7 @@ Result evaluate(final IdentityVerificationEntity idVerification) throws RemoteCo
             return Result.failed("Some documents not accepted");
         }
 
-        if (!requiredDocumentTypesCheck.evaluate(documentVerifications, identityVerificationId)) {
+        if (!requiredDocumentTypesCheck.evaluate(documentVerifications, processId)) {
```
michal-rozehnal-w, line 90: This was the root cause of several backend test failures. The method expects `processId`, not `identityVerificationId`. I am not sure how I missed that. I also adjusted and added tests to ensure this does not happen again.

banterCZ (approved):

PR #1465, opened 2026-01-12
Fix 1464: Return Bad Request for invalid base64

### LINKED ISSUE #1464: Return 400 bad request for invalid base64 document data
### Task specification

Currently the 500 error is returned when invalid base64 data are sent in request for endpoint `POST api/v2/identity/document/submit`. It should return 400 Bad Request

michal-rozehnal-w:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/errorhandling/DefaultExceptionHandler.java
```diff
@@ -251,4 +251,18 @@ public class DefaultExceptionHandler {
...
         return new ErrorResponse("ERROR_NOT_FOUND", "Resource not found.");
     }
+
+    /**
+     * Exception handler for illegal argument exception.
+     *
+     * @param e Exception.
+     * @return Response with error details.
+     */
+    @ExceptionHandler(IllegalArgumentException.class)
```
banterCZ: 🤔 Ok, this fixes the issue with invalid base64, but in general, `IllegalArgumentException` may be thrown elsewhere, and in some cases, it does not always have to be the client's error. Shouldn't we prefer bean validation?

banterCZ (commented):
Ok, that works, but have you considered bean validation instead of throwing `IllegalArgumentException`?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/errorhandling/DefaultExceptionHandler.java
michal-rozehnal-w: I tried using validation annotations, but they do not work correctly with encryption, so I created a [task](https://github.com/wultra/enrollment-server/issues/1466) for it.
Anyway, you are right that `IllegalArgumentException` is too broad, so I created a custom exception.

banterCZ: Thanks, custom exception looks good.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentProcessingService.java
```diff
@@ -456,14 +456,22 @@ private SubmittedDocument createSubmittedDocument(
...
         photo.setData(documentData);
 
         return submittedDoc;
     }
 
+    private static byte[] decodeBase64Data(final String base64Data) {
+        try {
+            return Base64.getDecoder().decode(base64Data);
+        } catch (final RuntimeException e) {
+            throw new IllegalArgumentException("Invalid base64 data", e);
```
banterCZ: Shouldn't you use the new exception here?

michal-rozehnal-w: Definitely, I missed the most important part 🙈 fixed and updated tests

banterCZ (approved):

PR #1468, opened 2026-01-15
Fix #1467: Fetch consent fails when the process in ACTIVATION_IN_PROG…

### LINKED ISSUE #1467: Fetch consent fails when the process in ACTIVATION_IN_PROGRESS

_No response_

### Environment

_No response_

Fetch consent fails when the process is in `ACTIVATION_IN_PROGRESS`

### Log output

_No response_

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
```diff
@@ -672,7 +672,7 @@ public ObjectResponse<OnboardingConsentTextResponse> fetchConsentText(
         logger.debug("Returning consent for {}", requestObject);
         OnboardingConsentTextRequestValidator.validate(requestObject);
 
-        final OwnerId ownerId = extractOwnerId(apiAuthentication);
+        final OwnerId ownerId = PowerAuthUtil.getOwnerId(apiAuthentication);
```
michal-rozehnal-w, line 675: Just note that `PowerAuthUtil#getOwnerId` does not check the `OnboardingStatus`, which is fine in my opinion, since calling this endpoint is independent of the onboarding process state (it doesn't move it to the next state).

michal-rozehnal-w (approved):

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
banterCZ: You are right, but we are verifying the process status three lines below.

```java
onboardingService.verifyProcessId(ownerId, processId, OnboardingStatus.NOT_YET_COMPLETED);
```

PR #1480, opened 2026-01-26
Fix #1479: Dockerfile and entrypoint script: Remove application.prope…

### LINKED ISSUE #1479: Use application.properties from war instead of copy to docker image
### Task specification

_No response_

korbelm:
…rties copy and update command


filipbursik-wultra (approved):
OK

PR #1507, opened 2026-02-05
Fix #1506: Add activationFlags into a response object

### LINKED ISSUE #1506: Add activationFlags into response object
### Task specification

_No response_

michal-rozehnal-w:


enrollment-server/src/main/resources/application.properties
```diff
@@ -82,6 +82,7 @@ enrollment-server.inbox.enabled=${ENROLLMENT_SERVER_INBOX_ENABLED:true}
 enrollment-server.activation-spawn.enabled=${ENROLLMENT_SERVER_ACTIVATION_SPAWN_ENABLED:false}
 enrollment-server.admin.enabled=${ENROLLMENT_SERVER_ADMIN_ENABLED:false}
 enrollment-server.auth-type=NONE
+enrollment-server.flag.reporting=true
```
zcgandcomp: please default as false

enrollment-server.flag.reporting=${ENROLLMENT_SERVER_FLAG_REPORTING:false}

zcgandcomp (changes requested):
Nice, one config remark

enrollment-server/src/main/resources/application.properties
michal-rozehnal-w: Done

zcgandcomp (approved):
The configuration for how to return flags in the vanilla Enrollment server.
The configuration is conditional.
The default behavior is the same as the current behavior.
The configuration option follows the pattern of other configurations.

PR #1536, opened 2026-02-09
Fix #1516: ApproveClient fails on null status (forward-port)

### LINKED ISSUE #1516: ApproveClient fails on null status

_No response_

### Environment

_No response_

Missing mapping in `RestOnboardingProvider`. Related to #1470 

### Log output

```
action: callApproveClient, state: initiated, attempt 2/3, previous failure: status is marked non-null but is null
```

zcgandcomp:


michal-rozehnal-w (approved):

banterCZ:
Thanks for fixing that.

PR #1543, opened 2026-02-12
Issues/front document

zcgandcomp:
ONLY FOR EXPLANATORY PURPOSES

### DIFF
--- a/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
+++ b/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
@@ -43,6 +43,9 @@ import org.springframework.transaction.annotation.Transactional;
 
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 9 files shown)


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
```diff
@@ -302,6 +311,28 @@ private static DocumentResultEntity selectLatestDocumentResult(final DocumentVer
                 .orElseThrow(() -> new IllegalStateException("Missing document result for %s".formatted(documentVerificationEntity)));
     }
 
+    private static LocalDate convertDate(final String date) {
```
michal-rozehnal-w, line 320: I recommend stream based approach than inner try-catch. For example:
```
private static final List<DateTimeFormatter> DATE_FORMATTERS = List.of(
        DateTimeFormatter.ofPattern("yyyy-MM-dd"),
        DateTimeFormatter.ofPattern("dd.MM.yyyy"),
        DateTimeFormatter.ofPattern("dd MM yyyy")
);
```

Then call use the constant in method:

```
private static LocalDate convertDate(String date) {
    if (date == null || date.isBlank()) {
        return null;
    }

    return DATE_FORMATTERS.stream()
            .map(f -> tryParse(date, f))
            .filter(Objects::nonNull)
            .findFirst()
            .orElse(null);
}

private static LocalDate tryParse(String date, DateTimeFormatter formatter) {
    try {
        return LocalDate.parse(date, formatter);
    } catch (DateTimeParseException e) {
        return null;
    }
}
```

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/rest/ClientEvaluateRequestDto.java
```diff
@@ -84,21 +86,21 @@ public record Document(
     public record Person(
```
michal-rozehnal-w, line 87: If you want to specify ISO format, you can use `@JsonFormat(pattern = "yyyy-MM-dd")` for each property in this DTO

PR #1546, opened 2026-02-17
Fix #1545: The OnboardingProviderException encapsulate the whole request object

### LINKED ISSUE #1545: The    OnboardingProviderException encapsulate the whole request object

_No response_

### Environment

_No response_

Fail in the process when the system raises OnboardingProviderException.

Expected result:
An error is thrown, with  important identifier(s)
Actual result:
An error is thrown with the whole request object (12kB).

This is a common pattern in the latest changes. Never log the whole object or use the whole object in exception message.

### Log output

_No response_

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/rest/RestOnboardingProvider.java
```diff
@@ -67,11 +67,11 @@ public LookupUserResponse lookupUser(final LookupUserRequest request) throws Onb
         try {
             response = restClient.post("/user/lookup", requestDto, null, createHeaders(), responseType).getBody();
         } catch (RestClientException e) {
-            throw new OnboardingProviderException("Unable to lookup user for " + request, e);
+            throw new OnboardingProviderException("Unable to lookup user, processId=%s".formatted(request.getProcessId()), e);
```
michal-rozehnal-w, line 70: I was thinking about creating separate method for creating exception instance, but I am not sure it will bring any benefit. Current approach is good

michal-rozehnal-w (approved):

zcgandcomp (approved):
Good from a content point of view. I have small concerns regarding the null safety..

PR #1556, opened 2026-02-19
Fix #1555: The ApproveClientResponseDto has mandatory resultReason

### LINKED ISSUE #1555: The ApproveClientResponseDto has mandatory resultReason

_No response_

### Environment

_No response_

Call Approval service that doesn't return resultReson

Expected result:
process continues

Actual Result
exception HTTP POST request failed

Root cause
ApproveClientResponseDto has marked resultReason as NonNull

### Log output

_No response_

banterCZ:


zcgandcomp (approved):
Looks ok.
Attribut is in compliance with contract.
Test can remain as is - we should start focusing on contract testing more.

michal-rozehnal-w (approved):
Looks good. I agree we should add test for mapping DTO to internal request model in future.

PR #1574, opened 2026-02-23
Fix #1563: Use OTP if generated and activation type is CODE

### LINKED ISSUE #1563: Use OTP if generated and activation type is CODE
### Task specification

If OTP for ACTIVATION of inital acrtiation is enabled (and generated) and the ActivationType is CODE, the generated OTP has to be propagated to the activation creation,

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -192,11 +189,12 @@ public OnboardingStartResponse startOnboarding(
             throw new TooManyProcessesException("Maximum number of processes per day reached for user: " + userId);
         }
 
-        createAndSendOtp(process, userId);
+        final Optional<String> otp = createAndSendOtp(process, userId);
```
michal-rozehnal-w, line 192: Nitpick, `var` can be used here

michal-rozehnal-w (approved):

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -213,18 +211,20 @@ public OnboardingStartResponse startOnboarding(
...
 
         final String otpCode = otpService.createOtpCode(process, OtpType.ACTIVATION);
         if (userId == null) {
             logger.info("User ID is null, OTP is not sent");
+            return Optional.empty();
         } else {
             logger.debug("Sending OTP for user ID: {}", userId);
             sendOtp(process, otpCode);
+            return Optional.of(otpCode);
         }
```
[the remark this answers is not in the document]
banterCZ: Interesting, but I do not want to change it in this PR. I think that the original reason for creating OTP even for non existing user is to avoid time attacks.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
banterCZ: Sure, but I am still quite old-fashioned and keep explicit typing. I use `var` in inner block, tests, etc.

PR #1597, opened 2026-02-27
Fix #1596: Cleaning onboarding data

### LINKED ISSUE #1596: Onboarding Data cleaning
### Task specification

_No response_

michal-rozehnal-w:
This is a draft, just to get some feedback on what is deleted and how.

TODOs:
- [x]  go through the entire flow and double-check the data we are storing
- [x]  add/update tests


enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -180,20 +180,7 @@ public Image getPhoto(String photoId) throws DocumentVerificationException {
...
-                .map(DocumentVerificationEntity::getPhotoId)
-                .filter(Objects::nonNull)
-                .collect(Collectors.toSet());
-
-        processedDocumentDataRepository.deleteAllById(facePhotoIds);
-
-        documentDataRepository.deleteAllById(uploadIds);
-
-        logger.info("action: cleanupDocuments, state: succeeded, provider: microblink");
+        logger.info("action: cleanupDocuments, state: succeeded, provider: microblink, ownerId: {}, uploadIds: {}", ownerId, String.join(",", uploadIds));
```
michal-rozehnal-w: Maybe we can use `state: skipped` or `state: ignored `instead of `state: succeeded`, since this does nothing.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -525,35 +523,6 @@ private void moveToDocumentUpload(final OwnerId ownerId, final IdentityVerificat
         moveToPhaseAndStatus(idVerification, DOCUMENT_UPLOAD, status, ownerId);
     }
 
-    private List<String> collectRejectionErrors(DocumentVerificationEntity entity) {
```
michal-rozehnal-w, line 528: Not related to this PR, removed, because IntelliJ marks it as not used

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -248,10 +248,21 @@ private void terminateAndAuditOtps(final List<String> otpIds, final Date now) {
...
-                documentVerificationRepository.findById(documentId).ifPresent(document ->
-                        auditService.audit(document, "Expired Document verification for user: {}, {}", document.getIdentityVerification().getUserId(), errorDetail)));
+        final var documentVerifications = documentVerificationRepository.findAllById(documentIds);
+        for (final var documentVerification : documentVerifications) {
+            documentVerification.setStatus(DocumentStatus.FAILED);
+            documentVerification.setErrorDetail(errorDetail);
+            documentVerification.setErrorOrigin(errorOrigin);
+            documentVerification.setTimestampLastUpdated(now);
+
+            documentVerification.getResults()
```
michal-rozehnal-w: For the `CleanTask`, we don’t take the value of `identityVerificationConfig.isDocumentVerificationCleanupEnabled()` into account. Everything is deleted after the expiration time. IMHO, this is fine and consistent with the behavior of other entities; I just wanted to mention it.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -572,4 +541,60 @@ private DocumentMetadataResponseDto toDocumentMetadata(DocumentVerificationEntit
         docMetadata.setType(entity.getType());
         return docMetadata;
     }
+
+    private void deleteDocumentData(final OwnerId ownerId, final List<DocumentVerificationEntity> documentVerifications) throws RemoteCommunicationException, DocumentVerificationException, IdentityVerificationNotFoundException {
```
michal-rozehnal-w: After a discussion with @banterCZ , we moved this from provider-specific to common logic, since the DB structure is generic and can be used by any provider in the future.

michal-rozehnal-w (commented):
This is a draft, just to get some feedback on what is deleted and how. 

TODOs:
- [x] go through the entire flow and double-check the data we are storing
- [x] add/update tests

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
banterCZ: 🤔 I would even skip logging at all, maybe just debug.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -453,19 +455,15 @@ public DocumentStatusResponse fetchDocumentStatusResponse(final DocumentStatusRe
     @Transactional
     public void cleanup(OwnerId ownerId)
             throws DocumentVerificationException, RemoteCommunicationException, IdentityVerificationException, OnboardingProcessLimitException, OnboardingProcessException {
-
-        List<String> uploadIds = documentVerificationRepository.findAllUploadIds(ownerId.getActivationId());
+        final var documentVerifications = documentVerificationRepository.findAllByActivationId(ownerId.getActivationId());
```
banterCZ: 🤔 Have you considered the performance? Before, it was done in a modifying query. I am not sure whether it is possible to realize all cases, just asking.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -572,4 +541,60 @@ private DocumentMetadataResponseDto toDocumentMetadata(DocumentVerificationEntit
...
+
+            if (identityVerificationConfig.isDocumentVerificationCleanupEnabled()) {
+                final var documentResults = documentVerification.getResults();
+                final var documentResultIds = documentResults.stream()
+                        .map(DocumentResultEntity::getId)
+                        .collect(Collectors.toSet());
+
+                logger.info("Cleaning DocumentResult ids: {}", documentResultIds);
+
+                documentResults.forEach(it -> {
```
banterCZ: Nitpick: I feel a bad taste of modifying in `.forEach`. I know that it is already used elsewhere, but...

banterCZ (commented):
Just some comments from the first reading.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
michal-rozehnal-w: Very good point. You’re right that we don’t need to fetch the entire entity for cleaning. I will try to refactor it.

enrollment-server-onboarding-provider-microblink/src/test/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProviderIntTest.java
```diff
@@ -392,32 +392,6 @@ void testGetPhoto_photoFound_correctImageIsReturned() throws Exception {
         assertEquals(expectedImage, actualImage);
     }
 
-    @Test
-    void testCleanupDocuments_verificationDataNotFound_noExceptionIsThrown() {
```
michal-rozehnal-w, line 403: Now covered in `IdentityVerificationRestServiceIntTest`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -247,11 +248,16 @@ private void terminateAndAuditOtps(final List<String> otpIds, final Date now) {
...
-    private void terminateAndAuditDocuments(final List<String> documentIds, final Date now, final String errorDetail, final ErrorOrigin errorOrigin) {
-        documentVerificationRepository.terminate(documentIds, now, errorDetail, errorOrigin);
-        documentIds.forEach(documentId ->
-                documentVerificationRepository.findById(documentId).ifPresent(document ->
-                        auditService.audit(document, "Expired Document verification for user: {}, {}", document.getIdentityVerification().getUserId(), errorDetail)));
+    private void terminateAndAuditDocuments(final List<String> documentVerificationIds, final Date now, final String errorDetail, final ErrorOrigin errorOrigin) {
+        documentVerificationRepository.terminate(documentVerificationIds, now, errorDetail, errorOrigin);
+        documentResultRepository.clean(documentVerificationIds);
+
+        final var documentVerifications = documentVerificationRepository.findAllById(documentVerificationIds);
```
michal-rozehnal-w, line 255: I replaced `findById` called in `forEach` by single `findAllById`. It should be ok as the batch size of ids list is not huge. But let me know if you think the `forEach` is better here.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestServiceIntTest.java
```diff
@@ -119,4 +144,96 @@ void testSubmitDocumentsV2_documentsAreResubmitted_responseOk() throws Onboardin
...
+     * - enrollment-server-onboarding.presence-check.enabled=true
+     * - enrollment-server-onboarding.presence-check.cleanupEnabled=true
+     * During cleanup, none of these limits are reached:
+     * - enrollment-server-onboarding.onboarding-process.max-error-score
+     * - enrollment-server-onboarding.identity-verification.max-failed-attempts
+     *
+     * @throws Exception Any exception that occurs during the test
+     */
+    @Test
+    void testCleanup_fullDocumentVerificationAndPresenceCheckCleanup_allDataAreCleaned() throws Exception {
```
michal-rozehnal-w, line 171: For this test, I think it is useful to describe it, since there is quite complex logic regarding what is deleted and updated based on the configuration.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestServiceIntTest.java
```diff
@@ -119,4 +144,96 @@ void testSubmitDocumentsV2_documentsAreResubmitted_responseOk() throws Onboardin
...
+
+        // when
+        tested.cleanup(new ObjectRequest<>(request), apiAuthentication);
+
+        // then
+        assertDocumentVerificationAndPresenceCheckCleanup();
+    }
+
+    private void assertDocumentVerificationAndPresenceCheckCleanup() throws Exception {
+        final var documentDataCount = jdbcTemplate.queryForObject("SELECT COUNT(*) FROM es_document_data", Integer.class);
```
michal-rozehnal-w, line 194: It is probably not necessary to use `jdbcTemplate`, but since `*Repository` classes are used in the code being tested, I wanted to use something else to check the database.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
michal-rozehnal-w: I just left it on `debug` level, but you are right. Since this method does nothing there is no reason to log on info `level`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
michal-rozehnal-w: Removed, only delete queries are used

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentVerificationIdsView.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.wultra.app.onboardingserver.common.database.entity;
+
+/**
+ * View for ids related to {@link DocumentVerificationEntity}.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public interface DocumentVerificationIdsView {
+    String getUploadId();
```
banterCZ, line 33: I know, it is self-describing, but still, interfaces should have JavaDoc.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -572,4 +539,35 @@ private DocumentMetadataResponseDto toDocumentMetadata(DocumentVerificationEntit
...
         return docMetadata;
     }
+
+    private void deleteDocumentData(final OwnerId ownerId, final List<DocumentVerificationIdsView> idsViews) throws RemoteCommunicationException, DocumentVerificationException, IdentityVerificationNotFoundException {
+        final var uploadIds = idsViews.stream()
+                .map(DocumentVerificationIdsView::getUploadId)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+
+        logger.info("Deleting DocumentData ids: {}", uploadIds);
```
banterCZ, line 549: If I recall correctly, there is a limit in Oracle for `IN` statement, 1000 elements only. Shouldn't be safe here?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -572,4 +539,35 @@ private DocumentMetadataResponseDto toDocumentMetadata(DocumentVerificationEntit
...
+
+        final var processedDocumentIds = idsViews.stream()
+                .map(DocumentVerificationIdsView::getPhotoId)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+
+        logger.info("Deleting ProcessedDocumentData ids: {}", processedDocumentIds);
+        processedDocumentDataRepository.deleteAllById(processedDocumentIds);
+
+        documentVerificationProvider.cleanupDocuments(ownerId, uploadIds.stream().toList());
```
banterCZ, line 560: Why do we have log statements for others but not here?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
banterCZ: I am not a database expert, but in general, it should be better for the engine to execute one bigger query than several smaller ones. On the other hand, loading a huge amount of data demands more memory for Java. But IMHO it should be ok for this case. Except for the Oracle limit for `IN`.

banterCZ (commented):
In general, looks good. Let's review it once again when you return.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentVerificationIdsView.java
michal-rozehnal-w: You are right, added

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
michal-rozehnal-w: IMHO, it should be fine because this method is called by the endpoint `/api/identity/cleanup`. In real life, there will be just a few IDs. I presume it shouldn’t be possible to create hundreds of them, since we reach some verification limit after a few attempts.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
michal-rozehnal-w: I left the logging because this is call to a provider and I expect the logging should be handled there (so I wanted to avoid duplication). Anyway I can add it here too if you see benefit in it.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
michal-rozehnal-w: I agree (although my DB knowledge is also limited) and I remember that having hundreds of items in an `IN` array is quite slow and leads to poor performance. However this is maybe no longe true.
We have a hardcoded maximum batch size of 1000. It is defined in the constant `BATCH_SIZE`.

banterCZ (approved):

PR #1600, opened 2026-03-03
Fix #1515: Transition to Onboarding Approval

### LINKED ISSUE #1515: Transition to Onboarding Approval
### Description

After the end of the presence check:

- If the check is OK  go to the  decision point for OTP Verification (PSEUDO SCA attribute applies)
- IF the check is not OK, go to the INIT PRESENCE CHECK (to repeat the presence check)
- if OTP Verification is enabled, go to the OTP Verification (or skip)
- After the OTP Verification is done (or skipped)
- If client approval is enabled, execute client approval (or skip)
- After the client approval is done OK (or skipped) fo to the FINIS_ACTIVATION stages

The diagram  APPROVAL LAST FLOW applies (it should be consistent with the description above)
The process scores also apply (no changes expected)

### Acceptance criteria

_No response_

### QA specification

_No response_

banterCZ:
Flow - Enrollment - Identity Verification - Approval Last Flow 2


enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessConfigurationValue.java
```diff
@@ -66,6 +68,7 @@ public static class OnboardingProcessConfigurationValueBuilder {
             activationType = ActivationType.IDENTITY;
             approvalEnabled = false;
             clientEvaluationEnabled = true;
+            pseudoScaEnabled = true;
```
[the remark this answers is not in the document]
banterCZ: Yes, the default is true because of the backward compatibility with the existing projects.

enrollment-server-onboarding-common/src/test/resources/com/wultra/app/onboardingserver/common/database/OnboardingProcessConfigurationRepositoryTest.sql
```diff
@@ -1,3 +1,3 @@
 INSERT INTO es_onboarding_process_configuration (id, process_type, config) VALUES
     (1, 'reactivation', '{"enabled":true,"activationType":"CODE","otpForIdentification":true,"otpForIdentityVerification":true,"useTemporaryActivation":true,"approvalEnabled":true,"documents":{"totalRequiredDocumentsCount":2,"groups":[{"requiredDocumentsCount":1,"items":[{"type":"ID_CARD","sideCount":2},{"type":"PASSPORT","sideCount":1}]},{"requiredDocumentsCount":0,"items":[{"type":"DRIVING_LICENCE","sideCount":1}]}]}}'),
-    (2, 'onboarding', '{"unknown": "value"}');
+    (2, 'onboarding', '{"unknown":"value","pseudoScaEnabled":false}');
```
michal-rozehnal-w: Maybe we can move the property `"pseudoScaEnabled": false` to the record `1` and keep the record `2` purely for defaults

banterCZ:
End-to-end tests passed.

michal-rozehnal-w (commented):
Looks good from an implementation perspective. I’m just not sure about the term `pseudo SCA`. Personally, I would avoid it because I see SCA as a pure boolean: either we have it, and we call it SCA, or we don’t have it, and the SCA term shouldn’t be used. The `pseudo` prefix could cause confusion about what it really means.
On the other hand, I don’t have a better idea for a replacement. Maybe something like `hidePresenceCheckResultBeforeValidOTP`, but I’m not sure what impact that would have on the implementation.
This is just my personal preference - I am okay if we decide to use the `pseudo SCA`. In that case, I recommend describing it more thoroughly in the documentation and mentioning some use cases where it makes sense to have it enabled or disabled.

enrollment-server-onboarding-common/src/test/resources/com/wultra/app/onboardingserver/common/database/OnboardingProcessConfigurationRepositoryTest.sql
banterCZ: Thanks for the opinion. The term `pseudoSca` was introduced by @zcgandcomp. Let's wait for his review.

banterCZ: cc @tech-kratas

michal-rozehnal-w (approved):
Approved, because from technical perspective it is ok. If we want to keep the name `pseudoSca`, then we can describe it in https://github.com/wultra/enrollment-server/blob/develop/docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md#configuration or in https://github.com/wultra/enrollment-server/blob/develop/docs/onboarding/Database-Structure.md#configuration

enrollment-server-onboarding-common/src/test/resources/com/wultra/app/onboardingserver/common/database/OnboardingProcessConfigurationRepositoryTest.sql
banterCZ: @michal-rozehnal-w 🤔 For the current reactivation, it is set to `true`, but you are right, that here we test default values. I will update that.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessConfigurationValue.java
```diff
@@ -38,6 +38,7 @@
  * @param approvalEnabled            Whether to call bank systems to approve the client.
  * @param documents                  List of documents.
  * @param clientEvaluationEnabled    Whether to call bank systems to evaluate the client.
+ * @param verifyPresenceWithOtp      Whether to verify the presence with OTP. If {@code true}, hide the result of presence check until valid OTP is entered.
```
michal-rozehnal-w, line 41: I was thinking we could add a warning like: “Setting this to false breaks SCA” or something similar. On the other hand, this might not be necessary in the Javadoc.
However, as this database configuration grows, we should probably have more detailed documentation for it. Maybe it already exists, but so far I am only avare of the [migration guide](https://github.com/wultra/enrollment-server/blob/develop/docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md). This is definitely out of scope for this PR—I just wanted to highlight it so it doesn’t get forgotten. cc @tech-kratas

michal-rozehnal-w (approved):
Thank you for renaming the `pseudSca`

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessConfigurationValue.java
zcgandcomp: Do not refer to SCA, as this doesn't break the SCA, and SCA was not actually SCA in a sence it was designed before. So Setting this to false breaks SCA” is a false statement.

As is it written is true, and fine.

PR #1606, opened 2026-03-10
Fix #1603: Return document images in evaluation service call

### LINKED ISSUE #1603: Return document images in Evaluation Service call
### Task specification

Call to the [Client Evaluation Service](https://github.com/wultra/enrollment-server/blob/develop/docs/onboarding/External-Onboarding-Services.md#client-evaluation-service) currently contains only `FACE` image extracted from the documents.

We need to add `FRONT` and `BACK` document images (back side applies only for two-sided documents) . It's pretty simple from the request/api perspective because `documentCheckResult.documents[].images[]` is array so we only add more items.

**Logic**

We have to options where to get front/back sides of the document:
1. images returned from verification provider (Microblink)
2. images already stored in the system from the mobile device and used for verification

We would like to combine both options so we need to add configuration of the image source.

**Backend**

1. Add [configuration](https://github.com/wultra/enrollment-server/blob/develop/docs/onboarding/Configuration-Properties.md#client-evaluation-configuration) for image source provider, eg. `enrollment-server-onboarding.client-evaluation.document-image-source` with values `PROVIDER` (default) and `ORIGINAL`
2. We don't use front/back images from the Microblink currently so we need to extend verification call with `options.returnFullDocumentImage` (boolean) parameter. Set `true` if the image source is set to `PROVIDER`.
3. We need to store images from verification response to the `es_processed_document_data` table (it's already user for `FACE` image). Images for image source `ORIGINAL` are stored in the table `es_document_data`
4. Modify Client Evaluation Service call accordingly

**Test**

Test which images are returned based on configuration.

Images from verification provider should be cropped and smaller in size (<500KB). Original images from mobile device should be "uncropped" with bigger size (>1MB).

michal-rozehnal-w:


docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260309-processed-document-data-verification-id.xml
```diff
@@ -0,0 +1,33 @@
...
+    <changeSet id="1" logicalFilePath="enrollment-server-onboarding/2.0.x/20260309-processed-document-data-verification-id.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="es_processed_document_data" columnName="document_verification_id" />
+            </not>
+        </preConditions>
+        <comment>Creates a new column document_verification_id in es_processed_document_data</comment>
+        <addColumn tableName="es_processed_document_data">
+            <column name="document_verification_id" type="varchar(36)">
+                <constraints foreignKeyName="es_processed_document_data_document_verification_id_fk"
```
michal-rozehnal-w, line 15: The NOT NULL constraint is intentionally not applied because the table may already contain data

enrollment-server-onboarding-adapter-mock/src/test/wiremock/__files/body-document-verification-response.json
```diff
@@ -673,6 +673,14 @@
...
+    },
+    {
```
michal-rozehnal-w, line 677: @zcgandcomp Just for your information, there is a change in the mock response. The front and back side images are now included in the response.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -439,18 +523,45 @@ final String saveFacePhoto(final Map<DocumentType, DocumentVerificationParsedRes
...
+            final String dataBase64
+    ) {
+        final var dataBytes = convert(dataBase64, dataType);
+
+        final var entity = new ProcessedDocumentDataEntity();
+        entity.setId(UUID.randomUUID().toString());
+        entity.setData(dataBytes);
+        entity.setDataType(dataType);
+        entity.setTimestampCreated(new Date());
+        entity.setDocumentVerificationId(documentVerificationId);
```
michal-rozehnal-w, line 574: I was thinking about the best mapping. In `es_document_verification`, we already had `photo_id` (which is the ID from `es_processed_document_data`). One option was to add a new column, for example `document_image_id`, but this is not maintainable in the future, as processed data types may grow, and we probably don’t want to add a new column every time.
Mapping records via documentVerificationId gives us flexibility and it is same pattern as used for `es_document_result` table.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -439,18 +523,45 @@ final String saveFacePhoto(final Map<DocumentType, DocumentVerificationParsedRes
...
+
+        return entity;
+    }
+
+    private static byte[] convert(final String dataBase64, final ProcessedDocumentDataType dataType) {
+        try {
+            return Base64.getDecoder().decode(dataBase64);
+        } catch (final RuntimeException e) {
+            logger.warn("Exception when decoding base64 data for data type: {}", dataType);
+            return new byte[0];
```
michal-rozehnal-w, line 584: Sonar recommends return empty array instead of `null`.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -439,18 +523,45 @@ final String saveFacePhoto(final Map<DocumentType, DocumentVerificationParsedRes
             return null;
         }
 
-        final var facePhotoDocumentId = UUID.randomUUID().toString();
-        final var facePhotoDocumentData = new ProcessedDocumentDataEntity();
-        facePhotoDocumentData.setId(facePhotoDocumentId);
-        facePhotoDocumentData.setData(Base64.getDecoder().decode(faceImageBase64));
-        facePhotoDocumentData.setDataType(ProcessedDocumentDataType.FACE_IMAGE);
-        facePhotoDocumentData.setTimestampCreated(new Date());
+        final var documentVerificationId = documentVerificationIdByDocumentTypeAndSide.getOrDefault(facePhotoDocumentType, Map.of())
```
michal-rozehnal-w: The face photo image is always linked with `documentVerificationId` for the front side. There is a discrepancy between the Microblink provider (which does not distinguish between sides and treats documents as a single entity) and our database (where we have a record per document type and side).

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesCheck.java
```diff
@@ -50,9 +49,7 @@
 @Slf4j
 public class RequiredDocumentTypesCheck {
 
-    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
-            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
-
+    private final ObjectMapper objectMapper;
```
michal-rozehnal-w, line 52: I used the injected mapper because there was an error when deserializing the local date field

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkConfigProperties.java
```diff
@@ -55,6 +56,11 @@ public class MicroblinkConfigProperties {
     @Valid
     private List<SdkConfig> mobileSdkConfigs = new ArrayList<>();
 
+    /**
+     * The {@code options} field used in REST API request body.
+     */
+    private DocumentVerificationProcessingOptions requestOptions;
+
```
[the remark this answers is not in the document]
michal-rozehnal-w: There are defaults in `application.properties`. Sending `"options": null` is valid state and provider uses its default values.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -555,20 +666,19 @@ private static DocumentVerificationImageSource buildImageSource(final DocumentVe
...
-        options.setReturnFaceImage(true);
-
         final var useCase = new DocumentVerificationUseCaseOptions();
 
         final var request = new DocumentVerificationRequest();
         request.setImageFront(frontImageSource);
         request.setImageBack(backImageSource);
-        request.setOptions(options);
+        request.setOptions(requestOptions);
         request.setUseCase(useCase);
```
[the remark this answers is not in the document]
michal-rozehnal-w: The `null` value is valid as I mentioned above.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -439,18 +523,45 @@ final String saveFacePhoto(final Map<DocumentType, DocumentVerificationParsedRes
...
+        }
 
-        processedDocumentDataRepository.save(facePhotoDocumentData);
+        final var facePhotoDocumentData = buildProcessedDocumentDataEntity(ProcessedDocumentDataType.FACE_IMAGE, documentVerificationId, faceImageBase64);
 
+        final var facePhotoDocumentId = facePhotoDocumentData.getId();
         logger.info("Face photo extracted from document type {} and stored with id {}", facePhotoDocumentType, facePhotoDocumentId);
 
-        return facePhotoDocumentId;
+        return facePhotoDocumentData;
```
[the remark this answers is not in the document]
michal-rozehnal-w: I will double check that

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -412,14 +413,97 @@ private List<DocumentSubmitResult> processMicroblinkResults(
...
+            processedDocumentData.add(facePhoto);
+        }
+        processedDocumentData.addAll(documents);
+
+        processedDocumentDataRepository.saveAll(processedDocumentData);
+
+        return Optional.ofNullable(facePhoto)
+                .map(ProcessedDocumentDataEntity::getId)
+                .orElse(null);
+    }
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is ok, we have cleanup job and it is also possible to do the cleanup via calling REST API endpoint

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -439,18 +523,45 @@ final String saveFacePhoto(final Map<DocumentType, DocumentVerificationParsedRes
...
+        entity.setDocumentVerificationId(documentVerificationId);
+
+        return entity;
+    }
+
+    private static byte[] convert(final String dataBase64, final ProcessedDocumentDataType dataType) {
+        try {
+            return Base64.getDecoder().decode(dataBase64);
+        } catch (final RuntimeException e) {
+            logger.warn("Exception when decoding base64 data for data type: {}", dataType);
```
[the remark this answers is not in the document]
michal-rozehnal-w: Good catch, fixed

docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260309-processed-document-data-verification-id.xml
```diff
@@ -0,0 +1,33 @@
...
+    <changeSet id="2" logicalFilePath="enrollment-server-onboarding/2.0.x/20260309-processed-document-data-verification-id.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists tableName="es_processed_document_data_document_verification_id_idx" />
+            </not>
+        </preConditions>
+        <comment>Creates a new index es_processed_document_data_document_verification_id_idx</comment>
+        <createIndex tableName="es_processed_document_data" indexName="es_processed_document_data_document_verification_id_idx">
+            <column name="document_verification_id" />
+        </createIndex>
```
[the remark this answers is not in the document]
michal-rozehnal-w: Correct, fixed

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesCheck.java
```diff
@@ -175,29 +172,29 @@ private static DocumentValidationItem convert(final OnboardingProcessConfigurati
...
             final var extractedData = documentResult.getExtractedData();
             if (extractedData == null) {
                 return null;
             }
 
-            return OBJECT_MAPPER.readValue(extractedData, DocumentExtractedDataValue.class)
+            return objectMapper.readValue(extractedData, DocumentExtractedDataValue.class)
                     .country();
         } catch (JsonProcessingException e) {
             logger.warn("Failed to parse extracted data for document result id {}", documentResult.getId(), e);
```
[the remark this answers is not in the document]
michal-rozehnal-w: The result of `objectMapper.isEnabled(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)` returns `false`

docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/db.changelog-version.xml
```diff
@@ -9,5 +9,6 @@
     <include file="20260126-selfie.xml" relativeToChangelogFile="true" />
     <include file="20260205-audit.xml" relativeToChangelogFile="true" />
     <include file="20251215-add-tag-2.0.0.xml" relativeToChangelogFile="true" />
+    <include file="20260309-processed-document-data-verification-id.xml" relativeToChangelogFile="true" />
```
banterCZ, line 11: Shouldn't this be placed before the 2.0 tag?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/ProcessedDocumentDataEntity.java
```diff
@@ -71,6 +71,12 @@ public class ProcessedDocumentDataEntity implements Serializable {
     @Column(name = "timestamp_created", nullable = false)
     private Date timestampCreated;
 
+    /**
+     * Associated {@link DocumentVerificationEntity}.
+     */
+    @Column(name = "document_verification_id")
```
banterCZ, line 80: 🤔 I have to recall how I managed validation of new data without database constraints affecting existing records.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/DocumentVerificationRepository.java
```diff
@@ -123,4 +125,11 @@ public interface DocumentVerificationRepository extends JpaRepository<DocumentVe
             WHERE d.uploadId IN :uploadIds
     """)
     List<DocumentVerificationEntity> findAllByUploadIds(final List<String> uploadIds);
+
+    @Query("""
```
banterCZ, line 129: Why not, but do we really need to specify the query if using Spring Data?

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -412,14 +413,97 @@ private List<DocumentSubmitResult> processMicroblinkResults(
...
+
+        for (final var entry : microblinkResponseByDocumentType.entrySet()) {
+            final var documentType = entry.getKey();
+
+            final var images = Optional.ofNullable(entry.getValue())
+                    .map(DocumentVerificationParsedResponse::images)
+                    .orElse(List.of());
+
+            for (final var image : images) {
+                final var imageName = image.name();
```
banterCZ: 🤔 Quite a long _for statement_. Could we extract a method creating `processedDocumentEntity` and keep `processedDocuments.add(processedDocumentEntity);` (or even better, change it to a collector)?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesCheck.java
banterCZ: Sure, no problem. The root cause is, I guess, that `new ObjectMapper()` does not register the Java Time module by default.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesCheckTest.java
```diff
@@ -47,12 +48,18 @@
...
     private OnboardingProcessConfigurationService onboardingProcessConfigurationService;
 
-    @InjectMocks
     private RequiredDocumentTypesCheck tested;
 
+    @BeforeEach
+    void setUp() {
+        tested = new RequiredDocumentTypesCheck(OBJECT_MAPPER, onboardingProcessConfigurationService);
+    }
+
```
banterCZ:
```suggestion
    @Spy
    @SuppressWarnings("unused")
    private final ObjectMapper objectMapper = new ObjectMapper();

    @Mock
    private OnboardingProcessConfigurationService onboardingProcessConfigurationService;

    @InjectMocks
    private RequiredDocumentTypesCheck tested;

```

banterCZ (approved):

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/ProcessedDocumentDataEntity.java
banterCZ: I found it, https://github.com/wultra/enrollment-server/pull/1437/changes

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/DocumentVerificationRepository.java
michal-rozehnal-w: I don’t think it’s necessary; however, I wanted to be consistent with the rest of the code, where we always have the query present.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/ProcessedDocumentDataEntity.java
michal-rozehnal-w: Thank you, added impl note and annotated with `NotNull`

docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/db.changelog-version.xml
michal-rozehnal-w: Good catch, I changed the order

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesCheck.java
michal-rozehnal-w: Yes, exactly

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/guard/document/RequiredDocumentTypesCheckTest.java
michal-rozehnal-w: I wasn't sure about using Spy, since there are arguments against it, but the suggestion seems more elegant.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
michal-rozehnal-w: You are right, I moved it into separate method and used `flatMap`

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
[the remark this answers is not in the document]
michal-rozehnal-w: In DB there isn't NOT NULL constraint for the column, so I adjusted the implementation and `side=null` won't cause any issue. I also added unit test for this scenario

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -412,14 +413,97 @@ private List<DocumentSubmitResult> processMicroblinkResults(
...
+                final var documentVerificationId = documentVerificationIdByDocumentTypeAndSide.getOrDefault(documentType, Map.of())
+                        .getOrDefault(documentSide, null);
+
+                if (documentVerificationId == null) {
+                    logger.warn("Document verification id not found document '{}' of side '{}'", documentType, documentSide);
+                }
+
+                final var processedDocumentEntity = buildProcessedDocumentDataEntity(processedDocumentType, documentVerificationId, image.base64());
+                processedDocuments.add(processedDocumentEntity);
+            }
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed as described above

PR #1618, opened 2026-03-16
Fix #1402: Consent should be optional

### LINKED ISSUE #1402: Consent should be optional
### Task specification

The consent step should be optional. Ideally, it should be taken from the process configuration, see #1364

After introducing _activation code_ in #1359 it does not work at all. I may be related to #1401

Right now, we call
- `/api/identity/consent/text`
- `/api/identity/consent/approve`

We cannot introduce a new state in the machine due to backward compatibility.

We need to extend the process configuration to determine whether consent is required.
The `identity/status` response returns `consentRequired=false` if consent is already approved or not required.
When `consent/approve` is called, record that the consent is approved and reflect that in the status response.

michal-rozehnal-w:
TODOs:

- [x] ~~Confirm with mobile team whether to rename the field `consentRequired`~~ We will use the `consentRequired` because of FE compatibility


docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260313-consent-accepted.xml
```diff
@@ -0,0 +1,18 @@
...
+
+    <changeSet id="1" logicalFilePath="enrollment-server-onboarding/2.0.x/20260313-consent-accepted.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="es_onboarding_process" columnName="consent_accepted" />
+            </not>
+        </preConditions>
+        <comment>Creates a new column consent_accepted in es_onboarding_process</comment>
+        <addColumn tableName="es_onboarding_process">
+            <column name="consent_accepted" type="boolean" />
```
michal-rozehnal-w, line 14: I considered adding a `not-null` constraint and setting the default value to `false`, but the information could be incorrect for legacy records because we don't persist consent acceptance from the client (it is only sent to the onboarding provider). That's why I decided to keep the column nullable in the database but treat it as non-null for new records in the code.

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/IdentityVerificationStatusResponse.java
```diff
@@ -51,4 +51,6 @@ public class IdentityVerificationStatusResponse {
     @Schema(description = "Reject reason. Apply for phases such as client evaluation, presence check, or client approval")
     private String rejectReason;
 
+    @Schema(description = "Specifies whether consent is required and pending.")
+    private boolean consentRequired;
```
michal-rozehnal-w, line 55: I propose renaming the property in the `api/identity/status` response from `consentRequired` to `consentPending` because the field can have the following states:
- Consent is not required in the process configuration → false
- Consent is required but not accepted → true
- Consent is required and accepted → false

We could also add `consentRequired` to `api/configuration` if this information is needed on mobile. The difference would then be:

- `api/identity/status`, `consentPending` – depends on the specific process and user; indicates whether consent is required by the process configuration and has not yet been accepted by the user.
- `api/configuration`, `consentRequired` – value defined in the process configuration, same value for all processes derived from the specific process config

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -250,7 +248,7 @@ private void configureInitialTransition(StateMachineTransitionConfigurer<Onboard
                 .withExternal()
                 .source(OnboardingState.INITIAL)
                 .event(OnboardingEvent.IDENTITY_VERIFICATION_INIT)
-                .guard(processIdentifierGuard)
+                .guard(createCompositeGuard(processIdentifierGuard, consentPendingGuard))
```
michal-rozehnal-w: Because we don't want to introduce a new state, I added an extra guard for consent.
These guards could be merged into one, since both fetch the process by ID from the repository. I don't think this is a performance issue because JPA should handle caching and the database should not actually be called twice (but please correct me if I'm wrong).
However, in my opinion it is cleaner to keep them separate and have multiple small guards rather than a single complex one.
If consent is required but not accepted the mobile know this information from response of `api/identity/status`

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessConfigurationValue.java
```diff
@@ -52,7 +52,7 @@ public record OnboardingProcessConfigurationValue(
         ActivationType activationType,
         boolean approvalEnabled,
         boolean clientEvaluationEnabled,
-        boolean verifyPresenceWithOtp
+        boolean consentRequired
```
banterCZ, line 57: 🤔 So the default will be false? Why not, but I have created `verifyPresenceWithOtp=true` because of backward compatibility. But the fact is that, for example, `otpForIdentification` is `false`. We have to prepare a database script for the current customer anyway. So, probably `verifyPresenceWithOtp` should be changed to false. What do you think?

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/IdentityVerificationStatusResponse.java
banterCZ, line 55: @marek-ch @kober32 This is important for you. Could you approve, please?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/ConsentResolvedGuard.java
```diff
@@ -0,0 +1,60 @@
...
+
+        if (process == null) {
+            logger.warn("Process not found, processId: {}", processId);
+            return false;
+        }
+
+        try {
+            final var consentPending = identityVerificationStatusService.isConsentPending(process);
+            logger.info("Consent pending: {}, processId: {}", consentPending, processId);
+            return !consentPending;
```
banterCZ, line 54: 🤔 I am a little bit confused. Why is the guard called `ConsentPending` and the variable `consentPending` is negated?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
banterCZ: Ok, let's keep it now, but we should reconsider `ProcessIdentifierGuard`. It does not make much sense to do such verification for each state machine step, but just at the start. I have opened an issue for that https://github.com/wultra/enrollment-server/issues/1619

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessConfigurationValue.java
michal-rozehnal-w: Yes, and I also wanted to keep backward compatibility because consent is already optional from the BE perspective. It is entirely up to the mobile app whether it calls `/api/identity/consent/approve`, and previously there was no check on the BE side. However, if `consentRequired` is now set to `true` in the process configuration and the user does not accept it, the state machine will not proceed to document upload because of the new guard.
About the `verifyPresenceWithOtp`, I am ok with changing it. Theoretically, we can set the default values as we want now; we just need to reflect them correctly in the customers' scripts. Respecting the current behavior makes it a little easier.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/ConsentResolvedGuard.java
michal-rozehnal-w: I admit this is a little tricky because if there is no pending consent, the guard should return `true` in order to pass. When `identityVerificationStatusService.isConsentPending(process)` returns `true`, it means we are waiting for the user to accept the consent, so the guard should return `false` in order not to move to the next state in the state machine.
However good point about the class name, I can rename it to `ConsentNotPendingGuard` to better reflect the boolean result from `evaluate` method. I wanted to avoid `ConsentAccepted` because if consent it not required by process then `accept` is not exactly correct since user doesn't accept it, just skipped it. Do you have any other suggestions?

banterCZ: Not sure. What about this?

```
/**
 * Guard that checks whether the consent has been resolved (not required or already accepted).
 */
ConsentResolvedGuard
```

michal-rozehnal-w: I like it, renamed to `ConsentResolvedGuard`

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/IdentityVerificationStatusResponse.java
```diff
@@ -51,4 +51,6 @@ public class IdentityVerificationStatusResponse {
     @Schema(description = "Reject reason. Apply for phases such as client evaluation, presence check, or client approval")
     private String rejectReason;
 
+    @Schema(description = "Specifies whether consent is required and pending.")
```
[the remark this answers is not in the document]
michal-rozehnal-w: Here we are waiting for feedback from mobile team whether we can rename it

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationStatusService.java
```diff
@@ -110,13 +95,33 @@ public IdentityVerificationStatusResponse checkIdentityVerificationStatus(Identi
         }
 
         final IdentityVerificationEntity idVerification = idVerificationOptional.get();
+        final var consentPending = isConsentPending(onboardingProcess);
+
         response.setIdentityVerificationStatus(idVerification.getStatus());
         response.setIdentityVerificationPhase(idVerification.getPhase());
         response.setRejectReason(idVerification.getRejectReason());
+        response.setConsentRequired(consentPending);
         return response;
```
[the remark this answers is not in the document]
michal-rozehnal-w: As mentioned above, renaming the field is possible option

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestServiceIntTest.java
```diff
@@ -251,4 +271,38 @@ SELECT COUNT(*)
...
+    }
+
+    private void stubListActivationFlags(final String activationId) throws Exception {
+        final var request = new ListActivationFlagsRequest();
+        request.setActivationId(activationId);
+
+        final var response = new ListActivationFlagsResponse();
+        response.setActivationId(activationId);
+
+        when(powerAuthClient.listActivationFlags(request, new LinkedMultiValueMap<>(), new LinkedMultiValueMap<>()))
```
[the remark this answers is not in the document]
michal-rozehnal-w: Good catch, fixed

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/ConsentResolvedGuard.java
```diff
@@ -0,0 +1,60 @@
...
+                .orElse(null);
+
+        if (process == null) {
+            logger.warn("Process not found, processId: {}", processId);
+            return false;
+        }
+
+        try {
+            final var consentPending = identityVerificationStatusService.isConsentPending(process);
+            logger.info("Consent pending: {}, processId: {}", consentPending, processId);
```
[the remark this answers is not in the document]
michal-rozehnal-w: Valid argument, changed the level to debug

enrollment-server-onboarding/src/main/resources/application.properties
```diff
@@ -32,6 +32,7 @@ spring.datasource.hikari.auto-commit=false
 spring.datasource.hikari.pool-name=HikariPool-Enrollment-Server-Onboarding
 spring.jpa.properties.hibernate.connection.characterEncoding=${ENROLLMENT_SERVER_ONBOARDING_JPA_CHARACTER_ENCODING:utf8}
 spring.jpa.properties.hibernate.connection.useUnicode=${ENROLLMENT_SERVER_ONBOARDING_JPA_USE_UNICODE:true}
+spring.jpa.properties.jakarta.persistence.validation.group.pre-persist=jakarta.validation.groups.Default,com.wultra.app.onboardingserver.common.validation.NotNullOnCreate
```
banterCZ, line 35: 👏 Nice, I like it!

enrollment-server-onboarding/src/test/resources/application-test.properties
```diff
@@ -20,6 +20,7 @@ spring.datasource.username=sa
 spring.datasource.password=password
 spring.jpa.hibernate.ddl-auto=create
 spring.jpa.properties.hibernate.validator.apply_to_ddl=false
+spring.jpa.properties.jakarta.persistence.validation.group.pre-persist=jakarta.validation.groups.Default,com.wultra.app.onboardingserver.common.validation.NotNullOnCreate
```
banterCZ: 🤔 It shouldn't be needed to override this property, should it?

banterCZ (approved):
Approving, but let's wait for the feedback from mobile development.

enrollment-server-onboarding/src/test/resources/application-test.properties
michal-rozehnal-w: Good catch, removed

michal-rozehnal-w:
> Approving, but let's wait for the feedback from mobile development.

Thank you for the review. I agree and added task for it

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/IdentityVerificationStatusResponse.java
marek-ch: on mobile we have this scheme in Onboarding SDK
```
struct IdentityStatusResponse: Decodable {
    
    private enum CodingKeys: String, CodingKey {
        case processId = "processId"
        case processType = "processType"
        case config = "config"
        case status = "identityVerificationStatus"
        case phase = "identityVerificationPhase"
        case consentRequired = "consentRequired"
    }
    
    let processId: String
    /// Configured name of onboarding process type.
    let processType: String
    let status: IdentityVerificationStatus
    let phase: IdentityVerificationPhase?
    let config: IdentityConfig
    let consentRequired: Bool?
}
```
So yea, we would like to keep the `consentRequired` name as it is, since renaming would be breaking change. I have also checked Java BE implementation and it looks the consentRequired is non-optional there.

marek-ch: @kober32 is your guy for this, since he has a bigger picture. From my perspective, the rename would mean new releases.. But I understand the motivation.

kober32 (approved):

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/IdentityVerificationStatusResponse.java
michal-rozehnal-w: OK, let’s stick with `consentRequired` in the API.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationStatusService.java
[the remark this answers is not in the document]
michal-rozehnal-w: The BE uses `consentPending`, but the FE `consentRequired` because of compatibility

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/IdentityVerificationStatusResponse.java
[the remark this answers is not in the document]
michal-rozehnal-w: I updated the description little bit

PR #1630, opened 2026-03-20
Fix #1592: Add Event to faster the NEXT_STATE_EVENT

### LINKED ISSUE #1592: Add Event to faster the NEXT_STATE_EVENT
### Task specification

Refactor state machine usage of NEXT_STATE_EVENT for a more event-like approach (instead of a scheduled job). Diagram applies TBD.

banterCZ:
- Limit `EVENT_NEXT_STATE` as much as possible and prefer an eventless approach
- Introduce more specific events
- Handle state-persisting logic in `StateMachineConfig` instead of scattered snippets


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/MoveToDocumentUploadVerificationPendingAction.java
```diff
@@ -1,51 +0,0 @@
...
-import static com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationPhase.DOCUMENT_UPLOAD;
-import static com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationStatus.VERIFICATION_PENDING;
-
-/**
- * Action to move the given identity verification to {@code DOCUMENT_UPLOAD / VERIFICATION_PENDING}.
- *
- * @author Lubos Racansky, lubos.racansky@wultra.com
- */
-@Component
-public class MoveToDocumentUploadVerificationPendingAction extends MoveActionAdapter {
```
banterCZ, line 35: We want to move the state machine in `stateEntryFunction` rather than in actions as much as possible.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/guard/document/DocumentUploadVerificationPendingGuard.java
```diff
@@ -1,43 +0,0 @@
...
- * Guard to ensure ({@link IdentityVerificationPhase#DOCUMENT_UPLOAD}, {@link IdentityVerificationStatus#VERIFICATION_PENDING}) status
- *
- * @author Lukas Lukovsky, lukas.lukovsky@wultra.com
- */
-@Component
-public class DocumentUploadVerificationPendingGuard implements Guard<OnboardingState, OnboardingEvent> {
-
-    @Override
-    public boolean evaluate(StateContext<OnboardingState, OnboardingEvent> context) {
-        IdentityVerificationEntity identityVerification = context.getExtendedState().get(ExtendedStateVariable.IDENTITY_VERIFICATION, IdentityVerificationEntity.class);
```
banterCZ, line 39: The guard does not make sense; it duplicates the state machine. Moreover, it breaks when the state is no longer moved in the action, because the guard is called before `stateEntryFunction`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -259,16 +255,9 @@ private void configureDocumentUploadTransitions(StateMachineTransitionConfigurer
...
-                .event(OnboardingEvent.EVENT_NEXT_STATE)
+                .event(OnboardingEvent.DOCUMENT_UPLOADED)
                 .guard(documentsVerificationPendingGuard)
-                .action(moveToDocumentUploadVerificationPendingAction)
-                .target(OnboardingState.CHOICE_DOCUMENT_UPLOAD)
-
-                .and()
-                .withChoice()
-                .source(OnboardingState.CHOICE_DOCUMENT_UPLOAD)
-                .first(OnboardingState.DOCUMENT_UPLOAD_VERIFICATION_PENDING, documentUploadVerificationPendingGuard)
```
banterCZ, line 293: Not much useful, it was just verifying that `moveToDocumentUploadVerificationPendingAction` has been executed.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -208,6 +203,7 @@ private Mono<Void> persistState(final StateContext<OnboardingState, OnboardingEv
         final IdentityVerificationEntity identityVerification = context.getExtendedState().get(ExtendedStateVariable.IDENTITY_VERIFICATION, IdentityVerificationEntity.class);
 
         return Mono.fromRunnable(() -> identityVerificationService.moveToPhaseAndStatus(identityVerification, state.getPhase(), state.getStatus(), ownerId))
+                .doOnError(e -> logger.error("Error during move to phase: {}, state: {}", state.getPhase(), state.getStatus(), e))
```
banterCZ: Already handled in `org.springframework.statemachine.state.ObjectState`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -259,43 +254,39 @@ private void configureDocumentUploadTransitions(StateMachineTransitionConfigurer
...
-                .event(OnboardingEvent.EVENT_NEXT_STATE)
                 .action(verificationDocumentStartAction)
-                .guard(createCompositeGuard(processIdentifierGuard, documentsVerificationPendingGuard))
+                .guard(processIdentifierGuard)
                 .target(OnboardingState.CHOICE_DOCUMENT_VERIFICATION_PROCESSING)
 
                 .and()
                 .withChoice()
                 .source(OnboardingState.CHOICE_DOCUMENT_VERIFICATION_PROCESSING)
-                .first(OnboardingState.DOCUMENT_VERIFICATION_IN_PROGRESS, statusInProgressGuard)
```
banterCZ, line 309: This looks like a change of flow, but the code was misleading because the logic in the action had been moving the flow to the different states anyway.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/interceptor/CustomStateMachineInterceptor.java
```diff
@@ -103,37 +91,4 @@ public Exception stateMachineError(StateMachine<OnboardingState, OnboardingEvent
 
         return e;
     }
-
-    @Override
-    public StateContext<OnboardingState, OnboardingEvent> postTransition(StateContext<OnboardingState, OnboardingEvent> context) {
```
banterCZ, line 102: The purpose of this interceptor was to validate transition changes made in actions against the state machine config. But we are moving the state persistence to the state machine.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/ClientEvaluationTransitionsTest.java
```diff
@@ -1,236 +0,0 @@
...
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.mockito.Mockito.*;
-
-/**
- * @author Lukas Lukovsky, lukas.lukovsky@wultra.com
- */
-@SpringBootTest(classes = EnrollmentServerTestApplication.class)
-@ActiveProfiles("test")
-@Transactional
-class ClientEvaluationTransitionsTest extends AbstractStateMachineTest {
```
banterCZ, line 62: Since we are heavily removing `EVENT_NEXT_STATE`, the process may reach a bigger step, and such small slices do not make much sense. We have to refactor the tests.

enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/enumeration/PresenceCheckStatus.java
```diff
@@ -24,11 +24,6 @@
  */
 public enum PresenceCheckStatus {
 
-    /**
-     * User presence is being checked.
-     */
-    IN_PROGRESS,
```
banterCZ, line 30: Completely removed. Never mapped at PresenceCheckProvider, and when we are leaving NEXT_EVENT, it would complicate handling.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -540,10 +505,8 @@ private void configureCompletedTransition(StateMachineTransitionConfigurer<Onboa
         transitions
                 .withChoice()
                 .source(OnboardingState.CHOICE_COMPLETED_STATE)
-                .first(OnboardingState.COMPLETED_ACCEPTED, statusAcceptedGuard)
-                .then(OnboardingState.COMPLETED_REJECTED, statusRejectedGuard)
```
banterCZ, line 566: `COMPLETED_REJECTED` has never been supported and does not make sense at all. This step is final validation only; it may just pass or fail.

enrollment-server-onboarding-provider-iproov/src/test/java/com/wultra/app/onboardingserver/provider/iproov/IProovPresenceCheckProviderTest.java
```diff
@@ -95,7 +95,7 @@ void getResultTest() throws Exception {
 
         PresenceCheckResult result = provider.getResult(ownerId, sessionInfo);
 
-        assertEquals(PresenceCheckStatus.IN_PROGRESS, result.getStatus());
+        assertEquals(PresenceCheckStatus.ACCEPTED, result.getStatus());
```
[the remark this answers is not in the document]
banterCZ: This integration test is disabled anyway, required `@ActiveProfiles("external-service")` and `IN_PROGRESS` was not parsed from the response.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/OnboardingTransitionTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Test transitions defined in {@link StateMachineConfig}.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@SpringBootTest(classes = EnrollmentServerTestApplication.class)
+@ActiveProfiles("test")
+@Slf4j
+class OnboardingTransitionTest extends AbstractStateMachineTest {
```
banterCZ, line 61: 👀 A new way to test state machine transitions.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/PresenceCheckTransitionsTest.java
```diff
@@ -115,131 +105,13 @@ void testPresenceCheckInProgress() throws Exception {
...
 
-        Message<OnboardingEvent> message =
-                stateMachineService.createMessage(OWNER_ID, idVerification.getProcessId(), OnboardingEvent.EVENT_NEXT_STATE);
-
         prepareTest(stateMachine)
-            .sendEvent(message)
             .expectState(OnboardingState.PRESENCE_CHECK_IN_PROGRESS)
             .and()
             .build()
             .test();
```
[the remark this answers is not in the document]
banterCZ: Keep it as is. We may remove this test in the future in advance of `OnboardingTransitionTest`.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/OtpTransitionsTest.java
```diff
@@ -153,13 +147,12 @@ void testOtpNotVerified() throws Exception {
...
                 .and()
                 .step()
-                .sendEvent(OnboardingEvent.EVENT_NEXT_STATE)
                 .expectState(OnboardingState.OTP_VERIFICATION_PENDING)
                 .and()
                 .build()
                 .test();
 
-        verify(verificationProcessResultAction, never()).execute(any());
+        verify(verificationResultService, never()).processVerificationResult(any(), any());
```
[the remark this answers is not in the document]
banterCZ: Keep it as is. We may remove this test in the future in advance of `OnboardingTransitionTest`.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/IdentityVerificationRepository.java
```diff
@@ -61,26 +61,8 @@ public interface IdentityVerificationRepository extends CrudRepository<IdentityV
     @Query("""
             SELECT id
                 FROM IdentityVerificationEntity id
-                WHERE ((id.phase = com.wultra.app.enrollmentserver.model.enumeration.IdentityVerificationPhase.DOCUMENT_UPLOAD
```
michal-rozehnal-w, line 64: I like that this condition has been simplified significantly.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/OnboardingApprovalAction.java
```diff
@@ -48,4 +49,45 @@ public void execute(StateContext<OnboardingState, OnboardingEvent> context) {
...
+     */
+    public static Guard<OnboardingState, OnboardingEvent> isResultInProgress() {
+        return isResult(ApproveClientResponse.ApprovalResult.WAIT);
+    }
+
+    private static Guard<OnboardingState, OnboardingEvent> isResult(ApproveClientResponse.ApprovalResult expectedResult) {
+        return context -> evaluateResult(context, expectedResult);
+    }
+
+    private static boolean evaluateResult(final StateContext<OnboardingState, OnboardingEvent> context, final ApproveClientResponse.ApprovalResult expectedResult) {
```
michal-rozehnal-w: This is definitely out of scope for this PR. We can create a common method for this status verification (if possible), since it’s repeated several times with different types.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/presencecheck/PresenceCheckVerificationAction.java
```diff
@@ -45,24 +41,48 @@
...
-            presenceCheckService.checkPresenceVerification(ownerId, identityVerification);
-        } catch (PresenceCheckException | RemoteCommunicationException e) {
-            logger.error("Checking presence verification failed, {}", ownerId, e);
-            identityVerification.setErrorDetail(IdentityVerificationEntity.PRESENCE_CHECK_FAILED);
-            identityVerification.setErrorOrigin(ErrorOrigin.PRESENCE_CHECK);
-            identityVerification.setTimestampFailed(ownerId.getTimestamp());
-            identityVerificationService.moveToPhaseAndStatus(identityVerification, identityVerification.getPhase(), FAILED, ownerId);
-        }
+        final var result = presenceCheckService.checkPresenceVerification(ownerId, identityVerification);
+        context.getExtendedState().getVariables().put(RESULT_KEY, result);
```
michal-rozehnal-w, line 54: The `put()` method doesn't allow `null`—it throws an NPE. I recommend being defensive and always using the pattern we have for example in the `ClientEvaluationService`:

```
context.getExtendedState().getVariables().put(RESULT_KEY, result != null ? result : new NullObject());
```

There are more calls of the `put` to be checked in classes:
- `DocumentVerificationFinalAction`
- `VerificationProcessResultAction`

michal-rozehnal-w (approved):
Really good work! I just added note about handling `null`. 
This is a bigger change/refactoring that should be well tested once the PR is merged.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/presencecheck/PresenceCheckVerificationAction.java
banterCZ: Thanks for pointing that out. I use `NullObject` where we signal failure via `NULL`. 🤔 Maybe this is a good code smell that we should refactor that and avoid `NULL`, but let's postpone it to another PR.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/OnboardingApprovalAction.java
banterCZ: I was thinking about it, but it was not a priority for me. Mainly, this is a _private static_ method encapsulated in a public method. The bigger issue is Java generics, which are not easy to handle since I prefer type safety for the result object.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/presencecheck/PresenceCheckVerificationAction.java
michal-rozehnal-w: I agree with having a separate issue for the refactoring, since the current solution is more of a workaround than an ideal solution.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/OnboardingApprovalAction.java
michal-rozehnal-w: This is a fair point. I think type safety is more important than redundancy, especially when it’s just a few lines of code.

enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/integration/OwnerId.java
```diff
@@ -83,4 +84,11 @@ public String getUserIdSecured() {
         return userIdSecured;
     }
 
+    private static byte[] sha256(final String source) {
+        try {
+            return Hash.sha256(source);
+        } catch (CryptoProviderException e) {
+            throw new IllegalStateException("sha256 is not supported", e);
```
banterCZ: ℹ️ unrelated change, should be fixed in the _develop_ branch, but needs to be here to make it compile

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedListenerIntTest.java
```diff
@@ -81,16 +80,15 @@ class OnboardingCompletedAcceptedListenerIntTest {
...
-
     @MockitoBean
     private UserDataStoreService userDataStoreService;
 
     @Captor
     private ArgumentCaptor<OnboardingCompletedAcceptedEvent> onboardingCompletedAcceptedEventCaptor;
 
     @Test
+    // TODO (racansky, 2026-05-04 need to be fixed after refactoring of state machine service and guards
+    @Disabled("need to be fixed after refactoring of state machine service and guards")
```
banterCZ: I need to fix this test with @michal-rozehnal-w

banterCZ: For `COMPLETED_ACCEPTED`, we have to combine the persist and publish entry functions into a single entry function. Spring State Machine only keeps the last function registered via `stateEntryFunction()` for a given state, so registering them separately would cause one of them to be silently dropped.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -212,17 +204,30 @@ private Function<StateContext<OnboardingState, OnboardingEvent>, Mono<Void>> per
...
-        return Mono.fromRunnable(() -> identityVerificationService.moveToPhaseAndStatus(identityVerification, state.getPhase(), state.getStatus(), ownerId))
-                .subscribeOn(Schedulers.boundedElastic())
-                .then();
+        logger.debug("action: persistState, state: initiated, target: {}", state);
+        return Mono.fromRunnable(() -> identityVerificationService.moveToPhaseAndStatus(identityVerification, state.getPhase(), state.getStatus(), ownerId));
     }
 
     private void registerPublishEventFunction(final StateConfigurer<OnboardingState, OnboardingEvent> configurer) {
-        configurer.stateEntryFunction(OnboardingState.COMPLETED_ACCEPTED, publishCompletedAcceptedEvent());
+        // For COMPLETED_ACCEPTED we have to combine the persist and publish entry functions into a single
```
banterCZ, line 223: ℹ️ This was quite tricky.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentVerificationService.java
```diff
@@ -22,117 +22,120 @@
...
 
-        final DocumentsVerificationResult result = documentVerificationProvider.verifyDocuments(ownerId, uploadIds);
-        final String verificationId = result.getVerificationId();
-        final DocumentVerificationStatus status = result.getStatus();
+        final DocumentsVerificationResult documentsVerificationResult;
+        try {
+            documentsVerificationResult = documentVerificationProvider.verifyDocuments(ownerId, uploadIds);
+        } catch (RemoteCommunicationException | DocumentVerificationException e) {
+            logger.error("action: executeFinalDocumentVerification, state: failed, exceptionMessage: {}", e.getMessage(), e);
+            return FinalDocumentVerificationResult.FAILED;
```
zcgandcomp: Why are we not calling fail(...) We are losing information stored in:

```
        identityVerification.setErrorDetail(ErrorDetail.DOCUMENT_VERIFICATION_FAILED);
        identityVerification.setErrorOrigin(ErrorOrigin.DOCUMENT_VERIFICATION);
        identityVerification.setTimestampFailed(ownerId.getTimestamp());
```
and the score increment.

But true this want process previously as well so no difference from previous behavior.

enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/enumeration/PresenceCheckStatus.java
zcgandcomp, line 30: Just in case it's not referenced, we should remove it.
But isn't it possible to have some data in the database? I would like to prevent some failures based on old data.

banterCZ: It is just a model object for the provider, not stored in the database.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentVerificationService.java
banterCZ: Should I update it or create a separate issue for that?

zcgandcomp: I would update it. Not must, but it be more consistent

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -177,15 +177,15 @@ public void submitDocuments(final DocumentSubmitV2Request request, final OwnerId
...
      *
      * @param ownerId Owner identification.
      * @param identityVerification Identity verification.
-     * @throws RemoteCommunicationException In case of remote communication error.
-     * @throws DocumentVerificationException In case of business logic error.
+     * @return Document evaluation status or {@code null} if the verification process failed at document verification provider.
      */
     @Transactional
-    public void startVerification(OwnerId ownerId, IdentityVerificationEntity identityVerification) throws DocumentVerificationException, RemoteCommunicationException {
+    public @Nullable DocumentEvaluationStatus startDocumentVerification(OwnerId ownerId, IdentityVerificationEntity identityVerification) {
```
zcgandcomp: My initial thought: Why are we returning Evaluation status instead of `DocumentVerificationStatus`?
Second thought: The DocumentEvaluationStatus is not a single document status, but a state where we gather all required documents. In this case I still have 2 concerns with naming:
- This will get confused with EvaluateClientResponse.EvaluationResult 
- OK and NOK - if we are going to use enum, let's name it self-explanatory e.g. `SUFFICIENT_VERIFIED_DOCUMENTS`, `INSUFFICIENT_VERIFIED_DOCUMENTS`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -206,14 +206,21 @@ public void startVerification(OwnerId ownerId, IdentityVerificationEntity identi
                 .map(DocumentVerificationEntity::getUploadId)
                 .toList();
 
-        final DocumentsVerificationResult result = documentVerificationProvider.verifyDocuments(ownerId, uploadIds);
+        final DocumentsVerificationResult result;
+        try {
+            result = documentVerificationProvider.verifyDocuments(ownerId, uploadIds);
+        } catch (RemoteCommunicationException | DocumentVerificationException e) {
+            logger.warn("action: startDocumentVerification, state: failed, exceptionMessage: {}", e.getMessage(), e);
+            return null;
```
zcgandcomp: Is there a particular reason we are not returning FAIL?

Or better - if this fails. We failed to start the verification and should go to FAIL. 
We return null. The guards  will work ok and move the state to fail, but there is a strange null relaying logic.

zcgandcomp: And In a way when document got rejected, we are keeping the loop and requestion new upload. But if this exeption occurs we break the cycle and go to FAIL - effectively throwing the whole IdentityVerification right? So I would suggest here to contine and take the document as FAILED.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -319,17 +324,17 @@ private void moveToDocumentVerificationAndStatusByDocuments(
                 .allMatch(it -> it == DocumentStatus.ACCEPTED)) {
             // The timestampFinished parameter is not set yet, there may be other steps ahead
             if (allRequiredDocumentsChecked) {
-                // Move to DOCUMENT_VERIFICATION / ACCEPTED only in case all documents were checked
-                moveToPhaseAndStatus(idVerification, IdentityVerificationPhase.DOCUMENT_VERIFICATION, ACCEPTED, ownerId);
+                logger.debug("All required documents are accepted");
```
zcgandcomp, line 327: Question:
If user upload one document which fails. And then starts uploading different documents. All other documents pass and there is enough `ACCEPTED` documents, we will still got `NOK`. And we will go to the handleDocumentStatus, which checks all uploaded document statuses - i.e., the process can only fail.

There might be a mechanism to discard FAILED / REJECTED documents to prevent such situation, but even so I recommend constructing the logic as such:

If the last document verification PASS and allRequiredDocumentsChecked is true return `OK` (or better `SUFFICIENT_VERIFIED_DOCUMENTS`)
If the last document PASS and allRequiredDocumentsChecked is false return `NOK` (or better `INSUFFICIENT_VERIFIED_DOCUMENTS`)
If the last document `FAILED` or got `REJECTED`, handle the status (increase the error score etc,..) and return `NOK`

This way you will not rely on document discarding mechanism.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -571,4 +571,21 @@ private void deleteDocumentData(final OwnerId ownerId, final List<DocumentVerifi
         auditService.auditDocumentVerificationProvider(identityVerification, "Cleaned up documents for user: {}", ownerId.getUserId());
         logger.info("All document data successfully deleted");
     }
+
+    public enum DocumentEvaluationStatus {
+        /**
+         * All documents are accepted.
+         */
+        OK,
```
zcgandcomp: see my comments to naming

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
```diff
@@ -104,12 +104,26 @@ public SessionInfo init(OwnerId ownerId, String processId)
...
     @Transactional
-    public void checkPresenceVerification(
+    public PresenceCheckStatus checkPresenceVerification(final OwnerId ownerId, final IdentityVerificationEntity identityVerification) {
+        logger.info("action: checkPresenceVerification, state: initiated, identityVerificationId: {}", identityVerification.getId());
+
+        try {
+            final var result = checkPresenceVerificationInternal(ownerId, identityVerification);
+            logger.info("action: checkPresenceVerification, state: succeeded, identityVerificationId: {}, result: {}", identityVerification.getId(), result);
+            return result;
+        } catch (final PresenceCheckException | RemoteCommunicationException e) {
```
zcgandcomp, line 117: I can see the pattern PresenceCheckException | RemoteCommunicationException is repeating. We should make some distinction if the error is caused by communication error or "something else"
but this could be done separately.

zcgandcomp, line 117: Do we need to do the catch block here? Because it would be IMHO better to do the error processing in the
`evaluatePresenceCheckResult` - we are evaluating `FAILED` result there anyway.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/presencecheck/PresenceCheckVerificationAction.java
```diff
@@ -45,24 +41,48 @@
...
-        try {
-            presenceCheckService.checkPresenceVerification(ownerId, identityVerification);
-        } catch (PresenceCheckException | RemoteCommunicationException e) {
-            logger.error("Checking presence verification failed, {}", ownerId, e);
-            identityVerification.setErrorDetail(IdentityVerificationEntity.PRESENCE_CHECK_FAILED);
-            identityVerification.setErrorOrigin(ErrorOrigin.PRESENCE_CHECK);
-            identityVerification.setTimestampFailed(ownerId.getTimestamp());
-            identityVerificationService.moveToPhaseAndStatus(identityVerification, identityVerification.getPhase(), FAILED, ownerId);
-        }
+        final var result = presenceCheckService.checkPresenceVerification(ownerId, identityVerification);
```
zcgandcomp, line 53: As one of my previous comments - this looks much cleaner, the catch block is in the underlying logic.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/DocumentsVerificationPendingGuard.java
```diff
@@ -16,55 +16,37 @@
...
-            return false;
-        }
-
-        final boolean pendingVerificationDocumentPresent = documentVerifications.stream()
-                .map(DocumentVerificationEntity::getStatus)
-                .anyMatch(it -> it == DocumentStatus.VERIFICATION_PENDING);
+        final boolean pendingVerificationDocumentPresent = documentVerificationService.hasDocumentsVerificationPending(identityVerification);
 
         if (pendingVerificationDocumentPresent) {
             logger.info("Pending verification document present for {}, {}", identityVerification, ownerId);
```
zcgandcomp: If I understood no we do not call this periocally so we can put both logger output to info. Or create one with key field representing pendingVerificationDocumentPresent

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/VerificationDocumentStartAction.java
```diff
@@ -18,52 +18,71 @@
...
 
-        try {
-            identityVerificationService.startVerification(ownerId, identityVerification);
-            logger.info("Started document verification process of {}", identityVerification);
-        } catch (DocumentVerificationException | RemoteCommunicationException e) {
-            identityVerificationService.moveToPhaseAndStatus(identityVerification, DOCUMENT_VERIFICATION, FAILED, ownerId);
-            logger.warn("Verification start failed, {}", ownerId, e);
-        }
+        final var result = identityVerificationService.startDocumentVerification(ownerId, identityVerification);
+        context.getExtendedState().getVariables().put(RESULT_KEY, result != null ? result : new NullObject());
```
zcgandcomp: I would prefere if the null was not used to evaluate status logic.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
```diff
@@ -482,7 +483,7 @@ public ResponseEntity<Response> resendOtp(
         logger.debug("Onboarding process will be locked using PESSIMISTIC_WRITE lock, {}", processId);
         onboardingService.verifyProcessIdAndLock(ownerId, processId, OnboardingStatus.VERIFICATION_IN_PROGRESS);
 
-        StateMachine<OnboardingState, OnboardingEvent> stateMachine = stateMachineService.processStateMachineEvent(ownerId, processId, OnboardingEvent.OTP_VERIFICATION_RESEND);
+        StateMachine<OnboardingState, OnboardingEvent> stateMachine = stateMachineService.processStateMachineEvent(ownerId, processId, OnboardingEvent.OTP_RESEND);
```
zcgandcomp, line 485: Let's solve it in different issue, but why do we have resend as a special step. It doesn't change the state of the state machine at all. And just sending OTP via StateMachine doesn't IMHO bring any additional value, but increase the SM complexity...

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -282,58 +287,50 @@ private void configureDocumentUploadTransitions(StateMachineTransitionConfigurer
         transitions
                 .withExternal()
                 .source(OnboardingState.DOCUMENT_UPLOAD_IN_PROGRESS)
-                .event(OnboardingEvent.EVENT_NEXT_STATE)
+                .event(OnboardingEvent.DOCUMENT_UPLOADED)
```
zcgandcomp, line 290: This could be very fragile IMHO.
What is the purpose of this guard?
If there are some documents in document status `VERIFICATION_PENDING,` this will move the state machine to a target state. BUT if not ( e.g., `UPLOAD_IN_PROGRESS`, `ACCEPTED`, `VERIFICATION_IN_PROGRESS`) The state machine is stuck in DOCUMENT_UPLOAD_IN_PROGRESS. Basically forever (or unless somebody uploads a new document)

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
banterCZ: It is related to our discussion with @kober32 about `resubmit` and `originalDocumentId`, isn't? I would rather solve it in a different issue. This PR is already so huge.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestService.java
banterCZ: Resend leads to `OtpVerificationResendAction` -> `IdentityVerificationOtpService#resendOtp`... –> `OtpServiceImp#createOtpCodeForResend` which verifies resend period and invalidates the previous one.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
banterCZ: As it behaves today. If you need to upload two documents, you won't be able to move on until you've uploaded the second one (or until the process times out). There's no point in hitting it every minute when the controller can trigger it.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/VerificationDocumentStartAction.java
banterCZ: @michal-rozehnal-w do you recall the reason? I guess we kept the `null` just from a legacy point of view, not to change too much code, where the service was returning `null`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -282,58 +287,50 @@ private void configureDocumentUploadTransitions(StateMachineTransitionConfigurer
...
-                .then(OnboardingState.DOCUMENT_VERIFICATION_ACCEPTED, statusAcceptedGuard)
-                .then(OnboardingState.DOCUMENT_VERIFICATION_REJECTED, statusRejectedGuard)
-                .then(OnboardingState.DOCUMENT_VERIFICATION_FAILED, statusFailedGuard)
-                .last(OnboardingState.UNEXPECTED_STATE)
+                .first(OnboardingState.DOCUMENT_VERIFICATION_ACCEPTED, VerificationDocumentStartAction.isResultOk())
+                .then(OnboardingState.DOCUMENT_UPLOAD_IN_PROGRESS, VerificationDocumentStartAction.isResultInProgress())
+                .last(OnboardingState.DOCUMENT_VERIFICATION_FAILED)
 
                 .and()
                 .withExternal()
```
zcgandcomp, line 311: if there is no event defined, shouldn't we use withChoice instead of withExternal?
Or if the guard is false, we get stuck in the state indefinitely.... The only option is to wait for the process expiration.

banterCZ: We would write
```java
.withChoice()
.source(OnboardingState.CHOICE_DOCUMENT_VERIFICATION_ACCEPTED)
.last(OnboardingState.DOCUMENT_VERIFICATION_FINAL_IN_PROGRESS)
```
but `CHOICE_*` is a pseudostate not stored in the database. With the current code, we have a track record that we went through states `DOCUMENT_VERIFICATION_ACCEPTED` and `DOCUMENT_VERIFICATION_FINAL_IN_PROGRESS`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
banterCZ: I have opened an issue to improve the _resubmit_ logic https://github.com/wultra/enrollment-server/issues/1783

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
banterCZ: I made that _null-safe_.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/VerificationDocumentStartAction.java
banterCZ: I made that _null-safe_.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentVerificationService.java
banterCZ: I am calling `fail()` now.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/PresenceCheckService.java
banterCZ: Here, the `PresenceCheckException` is thrown in case of parsing errors, etc. It is not a business error. I don't think it's worth the trouble to separate it.

banterCZ: `evaluatePresenceCheckResult` requires the whole result object; I extracted at least a common logic into a method `processFailedPresenceCheck`.

zcgandcomp (approved):
Ok functional-wise seems fine. Optimization should be splitted to separate issues. And the whole logic passed the test.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -570,4 +570,24 @@ private void deleteDocumentData(final OwnerId ownerId, final List<DocumentVerifi
...
+
+        /**
+         * All documents are accepted.
+         */
+        ALL_DOCUMENTS_ACCEPTED,
+
+        /**
+         * Some documents are not accepted or not all required documents are accepted yet.
+         */
+        INSUFFICIENT_DOCUMENTS,
```
zcgandcomp: INSUFFICIENT_DOCUMENT_COUNT

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -570,4 +570,24 @@ private void deleteDocumentData(final OwnerId ownerId, final List<DocumentVerifi
         auditService.auditDocumentVerificationProvider(identityVerification, "Cleaned up documents for user: {}", ownerId.getUserId());
         logger.info("All document data successfully deleted");
     }
+
+    public enum VerificationDocumentActionResult {
+
+        /**
+         * All documents are accepted.
+         */
+        ALL_DOCUMENTS_ACCEPTED,
```
zcgandcomp: REQUIRED_DOCUMENTS_VERIFIED

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationService.java
```diff
@@ -220,4 +227,21 @@ private void processEvaluationResponse(final IdentityVerificationEntity identity
             logger.info("Client evaluation waiting for identity verification id: {}", identityVerificationId);
         }
     }
+
+    public enum ClientEvaluationResult {
```
michal-rozehnal-w, line 231: Nitpick: We can have javadoc for all options

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/OnboardingApprovalAction.java
```diff
@@ -37,15 +36,56 @@
...
     private final OnboardingApprovalService onboardingApprovalService;
 
     @Override
     public void execute(StateContext<OnboardingState, OnboardingEvent> context) {
         final IdentityVerificationEntity identityVerification = context.getExtendedState().get(ExtendedStateVariable.IDENTITY_VERIFICATION, IdentityVerificationEntity.class);
 
-        final ApproveClientResponse.ApprovalResult result = onboardingApprovalService.approve(identityVerification);
-        context.getExtendedState().getVariables().put(RESULT_KEY, result != null ? result : new NullObject());
+        final OnboardingApprovalService.ApprovalResult result = onboardingApprovalService.approve(identityVerification);
+        context.getExtendedState().getVariables().put(RESULT_KEY, result);
```
michal-rozehnal-w, line 48: Just to be sure, the `result` is no longer nullable right?

michal-rozehnal-w (approved):
I checked the latest changes, and they look good to me. However, the PR is quite large and has a significant impact on the flow, so we need to test it thoroughly.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/action/verification/OnboardingApprovalAction.java
banterCZ: Exactly

PR #1643, opened 2026-03-24
Resolves #1642: Fix client evaluation test stubbing

### LINKED ISSUE #1642: Fix flaky strict stubbing in ClientEvaluationServiceTest
## Problem

The `Deploy with Maven` workflow on `develop` was failing in `enrollment-server-onboarding` because `ClientEvaluationServiceTest.testProcessClientEvaluation_tooManyAttempts` used a strict Mockito stub for `onboardingProvider.evaluateClient(...)`.

The production code now builds an `EvaluateClientRequest` whose `documentCheckResult` no longer matches the exact object expected by the test in this scenario, so Mockito reports a `PotentialStubbingProblem` even though the behavior under test is retry exhaustion.

## Additional CI error found during validation

After fixing the strict stubbing problem, the PR workflow still exposed another failure in `IdentityVerificationRestServiceIntTest.testCleanup_fullDocumentVerificationAndPresenceCheckCleanup_allDataAreCleaned`.

The test was racing with active scheduled cleanup jobs in the Spring test context. In CI, a background scheduler thread (`scheduling-3`) ran cleanup logic while the test was executing, which resulted in `ObjectOptimisticLockingFailureException` / `StaleObjectStateException` for `DocumentDataEntity` during auto-flush in `resetIdentityVerification(...)`.

## Reproduction

```bash
mvn -q -pl enrollment-server-onboarding -am -DskipTests=false -Dmaven.javadoc.skip=true test -Dtest=ClientEvaluationServiceTest#testProcessClientEvaluation_tooManyAttempts -Dsurefire.failIfNoSpecifiedTests=false
```

The second issue was observed in the PR workflow run after the first fix, where the affected cleanup integration test executed with scheduling enabled.

## Expected fix

- Make the client evaluation test verify retry/too-many-attempts behavior without depending on an exact `EvaluateClientRequest` payload that is not relevant to the test case.
- Keep background schedulers disabled for the cleanup integration test so the test environment is deterministic and does not race with scheduled jobs.

vita-kotacka:
## Summary
- relax the strict Mockito stub in `ClientEvaluationServiceTest.testProcessClientEvaluation_tooManyAttempts`
- keep the test focused on retry exhaustion instead of exact request payload shape
- fix the `Deploy with Maven` failure reproduced on `develop`

## Testing
- `mvn -q -pl enrollment-server-onboarding -am -DskipTests=false -Dmaven.javadoc.skip=true test -Dtest=ClientEvaluationServiceTest#testProcessClientEvaluation_tooManyAttempts -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn -q -pl enrollment-server-onboarding -am -DskipTests=false -Dmaven.javadoc.skip=true test -Dtest=com.wultra.app.onboardingserver.impl.service.ClientEvaluationServiceTest -Dsurefire.failIfNoSpecifiedTests=false`

### DIFF
--- a/enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationServiceTest.java
+++ b/enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationServiceTest.java
@@ -169,9 +169,7 @@ class ClientEvaluationServiceTest {
 
         when(identityVerificationConfig.getDocumentVerificationProvider()).thenReturn("testProvider");
 
-        final EvaluateClientRequest evaluateClientRequest = buildRequestWithoutExtractedData();
-
-        when(onboardingProvider.evaluateClient(evaluateClientRequest))
+        when(onboardingProvider.evaluateClient(any(EvaluateClientRequest.class)))
                 .thenThrow(new OnboardingProviderException());
 
         final OnboardingProcessConfigurationEntity processConfiguration = new OnboardingProcessConfigurationEntity();
--- a/enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestServiceIntTest.java
+++ b/enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestServiceIntTest.java
@@ -63,6 +63,7 @@ import static org.mockito.Mockito.when;
 @ActiveProfiles("test")
 @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_CLASS)
 @TestPropertySource(properties = {
+        "spring.task.scheduling.enabled=false",
         "enrollment-server-onboarding.document-verification.cleanupEnabled=true",
         "enrollment-server-onboarding.presence-check.cleanupEnabled=true"
 })


banterCZ (approved):
Nice improvement!

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationRestServiceIntTest.java
```diff
@@ -63,6 +63,7 @@
 @ActiveProfiles("test")
 @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_CLASS)
 @TestPropertySource(properties = {
+        "spring.task.scheduling.enabled=false",
```
michal-rozehnal-w, line 66: A good idea is to disable it.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/impl/service/ClientEvaluationServiceTest.java
```diff
@@ -169,9 +169,7 @@ void testProcessClientEvaluation_tooManyAttempts() throws Exception {
 
         when(identityVerificationConfig.getDocumentVerificationProvider()).thenReturn("testProvider");
 
-        final EvaluateClientRequest evaluateClientRequest = buildRequestWithoutExtractedData();
-
-        when(onboardingProvider.evaluateClient(evaluateClientRequest))
+        when(onboardingProvider.evaluateClient(any(EvaluateClientRequest.class)))
```
michal-rozehnal-w, line 172: This is perhaps a more general question: what do you see as the pros and cons of exact matching?
In the `processClientEvaluation` method, the `EvaluateClientRequest` is built, and it makes sense to verify that the request contains all the values we need. We have had situations in the past where we missed such a test and the request was incomplete.
On the other hand, the check in `when(...)` is not explicit, and this class contains unit tests rather than integration tests. I am interested in hearing other opinions.
However, I would keep some test for request building so we can be sure all expected values are set—unless we already have that covered somewhere else.

vita-kotacka: I agree exact matching is useful **when request construction is what the test is supposed to verify**.

Here, `testProcessClientEvaluation_tooManyAttempts` is really about **retry/error behavior**. The failure happened because the request shape changed (`documentCheckResult`), while the retry logic itself was still correct. So the strict `when(...)` made this test brittle by coupling it to a payload detail outside its main purpose.

I’d say request-building is better covered in more focused tests:
- `ClientEvaluationDocumentCheckResultFactoryTest`
- `RestOnboardingProviderTest` for request body shape
- plus the successful evaluation path can be used for service-level payload assertions if we want

michal-rozehnal-w: Okay, that is a valid reason. Thank you for the explanation. I agree with replacing the strict check

michal-rozehnal-w (approved):

PR #1653, opened 2026-03-26
Fix #1652: Remove JSON escape characters in audit

### LINKED ISSUE #1652: Remove escaping for Microblink response audit
### Task specification

When the original response from Microblink is stored into audit table then the response JSON is escaped because it is passed as `String`. We would like to have it without escape characters

michal-rozehnal-w:


enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/service/AuditService.java
```diff
@@ -230,7 +231,7 @@ public void auditOnboardingProvider(final IdentityVerificationEntity identityVer
      * @param message message, arguments may be put to via template {@code {}}
      * @param args message arguments
      */
-    public void auditDocumentVerificationProvider(final OwnerId ownerId, final String documentResponseJson, final String message, final Object... args) {
+    public void auditDocumentVerificationProvider(final OwnerId ownerId, final JsonNode documentResponseJson, final String message, final Object... args) {
```
michal-rozehnal-w, line 234: ~~This is a quick fix for the escape characters. However, I would like to refactor it a little tomorrow (that's why this PR is still in draft), unless we need to merge it ASAP.~~

Example how it looks now in the `audit_log.param` db column: [param_without_escapes.json](https://github.com/user-attachments/files/26279312/param_without_escapes.json)

zcgandcomp: From my POV, I cannot read the code. Important is the result. Storing unmodified JSON to audit is good.
The same goes to sending the unmodified JSON in evaluation endpoints.

I cannot check the code modification itself. I noticed we have additional (and suboptimal) mappers in the system due to a specific parsing of the response.

Keeping the structure as pure Object is fine from my POV - but again hard to tell if this isn't problem from a java point of view. Javascript can handle so :)

michal-rozehnal-w: Thank you very much for the feedback. I added you as a reviewer at this stage just to share the final structure in the audit log (in the attached file). The task here is trivial—it’s just about removing the escape characters—but it’s still better to have a second pair of eyes. If this looks good from your perspective, I will ask other devs for the review with focus on the code itself.

enrollment-server-onboarding-provider-microblink/src/test/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProviderIntTest.java
```diff
@@ -503,7 +505,7 @@ private void assertIdCardPassSubmitResult(final DocumentsSubmitResult result) {
         assertDoesNotThrow(() -> UUID.fromString(frontDocument.getUploadId()));
         assertNull(frontDocument.getRejectReason());
         assertEquals(frontNormalizedExtractedData, frontDocument.getExtractedData());
-        assertEquals(idCardPassValidationResult, frontDocument.getValidationResult());
+        assertJsonEquals(idCardPassValidationResult, frontDocument.getValidationResult());
```
michal-rozehnal-w, line 508: Order of properties is not guaranteed because of using `Map` in `DocumentVerificationResponseBundle::copyWithoutFields(final String... excludedFieldNames)`, so simple `String` comparison does't work.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -352,9 +359,12 @@ private DocumentVerificationParsedResponse sendApiRequest(
...
     }
 
-    private DocumentVerificationParsedResponse parseMicroblinkResponse(final String responseBodyJson) throws DocumentVerificationException {
+    private DocumentVerificationResponseBundle parseMicroblinkResponse(final String responseBodyJson) throws DocumentVerificationException {
         try {
-            return responseParser.parseResponse(responseBodyJson);
+            final var parsedResponseBody = objectMapper.readValue(responseBodyJson, DocumentVerificationResponse.class);
+            final var responseJson = objectMapper.readTree(responseBodyJson);
+
+            return new DocumentVerificationResponseBundle(parsedResponseBody, (ObjectNode) responseJson);
```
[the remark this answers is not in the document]
michal-rozehnal-w: I already checked in past that injected `ObjectMapper` has `FAIL_ON_UNKNOWN_PROPERTIES` disabled. See this [comment](https://github.com/wultra/enrollment-server/pull/1606#discussion_r2916562161)

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -352,9 +359,12 @@ private DocumentVerificationParsedResponse sendApiRequest(
         }
     }
 
-    private DocumentVerificationParsedResponse parseMicroblinkResponse(final String responseBodyJson) throws DocumentVerificationException {
+    private DocumentVerificationResponseBundle parseMicroblinkResponse(final String responseBodyJson) throws DocumentVerificationException {
         try {
-            return responseParser.parseResponse(responseBodyJson);
+            final var parsedResponseBody = objectMapper.readValue(responseBodyJson, DocumentVerificationResponse.class);
+            final var responseJson = objectMapper.readTree(responseBodyJson);
+
```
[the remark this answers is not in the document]
michal-rozehnal-w: There is already a [check](https://github.com/wultra/enrollment-server/blob/develop/enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java#L324) for a `null` body, and I see ObjectNode as part of the API contract. If anything else is returned, it means the API has changed.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkConfig.java
```diff
@@ -80,7 +70,7 @@ public MicroblinkDocumentVerificationProvider microblinkDocumentVerificationProv
             AuditService auditService) {
         return new MicroblinkDocumentVerificationProvider(
                 restClient,
-                responseParser,
+                objectMapper,
                 properties,
```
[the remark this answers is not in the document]
michal-rozehnal-w: As I mentioned above, the property is disable

enrollment-server-onboarding-provider-microblink/src/test/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProviderTest.java
```diff
@@ -1302,4 +1303,9 @@ private static DocumentVerificationProcessingOptions buildRequestOptions() {
         options.setReturnFullDocumentImage(true);
         return options;
     }
+
+    @SneakyThrows
```
banterCZ: Do we really need it? I do not see any lambda usage.

banterCZ (approved):

enrollment-server-onboarding-provider-microblink/src/test/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProviderTest.java
michal-rozehnal-w: The annotation is used because the `assertEquals` method throws a checked `JSONException`, and I didn't want to add it to every method signature where the method is used. I can add `throws Exception` everywhere, but the annotation seems cleaner to me. However, if you prefer the first approach, let me know and I’ll update it.

banterCZ: It is not crutial but I do not mind `throws Exception` in every test. Or you may use `try/catch` block in the method and use `fail()` in the `catch` branch. It is really a minor issue to discuss.

michal-rozehnal-w: This is the solution I like the most—I used the `fail` method.

PR #1669, opened 2026-04-02
Fix #1650: Add country to the submit document endpoint

### LINKED ISSUE #1650: Add country to the submit document endpoint
### Task specification

BE creates a new document entity every time `/api/v2/identity/document/submit `is called with a document that does not include `resubmit=true`. So when the endpoint is called twice—each time with a document of the same type and side—the BE treats them as unrelated entities. To update an existing document, the request must include `resubmit=true`.

This causes inconsistencies in a specific flow between the FE and BE. For example:

1. The FE uploads a document of type `ID_CARD` with only the `FRONT` side. The BE creates an entity and internally marks the document as failed because the BACK side is also required.
2. The FE uploads the same document type again, this time with both `FRONT` and `BACK` sides. The BE creates a new entity (unrelated to the one from the previous step) and marks it as passed.
3. Because the process is configured for one `ID_CARD`, the BE proceeds to the next steps in the onboarding process. However, the FE gets stuck because the `/api/identity/document/status` endpoint returns both documents, and the first one has an error. This blocks the FE, which requires all documents to be passed in order to proceed.

Possible solution:
The BE cannot reliably group documents by type and side because the current configuration allows multiple documents with the same type but different countries, and the BE does not have all relevant information during upload. It relies on the FE to use `resubmit=true` when re-uploading.

We could add the country to the request body of the upload endpoint and ensure that any new condition is reflected there. This would allow the BE to group documents more reliably.

Follow-up question:
Would this behavior deprecate the resubmit field?

banterCZ:
- [x] Add `country` field to `DocumentSubmitV2Request` with bean validation (`@Size(min=3, max=3)`)
- [x] Add `country` field to `DocumentVerificationEntity` with JavaDoc
- [x] Add `country` field to `DocumentMetadataResponseDto` with JavaDoc
- [x] Persist `country` in service layer
- [x] Add DB migration changeset for `country` column
- [x] Update migration docs

### DIFF
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/db.changelog-version.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/db.changelog-version.xml
@@ -10,6 +10,7 @@
     <include file="20260205-audit.xml" relativeToChangelogFile="true" />
     <include file="20260309-processed-document-data-verification-id.xml" relativeToChangelogFile="true" />
     <include file="20260313-consent-accepted.xml" relativeToChangelogFile="true" />
+    <include file="20260402-document-country.xml" relativeToChangelogFile="true" />
     <include file="20251215-add-tag-2.0.0.xml" relativeToChangelogFile="true" />
 
 </databaseChangeLog>
--- a/docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md
+++ b/docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md
@@ -59,6 +59,11 @@ The following changes were made to the onboarding start endpoint:
 Endpoint `POST api/identity/document/upload` was removed because it was never used in production.
 
 
+## Document Submit
+
+The endpoint `/api/identity/document/submit` has been deprecated in favor of `/api/v2/identity/document/submit`.
+
+
 ## External Onboarding Services Changes
 
 
--- a/docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md
+++ b/docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md
@@ -38,6 +38,11 @@ linked to `es_document_data` records via `upload_id` column.
 A new table `es_selfie` has been added to temporarily store selfie images of identity verification.
 
 
+### Document Verification
+
+A new column `country` has been added to the table `es_document_verification`.
+
+
 ## REST API Changes
 
 
--- a/docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md
+++ b/docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md
@@ -59,7 +59,7 @@ The following changes were made to the onboarding start endpoint:
 Endpoint `POST api/identity/document/upload` was removed because it was never used in production.
 
 
-## Document Submit
+### Document Submit
 
 The endpoint `/api/identity/document/submit` has been deprecated in favor of `/api/v2/identity/document/submit`.
 
--- a/enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/request/DocumentSubmitV2Request.java
+++ b/enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/request/DocumentSubmitV2Request.java
@@ -68,6 +68,9 @@ public record DocumentSubmitV2Request(
             @NotEmpty
             CardSide side,
 
+            @Schema(description = "Document country as an ISO 3166-1 alpha-3 code", example = "CZE", maxLength = 3, minLength = 3)
+            String country,
+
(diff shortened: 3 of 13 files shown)


docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/db.changelog-version.xml
```diff
@@ -10,6 +10,7 @@
     <include file="20260205-audit.xml" relativeToChangelogFile="true" />
     <include file="20260309-processed-document-data-verification-id.xml" relativeToChangelogFile="true" />
     <include file="20260313-consent-accepted.xml" relativeToChangelogFile="true" />
+    <include file="20260402-document-country.xml" relativeToChangelogFile="true" />
     <include file="20251215-add-tag-2.0.0.xml" relativeToChangelogFile="true" />
```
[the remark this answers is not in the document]
banterCZ: Let's keep it. This is a little bit tricky, because the Onboarding 2.0.0 has not been released yet.

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.0.0.md
```diff
@@ -38,6 +38,11 @@ linked to `es_document_data` records via `upload_id` column.
 A new table `es_selfie` has been added to temporarily store selfie images of identity verification.
 
 
+### Document Verification
+
+A new column `country` has been added to the table `es_document_verification`.
+
+
 ## REST API Changes
```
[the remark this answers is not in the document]
banterCZ: V2 has been introduced in 2.0, but you are right, it is completely missing.

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/response/data/DocumentMetadataResponseDto.java
```diff
@@ -62,4 +64,11 @@ public class DocumentMetadataResponseDto {
      */
     private List<String> errors;
 
+    /**
+     * Document country as an ISO 3166-1 alpha-3 code.
+     */
+    @Schema(description = "Document country as an ISO 3166-1 alpha-3 code", example = "CZE", maxLength = 3, minLength = 3)
+    @Size(min = 3, max = 3)
+    private String country;
```
banterCZ, line 72: ℹ️ @kober32

enrollment-server-onboarding-api-model/src/main/java/com/wultra/app/enrollmentserver/api/model/onboarding/request/DocumentSubmitV2Request.java
```diff
@@ -68,6 +68,9 @@ public record Document(
             @NotEmpty
             CardSide side,
 
+            @Schema(description = "Document country as an ISO 3166-1 alpha-3 code", example = "CZE", maxLength = 3, minLength = 3)
+            String country,
```
banterCZ, line 72: ℹ️ @kober32

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentVerificationEntity.java
```diff
@@ -88,6 +88,12 @@ public class DocumentVerificationEntity implements Serializable {
     @Column(name = "other_side_id")
     private String otherSideId;
 
+    /**
+     * Country stored as an ISO 3166-1 alpha-3 code, e.g. CZE.
```
michal-rozehnal-w: Nitpick, we can use `{@code CZE}`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -527,7 +527,7 @@ private void moveToDocumentUpload(final OwnerId ownerId, final IdentityVerificat
      * @return Document metadata for response
      */
     private DocumentMetadataResponseDto toDocumentMetadata(DocumentVerificationEntity entity) {
-        DocumentMetadataResponseDto docMetadata = new DocumentMetadataResponseDto();
+        final DocumentMetadataResponseDto docMetadata = new DocumentMetadataResponseDto();
```
michal-rozehnal-w: Nitpick, we can use `var` instead of full type

michal-rozehnal-w (approved):
Looks good from BE perspective, let's wait for FE feedback

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentVerificationEntity.java
banterCZ: Sure

```suggestion
     * Country stored as an ISO 3166-1 alpha-3 code, e.g. {@code CZE}.
```

kober32:
Hey @copilot , are te frontend-facing API for the countryCode (1 response and 1 request) optional?

kober32 (approved):
lgtm

PR #1671, opened 2026-04-02
Fix #1659: Set specific subjectId for enrollment-server-onboarding audit records

### LINKED ISSUE #1659: Set specific subjectId for enrollment-server-onboarding audit records
### Description

enrollment-server-onboarding shall set the `subjectId` to the audit records for individual audit types as follows:
| Audit type                     | Subject id   |
|--------------------------------|--------------|
| PROCESS                        | processId    |
| OTP                            | otpId       |
| IDENTITY_VERIFICATION          | identityVerificationId       |
| ACTIVATION                     | activationId |
| DOCUMENT_VERIFICATION          | documentVerificationId       |
| PRESENCE_CHECK_PROVIDER        | identityVerificationId       |
| DOCUMENT_VERIFICATION_PROVIDER | identityVerificationId       |
| ONBOARDING_PROVIDER            | processId    |

### Acceptance criteria

_No response_

### Technical specification

ONBOARDING_PROVIDER type should have a better subject, but we are using it in the scope of the main onboarding process, so the processId is the only possible one.

### QA specification

_No response_

banterCZ:


enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -799,13 +795,16 @@ private static DocumentDataEntity buildDocumentData(final DocumentVerificationDa
         return documentData;
     }
 
-    private void auditMicroblinkResponse(final Map<DocumentType, DocumentVerificationResponseBundle> microblinkResponseByDocumentType, final OwnerId ownerId) {
-        for (final var entry : microblinkResponseByDocumentType.entrySet()) {
-            final var documentType = entry.getKey();
-            final var microblinkResponse = entry.getValue().getResponseWithoutPersonalData();
-
-            auditService.auditDocumentVerificationProvider(ownerId, microblinkResponse, "Document verification response, user: {}, provider: Microblink, documentType: {}", ownerId.getUserId(), documentType);
```
banterCZ, line 807: It used to audit `provider: Microblink` but now it is lower case `provider: microblink` since taken from configuration properties. Is the case important?

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -138,7 +134,6 @@ public DocumentsSubmitResult submitDocuments(OwnerId ownerId, List<SubmittedDocu
 
         final var documentsByTypeAndSide = groupDocumentsByTypeAndSide(documentsVerificationData);
         final var microblinkResponseByDocumentType = fetchMicroblinkResults(documentsByTypeAndSide);
-        auditMicroblinkResponse(microblinkResponseByDocumentType, ownerId);
```
banterCZ, line 141: Auditing moved from the provider to the service, which is more conceptual.

enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/integration/DocumentsSubmitResult.java
```diff
@@ -50,4 +53,8 @@ public class DocumentsSubmitResult {
      */
     private String extractedPhotoId;
 
+    /**
+     * Response without personal data to audit.
+     */
+    private Map<DocumentType, ObjectNode> auditData;
```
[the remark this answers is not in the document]
michal-rozehnal-w: True, but on the other hand, the `ObjectNode` is widely used in our components. In my opinion, `Object` is too generic, and we don’t want to use `String` because it introduces escape characters `\"` in the database when serialized. We also don’t want to use `byte[]` since it should remain human-readable.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
michal-rozehnal-w: I agree, and thank you for improving it. The blame is on me—I did it in a hurry, and it wasn’t the best conceptual solution.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
michal-rozehnal-w: I am not aware of any issue. I agree to use one approach to be consistent

michal-rozehnal-w (approved):
Looks good to me

PR #1679, opened 2026-04-08
Fix #1674: Return consent as not required if user already consent

### LINKED ISSUE #1674: Return consent as not required if user already consent
### Task specification

If from the user lookup endpoint, we receive the optional attribute `consentRequired` in the Identification Stage (i.e. the first call to userLookup). The consent is required iff:
- process is configured for consents
- consentRequired is true from the userLookup Service.

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/rest/RestOnboardingProvider.java
```diff
@@ -76,6 +76,7 @@ public LookupUserResponse lookupUser(final LookupUserRequest request) throws Onb
         logger.debug("Looked up {} for {}", response, request);
         return LookupUserResponse.builder()
                 .userId(response.getUserId())
+                .consentRequired(response.getConsentRequired() == null || response.getConsentRequired())
```
banterCZ: I am waiting for the feedback from the current customers, but anyway, even if they return `null`, we evaluate that as _required_, which is expected behavior for all previous onboarding processes.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -541,9 +542,11 @@ private OnboardingProcessEntity createNewProcessAndLookupUser(
 
         final OnboardingProcessEntity process = createNewProcess(request, identificationData, requestContext);
         logger.debug("Created process ID: {}", process.getId());
-        final String userId = lookupUserService.lookupUser(process, request.identification()).orElse(null);
+        final Optional<LookupUserResponse> lookupUserResponse = lookupUserService.lookupUser(process, request.identification());
+        final String userId = lookupUserResponse.map(LookupUserResponse::getUserId).orElse(null);
         process.setUserId(userId);
-        auditService.audit(process, "Process started for user: {}", userId);
+        process.setConsentAccepted(lookupUserResponse.map(LookupUserResponse::isConsentNotRequired).orElse(false));
```
banterCZ: If not _required_, we mark consent as accepted.

docs/onboarding/External-Onboarding-Services.md
```diff
@@ -179,18 +179,16 @@ Service to identify the prospect and assign user identifier.
...
 
-##### Response  Params
+##### Response Params
 
-| Attribute                      | Type      | Description                                                                                                                            |
-|:-------------------------------|:----------|:---------------------------------------------------------------------------------------------------------------------------------------|
-| `userId`                       | `String`  | The bank assigned ID for the user being onboarded.                                                                                     |
-| `consent`                      | `String`  | Tells if the user has to consent the onboarding (REQUIRED/NOT_REQUIRED) or if it's not evaluated by backend (NOT_EVALUATED or `null`). |
-| `consentRequired (deprecated)` | `Boolean` | Tells if the user has to consent the onboarding.                                                                                       |
+| Attribute         | Type      | Description                                                                     |
```
banterCZ: @tech-kratas please review at least the documentation.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
banterCZ: 🤔 Should we take `processConfig.consentRequired()` into account?

michal-rozehnal-w: Good question, @tech-kratas which should take priority: the consent setting from userLookup service or the onboarding process configuration stored in the database?

michal-rozehnal-w: Nitpick: We can have variable for the consent and then use it in setter:

```
final var consentAccepted = lookupUserResponse.map(LookupUserResponse::isConsentNotRequired).orElse(false)

process.setConsentAccepted(consentAccepted);
```

And I would consider using `!consentRequired` instead of `isConsentNotRequired`, but this is more personal preference

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/model/response/LookupUserResponse.java
```diff
@@ -40,9 +40,7 @@ public final class LookupUserResponse {
     @NonNull
     private String userId;
 
-    // not propagated yet; consistent with the client which always considers it as true
-    @Builder.Default
-    private boolean consentRequired = true;
+    private boolean consentRequired;
```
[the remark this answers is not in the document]
michal-rozehnal-w: It is ok here. The documentation says:

> Tells if the user has to consent the onboarding, `null` is evaluated as `true`.

But the `null` is handled in `UserLookupResponseDto`, which is response from the REST API

michal-rozehnal-w (approved):
Looks good from BE perspective. Let's wait for approval from @tech-kratas

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
banterCZ: It is not much about priority, but data cleanliness. I mean not to set `consentAccepted` if the process is not configured to do so.

banterCZ: I prefer the method name. The exclamation mark could be overlooked, compare `!StringUtils.isBlank()` versus `StringUtils.isNotBlank()`. Moreover, we may use a method reference, which is not possible for `r -> !r.isConsentRequired`.

tech-kratas: I agree, we should take processConfig into account. We shouldn't set the value if the process isn't configured to use the consents.

docs/onboarding/External-Onboarding-Services.md
tech-kratas: I would add the note that `true` only works if the process is configured to use the consents (it's related to [this discussion](https://github.com/wultra/enrollment-server/pull/1679#discussion_r3051010149)).

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
banterCZ: Ok, I will update the logic.
The value will be `false`, in that case, because the column is _not null_.

docs/onboarding/External-Onboarding-Services.md
banterCZ: Good idea, we should cross-link the documentation.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
```diff
@@ -541,12 +542,34 @@ private OnboardingProcessEntity createNewProcessAndLookupUser(
...
         final OnboardingProcessEntity process = createNewProcess(request, identificationData, requestContext);
         logger.debug("Created process ID: {}", process.getId());
-        final String userId = lookupUserService.lookupUser(process, request.identification()).orElse(null);
+
+        final Optional<LookupUserResponse> lookupUserResponse = lookupUserService.lookupUser(process, request.identification());
+        final String userId = lookupUserResponse.map(LookupUserResponse::getUserId).orElse(null);
         process.setUserId(userId);
-        auditService.audit(process, "Process started for user: {}", userId);
+        storeConsent(process, lookupUserResponse.map(LookupUserResponse::isConsentNotRequired).orElse(false));
+        auditService.audit(process, "Process started for user: {}", process.getUserId());
```
michal-rozehnal-w, line 550: Do we want to store `consentRequired` from `LookupUserResponse` in the audit log? It may not be necessary—just asking.

michal-rozehnal-w (approved):
Looks good

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
banterCZ: It is up to analytics. From my point of view, it is no so important to audit it separately. We should always consider data load.

We already audit
```java
auditService.auditOnboardingProvider(process, "Looked up user: {}", response.getUserId());
```
So we may add the consent result there.

@tech-kratas any opinion?

tech-kratas (approved):
looks fine from documentation (and logic) perspective

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingServiceImpl.java
banterCZ: 🤔 But an interesting finding is that if we audit it in the provider, we should map `null` in the layer above. I will update it.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/rest/RestOnboardingProvider.java
banterCZ: Confirmed, the current customer does not send it, and in their case, consent is required.

PR #1680, opened 2026-04-08
Fix #1678: Process data are not cleaned after defined period

### LINKED ISSUE #1678: Process data are not cleaned after defined period

_No response_

### Environment

_No response_

Finish the onboarding process (doesn't matter the result)

expected result: data are deleted after retention period after the process ends
actual result: data are deleted after the process expires.

### Log output

_No response_

michal-rozehnal-w:
The property `enrollment-server-onboarding.identity-verification.data-retention` controls personal data retention for completed processes. There is only one retention period for all process types.
A process is considered complete when it is in the `FINISHED` or `FAILED` state.

The retention time is counted from `timestamp_finished` or `timestamp_failed` from `es_onboarding_proces` table, with a default value of 1 hour (we can change it). After the retention period, records related to the process are deleted from the following tables:
- `es_document_data`
- `es_processed_document`
- `es_selfie`

A new column `es_onboarding_process.timestamp_personal_data_cleaned` has been added. It is used by the cleaning task to ensure that each process is cleaned only once.

### DIFF
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
@@ -6,14 +6,14 @@
     <changeSet id="1" author="Michal Rozehnal">
         <preConditions onFail="MARK_RAN">
             <not>
-                <columnExists tableName="es_onboarding_process" columnName="timestamp_identity_data_cleaned"/>
+                <columnExists tableName="es_onboarding_process" columnName="timestamp_personal_data_cleaned"/>
             </not>
         </preConditions>
         <comment>
-            Create a new column timestamp_identity_data_cleaned in es_onboarding_process table to store the timestamp when the identity data was cleaned
+            Create a new column timestamp_personal_data_cleaned in es_onboarding_process table to store the timestamp when the identity data was cleaned
         </comment>
         <addColumn tableName="es_onboarding_process">
-            <column name="timestamp_identity_data_cleaned" type="timestamp" />
+            <column name="timestamp_personal_data_cleaned" type="timestamp" />
         </addColumn>
     </changeSet>
 
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
@@ -18,7 +18,7 @@
     </changeSet>
 
     <changeSet id="2" author="Michal Rozehnal">
-        <preConditions>
+        <preConditions onFail="MARK_RAN">
             <not>
                 <indexExists indexName="es_onboarding_process_timestamp_identity_data_cleaned_idx"/>
             </not>
@@ -32,7 +32,7 @@
     </changeSet>
 
     <changeSet id="3" author="Michal Rozehnal">
-        <preConditions>
+        <preConditions onFail="MARK_RAN">
             <not>
                 <indexExists indexName="es_onboarding_process_timestamp_finished_idx" />
             </not>
@@ -46,7 +46,7 @@
     </changeSet>
 
     <changeSet id="4" author="Michal Rozehnal">
-        <preConditions>
+        <preConditions onFail="MARK_RAN">
             <not>
                 <indexExists indexName="es_onboarding_process_timestamp_failed_idx" />
             </not>
@@ -60,7 +60,7 @@
     </changeSet>
 
     <changeSet id="5" author="Michal Rozehnal">
-        <preConditions>
+        <preConditions onFail="MARK_RAN">
             <not>
                 <indexExists indexName="es_identity_verification_process_id_idx" />
             </not>
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.33.xsd">
+
+    <changeSet id="1" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="es_onboarding_process" columnName="timestamp_identity_data_cleaned"/>
+            </not>
+        </preConditions>
+        <comment>
+            Create a new column timestamp_identity_data_cleaned in es_onboarding_process table to store the timestamp when the identity data was cleaned
+        </comment>
+        <addColumn tableName="es_onboarding_process">
+            <column name="timestamp_identity_data_cleaned" type="timestamp" />
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="2" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_onboarding_process_timestamp_identity_data_cleaned_idx"/>
+            </not>
+        </preConditions>
+        <comment>
+            Create a new index es_onboarding_process_timestamp_identity_data_cleaned_idx
+        </comment>
+        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_identity_data_cleaned_idx">
+            <column name="timestamp_identity_data_cleaned" />
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="3" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_onboarding_process_timestamp_finished_idx" />
+            </not>
+        </preConditions>
+        <comment>
+            Create a new index es_onboarding_process_timestamp_finished_idx
+        </comment>
+        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_finished_idx">
+            <column name="timestamp_finished"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="4" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_onboarding_process_timestamp_failed_idx" />
+            </not>
+        </preConditions>
+        <comment>
+            Create a new index es_onboarding_process_timestamp_failed_idx
+        </comment>
+        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_failed_idx">
+            <column name="timestamp_failed"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="5" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_identity_verification_process_id_idx" />
+            </not>
+        </preConditions>
+        <comment>
+            Create a new index es_identity_verification_process_id_idx
+        </comment>
+        <createIndex tableName="es_identity_verification" indexName="es_identity_verification_process_id_idx">
+            <column name="process_id"/>
+        </createIndex>
+    </changeSet>
+
+</databaseChangeLog>
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
@@ -73,4 +73,34 @@
         </createIndex>
     </changeSet>
 
+    <changeSet id="6" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="es_document_data" columnName="document_verification_id"/>
+            </not>
+        </preConditions>
+        <comment>
+            Create a new column document_verification_id in es_document_data table
+        </comment>
+        <addColumn tableName="es_document_data">
+            <column name="document_verification_id" type="varchar(36)">
+                <constraints foreignKeyName="es_document_verification_id_fk" referencedTableName="es_document_verification" referencedColumnNames="id" />
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="7" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_document_data_document_verification_id_idx" />
+            </not>
+        </preConditions>
+        <comment>
+            Create a new index es_document_data_document_verification_id_idx
+        </comment>
+        <createIndex tableName="es_document_data" indexName="es_document_data_document_verification_id_idx">
+            <column name="document_verification_id" />
+        </createIndex>
+    </changeSet>
+
 </databaseChangeLog>
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-personal-data-cleaned-timestamp.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-personal-data-cleaned-timestamp.xml
@@ -20,46 +20,18 @@
     <changeSet id="2" author="Michal Rozehnal">
         <preConditions onFail="MARK_RAN">
             <not>
-                <indexExists indexName="es_onboarding_process_timestamp_identity_data_cleaned_idx"/>
+                <indexExists indexName="es_onboarding_process_timestamp_personal_data_cleaned_idx"/>
             </not>
         </preConditions>
         <comment>
-            Create a new index es_onboarding_process_timestamp_identity_data_cleaned_idx
+            Create a new index es_onboarding_process_timestamp_personal_data_cleaned_idx
         </comment>
-        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_identity_data_cleaned_idx">
-            <column name="timestamp_identity_data_cleaned" />
+        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_personal_data_cleaned_idx">
+            <column name="timestamp_personal_data_cleaned" />
         </createIndex>
     </changeSet>
 
     <changeSet id="3" author="Michal Rozehnal">
-        <preConditions onFail="MARK_RAN">
-            <not>
-                <indexExists indexName="es_onboarding_process_timestamp_finished_idx" />
-            </not>
-        </preConditions>
-        <comment>
-            Create a new index es_onboarding_process_timestamp_finished_idx
-        </comment>
-        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_finished_idx">
-            <column name="timestamp_finished"/>
-        </createIndex>
-    </changeSet>
-
-    <changeSet id="4" author="Michal Rozehnal">
-        <preConditions onFail="MARK_RAN">
-            <not>
-                <indexExists indexName="es_onboarding_process_timestamp_failed_idx" />
-            </not>
-        </preConditions>
-        <comment>
-            Create a new index es_onboarding_process_timestamp_failed_idx
-        </comment>
-        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_failed_idx">
-            <column name="timestamp_failed"/>
-        </createIndex>
-    </changeSet>
-
-    <changeSet id="5" author="Michal Rozehnal">
         <preConditions onFail="MARK_RAN">
             <not>
                 <indexExists indexName="es_identity_verification_process_id_idx" />
@@ -73,7 +45,7 @@
         </createIndex>
     </changeSet>
 
-    <changeSet id="6" author="Michal Rozehnal">
+    <changeSet id="4" author="Michal Rozehnal">
         <preConditions onFail="MARK_RAN">
             <not>
                 <columnExists tableName="es_document_data" columnName="document_verification_id"/>
@@ -89,7 +61,7 @@
         </addColumn>
     </changeSet>
 
-    <changeSet id="7" author="Michal Rozehnal">
+    <changeSet id="5" author="Michal Rozehnal">
         <preConditions>
             <not>
                 <indexExists indexName="es_document_data_document_verification_id_idx" />
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-personal-data-cleaned-timestamp.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-personal-data-cleaned-timestamp.xml
@@ -62,7 +62,7 @@
     </changeSet>
 
     <changeSet id="5" author="Michal Rozehnal">
-        <preConditions>
+        <preConditions onFail="MARK_RAN">
             <not>
                 <indexExists indexName="es_document_data_document_verification_id_idx" />
             </not>
--- a/docs/onboarding/Configuration-Properties.md
+++ b/docs/onboarding/Configuration-Properties.md
@@ -29,18 +29,18 @@ The configuration properties below are global to all the onboarding processes.
 Specific process options are defined in the database.
 See [Configuration of Onboarding Process](./Configuration-Onboarding-Process.md) for the details of JSON configuration.
 
-| Property                                                                            | Default   | Note                                                                                                   |
-|-------------------------------------------------------------------------------------|-----------|--------------------------------------------------------------------------------------------------------|
-| `enrollment-server-onboarding.onboarding-process.otp.length`                        | `8`       | Length of generated digital OTP codes.                                                                 |
-| `enrollment-server-onboarding.onboarding-process.otp.expiration`                    | `5m`      | Expiration time for OTP codes.                                                                         |
-| `enrollment-server-onboarding.onboarding-process.otp.max-failed-attempts`           | `5`       | Maximum number of failed attempts for OTP verification.                                                |
-| `enrollment-server-onboarding.onboarding-process.otp.resend-period`                 | `30s`     | A time period after which next OTP can be sent.                                                        |
-| `enrollment-server-onboarding.onboarding-process.expiration`                        | `3h`      | Onboarding process expiration time.                                                                    |
-| `enrollment-server-onboarding.onboarding-process.activation.expiration`             | `5m`      | Expiration of activations used within an onboarding process.                                           |
-| `enrollment-server-onboarding.onboarding-process.verification.expiration`           | `1h`      | Expiration of identity verification within an onboarding process.                                      |
-| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`             | `5`       | Maximum number of onboarding processes during last 24 hours per user.                                  |
-| `enrollment-server-onboarding.onboarding-process.max-error-score`                   | `15`      | Maximum error score for an onboarding process.                                                         |
-| `enrollment-server-onboarding.onboarding-process.default-type`                      | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application.          |
+| Property                                                                   | Default   | Note                                                                                            |
+|----------------------------------------------------------------------------|-----------|-------------------------------------------------------------------------------------------------|
+| `enrollment-server-onboarding.onboarding-process.otp.length`               | `8`       | Length of generated digital OTP codes.                                                          |
+| `enrollment-server-onboarding.onboarding-process.otp.expiration`           | `5m`      | Expiration time for OTP codes.                                                                  |
+| `enrollment-server-onboarding.onboarding-process.otp.max-failed-attempts`  | `5`       | Maximum number of failed attempts for OTP verification.                                         |
+| `enrollment-server-onboarding.onboarding-process.otp.resend-period`        | `30s`     | A time period after which next OTP can be sent.                                                 |
+| `enrollment-server-onboarding.onboarding-process.expiration`               | `3h`      | Onboarding process expiration time.                                                             |
+| `enrollment-server-onboarding.onboarding-process.activation.expiration`    | `5m`      | Expiration of activations used within an onboarding process.                                    |
+| `enrollment-server-onboarding.onboarding-process.verification.expiration`  | `1h`      | Expiration of identity verification within an onboarding process.                               |
+| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`    | `5`       | Maximum number of onboarding processes during last 24 hours per user.                           |
+| `enrollment-server-onboarding.onboarding-process.max-error-score`          | `15`      | Maximum error score for an onboarding process.                                                  |
+| `enrollment-server-onboarding.onboarding-process.default-type`             | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application.   |
 
 ## Identity Verification Configuration
 
--- a/docs/onboarding/Configuration-Properties.md
+++ b/docs/onboarding/Configuration-Properties.md
@@ -29,18 +29,19 @@ The configuration properties below are global to all the onboarding processes.
 Specific process options are defined in the database.
 See [Configuration of Onboarding Process](./Configuration-Onboarding-Process.md) for the details of JSON configuration.
 
-| Property                                                                   | Default   | Note                                                                                            |
-|----------------------------------------------------------------------------|-----------|-------------------------------------------------------------------------------------------------|
-| `enrollment-server-onboarding.onboarding-process.otp.length`               | `8`       | Length of generated digital OTP codes.                                                          |
-| `enrollment-server-onboarding.onboarding-process.otp.expiration`           | `5m`      | Expiration time for OTP codes.                                                                  |
-| `enrollment-server-onboarding.onboarding-process.otp.max-failed-attempts`  | `5`       | Maximum number of failed attempts for OTP verification.                                         |
-| `enrollment-server-onboarding.onboarding-process.otp.resend-period`        | `30s`     | A time period after which next OTP can be sent.                                                 |
-| `enrollment-server-onboarding.onboarding-process.expiration`               | `3h`      | Onboarding process expiration time.                                                             |
-| `enrollment-server-onboarding.onboarding-process.activation.expiration`    | `5m`      | Expiration of activations used within an onboarding process.                                    |
-| `enrollment-server-onboarding.onboarding-process.verification.expiration`  | `1h`      | Expiration of identity verification within an onboarding process.                               |
-| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`    | `5`       | Maximum number of onboarding processes during last 24 hours per user.                           |
-| `enrollment-server-onboarding.onboarding-process.max-error-score`          | `15`      | Maximum error score for an onboarding process.                                                  |
-| `enrollment-server-onboarding.onboarding-process.default-type`             | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application.   |
+| Property                                                                  | Default   | Note                                                                                          |
+|---------------------------------------------------------------------------|-----------|-----------------------------------------------------------------------------------------------|
+| `enrollment-server-onboarding.onboarding-process.otp.length`              | `8`       | Length of generated digital OTP codes.                                                        |
+| `enrollment-server-onboarding.onboarding-process.otp.expiration`          | `5m`      | Expiration time for OTP codes.                                                                |
+| `enrollment-server-onboarding.onboarding-process.otp.max-failed-attempts` | `5`       | Maximum number of failed attempts for OTP verification.                                       |
+| `enrollment-server-onboarding.onboarding-process.otp.resend-period`       | `30s`     | A time period after which next OTP can be sent.                                               |
+| `enrollment-server-onboarding.onboarding-process.expiration`              | `3h`      | Onboarding process expiration time.                                                           |
+| `enrollment-server-onboarding.onboarding-process.activation.expiration`   | `5m`      | Expiration of activations used within an onboarding process.                                  |
+| `enrollment-server-onboarding.onboarding-process.verification.expiration` | `1h`      | Expiration of identity verification within an onboarding process.                             |
+| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`   | `5`       | Maximum number of onboarding processes during last 24 hours per user.                         |
+| `enrollment-server-onboarding.onboarding-process.max-error-score`         | `15`      | Maximum error score for an onboarding process.                                                |
+| `enrollment-server-onboarding.onboarding-process.default-type`            | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application. |
+| `enrollment-server-onboarding.onboarding-process.cleanup-limit`           | `10000`   | Maximum number of records processed in a single cleaning task run.                            |
 
 ## Identity Verification Configuration
 
--- a/docs/onboarding/Configuration-Properties.md
+++ b/docs/onboarding/Configuration-Properties.md
@@ -41,15 +41,15 @@ See [Configuration of Onboarding Process](./Configuration-Onboarding-Process.md)
 | `enrollment-server-onboarding.onboarding-process.max-processes-per-day`             | `5`       | Maximum number of onboarding processes during last 24 hours per user.                                  |
 | `enrollment-server-onboarding.onboarding-process.max-error-score`                   | `15`      | Maximum error score for an onboarding process.                                                         |
 | `enrollment-server-onboarding.onboarding-process.default-type`                      | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application.          |
-| `enrollment-server-onboarding.onboarding-process.completedProcessDataRetentionTime` | `1h`      | Time period for which the identity data of completed onboarding processes is retained in the database. |
 
 ## Identity Verification Configuration
 
-| Property                                                                                 | Default | Note                                                              |
-|------------------------------------------------------------------------------------------|---------|-------------------------------------------------------------------|
-| `enrollment-server-onboarding.identity-verification.enabled`                             | `false` | Whether identity verification is enabled.                         |
-| `enrollment-server-onboarding.identity-verification.max-failed-attempts`                 | `5`     | Maximum failed attempts for identity verification.                |
-| `enrollment-server-onboarding.identity-verification.max-failed-attempts-document-upload` | `5`     | Maximum failed attempts for document upload.                      |
+| Property                                                                                 | Default | Note                                                                                                   |
+|------------------------------------------------------------------------------------------|---------|--------------------------------------------------------------------------------------------------------|
+| `enrollment-server-onboarding.identity-verification.enabled`                             | `false` | Whether identity verification is enabled.                                                              |
+| `enrollment-server-onboarding.identity-verification.max-failed-attempts`                 | `5`     | Maximum failed attempts for identity verification.                                                     |
+| `enrollment-server-onboarding.identity-verification.max-failed-attempts-document-upload` | `5`     | Maximum failed attempts for document upload.                                                           |
+| `enrollment-server-onboarding.identity-verification.data-retention`                      | `1h`    | Time period for which the identity data of completed onboarding processes is retained in the database. |
 
 ## Digital Onboarding Adapter Configuration
 
--- a/docs/onboarding/Configuration-Properties.md
+++ b/docs/onboarding/Configuration-Properties.md
@@ -29,18 +29,19 @@ The configuration properties below are global to all the onboarding processes.
 Specific process options are defined in the database.
 See [Configuration of Onboarding Process](./Configuration-Onboarding-Process.md) for the details of JSON configuration.
 
-| Property                                                                  | Default   | Note                                                                                          |
-|---------------------------------------------------------------------------|-----------|-----------------------------------------------------------------------------------------------|
-| `enrollment-server-onboarding.onboarding-process.otp.length`              | `8`       | Length of generated digital OTP codes.                                                        |
-| `enrollment-server-onboarding.onboarding-process.otp.expiration`          | `5m`      | Expiration time for OTP codes.                                                                |
-| `enrollment-server-onboarding.onboarding-process.otp.max-failed-attempts` | `5`       | Maximum number of failed attempts for OTP verification.                                       |
-| `enrollment-server-onboarding.onboarding-process.otp.resend-period`       | `30s`     | A time period after which next OTP can be sent.                                               |
-| `enrollment-server-onboarding.onboarding-process.expiration`              | `3h`      | Onboarding process expiration time.                                                           |
-| `enrollment-server-onboarding.onboarding-process.activation.expiration`   | `5m`      | Expiration of activations used within an onboarding process.                                  |
-| `enrollment-server-onboarding.onboarding-process.verification.expiration` | `1h`      | Expiration of identity verification within an onboarding process.                             |
-| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`   | `5`       | Maximum number of onboarding processes during last 24 hours per user.                         |
-| `enrollment-server-onboarding.onboarding-process.max-error-score`         | `15`      | Maximum error score for an onboarding process.                                                |
-| `enrollment-server-onboarding.onboarding-process.default-type`            | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application. |
+| Property                                                                            | Default   | Note                                                                                                   |
+|-------------------------------------------------------------------------------------|-----------|--------------------------------------------------------------------------------------------------------|
+| `enrollment-server-onboarding.onboarding-process.otp.length`                        | `8`       | Length of generated digital OTP codes.                                                                 |
+| `enrollment-server-onboarding.onboarding-process.otp.expiration`                    | `5m`      | Expiration time for OTP codes.                                                                         |
+| `enrollment-server-onboarding.onboarding-process.otp.max-failed-attempts`           | `5`       | Maximum number of failed attempts for OTP verification.                                                |
+| `enrollment-server-onboarding.onboarding-process.otp.resend-period`                 | `30s`     | A time period after which next OTP can be sent.                                                        |
+| `enrollment-server-onboarding.onboarding-process.expiration`                        | `3h`      | Onboarding process expiration time.                                                                    |
+| `enrollment-server-onboarding.onboarding-process.activation.expiration`             | `5m`      | Expiration of activations used within an onboarding process.                                           |
+| `enrollment-server-onboarding.onboarding-process.verification.expiration`           | `1h`      | Expiration of identity verification within an onboarding process.                                      |
+| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`             | `5`       | Maximum number of onboarding processes during last 24 hours per user.                                  |
+| `enrollment-server-onboarding.onboarding-process.max-error-score`                   | `15`      | Maximum error score for an onboarding process.                                                         |
+| `enrollment-server-onboarding.onboarding-process.default-type`                      | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application.          |
+| `enrollment-server-onboarding.onboarding-process.completedProcessDataRetentionTime` | `1h`      | Time period for which the identity data of completed onboarding processes is retained in the database. |
 
 ## Identity Verification Configuration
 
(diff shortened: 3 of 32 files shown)


enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessEntity.java
```diff
@@ -148,6 +148,12 @@ public class OnboardingProcessEntity implements Serializable {
     @NotNull
     private OnboardingProcessConfigurationEntity processConfiguration;
 
+    /**
+     * Timestamp when the identity data (e.g. selfie, uploaded documents) was cleaned up.
+     */
+    @Column(name = "timestamp_identity_data_cleaned")
+    private Date timestampIdentityDataCleaned;
```
michal-rozehnal-w: This could just be a boolean column, but I think a timestamp conveys more information than a flag and could help with any potential investigation in the feature.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
```diff
@@ -196,4 +197,38 @@ default List<String> findActiveByTimestampWithLock(Date dateCreatedBefore) {
...
+               dv.id AS documentVerificationId,
+               dv.uploadId AS uploadId
+        FROM OnboardingProcessEntity p
+        JOIN IdentityVerificationEntity iv ON iv.processId = p.id
+        JOIN DocumentVerificationEntity dv ON dv.identityVerification.id = iv.id
+        WHERE p.status IN :statuses
+          AND (p.timestampFinished < :processCompletedBefore OR p.timestampFailed < :processCompletedBefore)
+          AND p.timestampIdentityDataCleaned IS NULL
+        """)
+    List<OnboardingProcessIdentityDataIdsView> findIdentityDataForCleanup(final Set<OnboardingStatus> statuses, final Date processCompletedBefore);
```
michal-rozehnal-w: There are two joins, which is usually a red flag for performance. However, I checked that we have indexes on all columns and adde missing ones. Alternatively, this could be transformed into subqueries or separate calls in the service layer. Since I am not a database expert, I welcome any other ideas.

Regarding the condition, the process is considered complete when the status is `FINISHED` or `FAILED`.
- `FINISHED` is set together with `timestampFinished`
- `FAILED` is set together with `timestampFailed`
I checked in the code that these timestamp are always set together with the status.

We also have a cleanup task for expired processes, which ensures that the `FAILED` status is set.

The `p.timestampIdentityDataCleaned IS NULL` condition ensures that a process is not cleaned more than once. It is an independent operation, but the number of processes matching the condition would grow with each run.
Alternatively, I considered using the condition `p.timestampCreated > (now - process_expiration - data_retention)`, but this would still process the same records multiple times and is not reliable—for example, if the server is down for a few hours, some entities could be stored indefinitely, which is not acceptable.

One more concern regarding the version upgrade: the first run will process a lot of records because `timestampIdentityDataCleaned` will be `null` for all historical records, causing them to be processed.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/configuration/OnboardingConfig.java
```diff
@@ -59,4 +59,10 @@ public class OnboardingConfig extends CommonOnboardingConfig {
     @Value("${enrollment-server-onboarding.onboarding-process.default-type:}")
     private String defaultProcessType;
 
+    /**
+     * How long are identity data (e.g. selfie, uploaded documents) kept in database after the process is completed.
+     */
+    @Value("${enrollment-server-onboarding.onboarding-process.completedProcessDataRetentionTime:1h}")
```
michal-rozehnal-w: Feel free to propose different default value

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningTask.java
```diff
@@ -150,4 +114,16 @@ public void cleanupActivations() {
         logger.debug("cleanupActivations");
         activationCleaningService.cleanupActivations();
     }
+
+    /**
+     * Cleanup identity data of completed onboarding processes.
+     */
+    @Scheduled(fixedDelayString = "PT15S", initialDelayString = "PT10S")
+    @SchedulerLock(name = SchedulerLockNames.CLEANUP_COMPLETED_PROCESS_IDENTITY_DATA_LOCK, lockAtMostFor = "5m")
+    public void cleanupCompletedProcessIdentityData() {
```
michal-rozehnal-w: Also feel free to propose different values for the scheduler

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -197,8 +167,60 @@ public void terminateExpiredIdentityVerifications() {
...
+        logger.info("Deleted {} data records of completed processes", counter);
+    }
+
+    private void cleanCompletedProcessIdentityData(
+            final Set<String> processIds,
+            final Set<String> identityVerificationIds,
+            final Set<String> documentVerificationIds,
+            final Set<String> uploadIds
+    ) {
+        selfieRepository.deleteAllByIdentityVerificationIds(identityVerificationIds);
```
michal-rozehnal-w: Do we want some audit logging here?

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,76 @@
...
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="es_onboarding_process" columnName="timestamp_identity_data_cleaned"/>
+            </not>
+        </preConditions>
+        <comment>
+            Create a new column timestamp_identity_data_cleaned in es_onboarding_process table to store the timestamp when the identity data was cleaned
+        </comment>
+        <addColumn tableName="es_onboarding_process">
+            <column name="timestamp_identity_data_cleaned" type="timestamp" />
```
michal-rozehnal-w: I use the term `identity data` for all personal data, such as selfies, images of uploaded documents, extracted face photos, etc. Let me know if you have a better name for it.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/document/DocumentProcessingService.java
```diff
@@ -373,6 +373,8 @@ private static byte[] decodeBase64Data(final String base64Data) {
 
     private void processDocsSubmitResults(OwnerId ownerId, DocumentVerificationEntity docVerification,
                                           DocumentsSubmitResult docsSubmitResults, DocumentSubmitResult docSubmitResult) {
+        docVerification.setUploadId(docSubmitResult.getUploadId());
```
michal-rozehnal-w, line 376: Not directly related to the cleanup—this is a bug found during testing.

Originally, the `uploadId` was stored in `es_document_verification` only if the document upload was successful. However, the transaction is not rolled back in case of rejection or failure—in these cases, different values are stored in the database.

As a result, for unsuccessful outcomes, the `uploadId` was never set (resulting in a `null` value in the database), and records in `es_document_data` were not linked to `es_document_verification`. Consequently, the cleanup task would never fetch them, and they would remain in the database indefinitely.

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/task/cleaning/CleaningServiceTest.java
```diff
@@ -268,6 +246,61 @@ void testPartition_tooSmall() {
...
+
+        tested.cleanupCompletedProcessIdentityData();
+        entityManager.flush();
+        entityManager.clear();
+
+        final var failedProcess = fetchOnboardingProcess(failedProcessId);
+        assertEquals(System.currentTimeMillis(), failedProcess.getTimestampIdentityDataCleaned().getTime(), TIME_ASSERT_DELTA);
+
+        final var successfulProcess = fetchOnboardingProcess(successfulProcessId);
+        assertEquals(System.currentTimeMillis(), successfulProcess.getTimestampIdentityDataCleaned().getTime(), TIME_ASSERT_DELTA);
```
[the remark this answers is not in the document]
michal-rozehnal-w: It should be fine, this is reason why the `TIME_ASSERT_DELTA` is 3 seconds.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-personal-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,76 @@
...
+            <column name="timestamp_identity_data_cleaned" type="timestamp" />
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="2" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_onboarding_process_timestamp_identity_data_cleaned_idx"/>
+            </not>
+        </preConditions>
```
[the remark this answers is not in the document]
michal-rozehnal-w: Yes, I completely missed this, fixed

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,76 @@
...
+            <column name="timestamp_identity_data_cleaned" />
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="3" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_onboarding_process_timestamp_finished_idx" />
+            </not>
+        </preConditions>
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,76 @@
...
+            <column name="timestamp_finished"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="4" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_onboarding_process_timestamp_failed_idx" />
+            </not>
+        </preConditions>
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,76 @@
...
+            <column name="timestamp_failed"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="5" author="Michal Rozehnal">
+        <preConditions>
+            <not>
+                <indexExists indexName="es_identity_verification_process_id_idx" />
+            </not>
+        </preConditions>
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -18,3 +18,13 @@ Added a new indexed column `subject_id` holding an identifier linking the audit
 <!-- begin box warning -->
 The auditing tables may be already updated in your database schema if the database schema is not separated for different PowerAuth applications. In case the column `audit_log.subject_id` and its index `audit_log_subject_id_idx` are already present, you can safely skip this migration step.
 <!-- end -->
+
+
+### Onboarding process identity data retention
+
+The identity data retention period is configured in the new `property enrollment-server-onboarding.onboarding-process.completedProcessDataRetentionTime`.
```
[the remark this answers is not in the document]
michal-rozehnal-w: Applied

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -197,8 +167,62 @@ public void terminateExpiredIdentityVerifications() {
...
+        final var identityVerificationIds = new HashSet<String>();
+        final var documentVerificationIds = new HashSet<String>();
+        final var uploadIds = new HashSet<String>();
+
+        var counter = 0;
+        for (final var dataIdsItem : dataIds) {
+            processIds.add(dataIdsItem.getProcessId());
+            identityVerificationIds.add(dataIdsItem.getIdentityVerificationId());
+            documentVerificationIds.add(dataIdsItem.getDocumentVerificationId());
+            Optional.of(dataIdsItem.getUploadId())
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

michal-rozehnal-w:
@tech-kratas @zcgandcomp you don’t need to read the code—I summarized the changes in the PR description. I have a few follow-up questions:

1. What should the default retention period be? I set it to 1 hour after process completion, but it should probably be longer?
2. Do we want any auditing around deleting records from the `es_selfie`, `es_document_data`, or `es_processed_document_data` tables? Historically, we haven’t had any.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
banterCZ: 🤔 What about `timestamp_data_cleaned`, `timestamp_personal_data_cleaned` or `timestamp_pii_data_cleaned`?
Moreover, I miss this change in the migration guide.

docs/onboarding/Configuration-Properties.md
```diff
@@ -29,18 +29,19 @@ The configuration properties below are global to all the onboarding processes.
...
+| `enrollment-server-onboarding.onboarding-process.verification.expiration`           | `1h`      | Expiration of identity verification within an onboarding process.                                      |
+| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`             | `5`       | Maximum number of onboarding processes during last 24 hours per user.                                  |
+| `enrollment-server-onboarding.onboarding-process.max-error-score`                   | `15`      | Maximum error score for an onboarding process.                                                         |
+| `enrollment-server-onboarding.onboarding-process.default-type`                      | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application.          |
+| `enrollment-server-onboarding.onboarding-process.completedProcessDataRetentionTime` | `1h`      | Time period for which the identity data of completed onboarding processes is retained in the database. |
```
banterCZ: I think that global configuration is fine (and easier to handle), but just asking whether we don't need specific settings per process.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessEntity.java
banterCZ: Why not. It is good for auditing, but for queries, the index is not so effective for `where timestamp_identity_data_cleaned is not null` compared to a simple boolean, is it?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessIdentityDataIdsView.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.wultra.app.onboardingserver.common.database.entity;
+
+
+/**
+ * View for {@link OnboardingProcessEntity} identity data IDs.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public interface OnboardingProcessIdentityDataIdsView {
```
banterCZ: 🤔 Is it really a view or a projection only?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
banterCZ: I don't have a strong opinion about the two joins, but thanks for pointing that out. I let it on @zcgandcomp
Regarding the data of the existing customers, indeed, they should be migrated.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -197,8 +167,62 @@ public void terminateExpiredIdentityVerifications() {
...
+
+        if (counter % BATCH_SIZE != 0) {
+            cleanCompletedProcessIdentityData(processIds, identityVerificationIds, documentVerificationIds, uploadIds);
+        }
+
+        logger.info("Deleted {} data records of completed processes", counter);
+    }
+
+    private void cleanCompletedProcessIdentityData(
+            final Set<String> processIds,
```
banterCZ: In general, I consider having several parameters of the same type error-prone.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -197,8 +167,62 @@ public void terminateExpiredIdentityVerifications() {
...
+            final Set<String> identityVerificationIds,
+            final Set<String> documentVerificationIds,
+            final Set<String> uploadIds
+    ) {
+        selfieRepository.deleteAllByIdentityVerificationIds(identityVerificationIds);
+        processedDocumentDataRepository.deleteAllByDocumentVerificationIds(documentVerificationIds);
+        documentDataRepository.deleteAllById(uploadIds);
+        onboardingProcessRepository.updateIdentityDataCleanup(processIds, new Date());
```
banterCZ: 🤔 May we have the same timestamp for the whole commit?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -197,8 +167,62 @@ public void terminateExpiredIdentityVerifications() {
...
+            return;
+        }
+
+        final var processIds = new HashSet<String>();
+        final var identityVerificationIds = new HashSet<String>();
+        final var documentVerificationIds = new HashSet<String>();
+        final var uploadIds = new HashSet<String>();
+
+        var counter = 0;
+        for (final var dataIdsItem : dataIds) {
```
banterCZ: It seems unnecessarily complicated to me. I suggest keeping _command-query separation_, I mean, collect ids and then execute the change. It is not ideal, but there already is a utility to split a huge input for `in` statement, see `ListUtils.partition`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningTask.java
banterCZ: 🤔 If the retention period is 1 hour, do we need it to run so often? cc @zcgandcomp

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessEntity.java
michal-rozehnal-w: I will check documentation. If index is not helpful then it can be removed

- [x] Check whether the index help in query

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
michal-rozehnal-w: I used the `timestamp_personal_data_cleaned`, thank you for ideas

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessIdentityDataIdsView.java
michal-rozehnal-w: It is not view in database and I understand it can be confusing in the code. I renamed it to `projection`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
michal-rozehnal-w: Same timestamp is used for one task run

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
michal-rozehnal-w: This method was removed during refactoring

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessEntity.java
banterCZ: I mean that the index over a boolean would be more effective.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
michal-rozehnal-w: I wanted to be efficient here, but you’re right—on a second look, it was old-school code and harder to follow. I used streams, we work only with ids here (not entire objects) so performance shouldn't be an issue

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentDataEntity.java
```diff
@@ -66,6 +67,15 @@ public class DocumentDataEntity implements Serializable {
...
     private Date timestampCreated;
 
+    /**
+     * ID of linked {@link DocumentVerificationEntity}.
+     *
+     * @implNote Nullable for backward compatibility. New records should be non-null.
+     */
+    @Column(name = "document_verification_id")
+    @NotNull
+    private String documentVerificationId;
```
michal-rozehnal-w, line 77: I added this binding to ensure that there are no orphan records left in the database permanently. This field also has a foreign key. It is also consistent now with `es_processed_document_data`

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -118,44 +118,59 @@ public DocumentsSubmitResult checkDocumentUpload(OwnerId ownerId, DocumentVerifi
         throw new UnsupportedOperationException("Method checkDocumentUpload is not supported by Microblink provider.");
     }
 
+    /**
+     * @implNote If any DocumentDataEntity is saved, then uploadId must always be returned so it can be bound to the DocumentVerificationEntity.
+     * A foreign key cannot be used for DocumentVerificationEntity.uploadId because, depending on the provider, data may be uploaded to the cloud
+     * and there may be no corresponding DocumentDataEntity record.
+     */
     @Override
     public DocumentsSubmitResult submitDocuments(OwnerId ownerId, List<SubmittedDocument> submittedDocuments) throws DocumentVerificationException, RemoteCommunicationException {
```
michal-rozehnal-w, line 127: This method now always returns an object. In case of an error, it uses errorDetail, which helps handle the result in upper layers without any data loss.
Specific error is not exposed in status endpoints - we always [hide](https://github.com/wultra/enrollment-server/blob/develop/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java#L533) original error.

docs/onboarding/Configuration-Properties.md
zcgandcomp: Per process type, retention is more problematic; one setting is clear.

zcgandcomp: What is wrong with enrollment-server-onboarding.identity-verification.data-retention ?
and at least keep the same style for the property, no snake to camel please.

michal-rozehnal-w: Ok, so we need to be able to configure, for example, that the `onboarding` process retains data for 6 hours, the `reactivation` process for 3 hours, etc., correct? And if there is no specific retention time for the process in config then there should be fallback to default one

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,106 @@
...
+        </comment>
+        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_identity_data_cleaned_idx">
+            <column name="timestamp_identity_data_cleaned" />
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="3" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="es_onboarding_process_timestamp_finished_idx" />
```
zcgandcomp: This is overkill. No performance gain and more cost for table maintenance.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,106 @@
...
+    <changeSet id="4" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="es_onboarding_process_timestamp_failed_idx" />
+            </not>
+        </preConditions>
+        <comment>
+            Create a new index es_onboarding_process_timestamp_failed_idx
+        </comment>
+        <createIndex tableName="es_onboarding_process" indexName="es_onboarding_process_timestamp_failed_idx">
```
zcgandcomp: This is overkill. No performance gain and more cost for table maintenance.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-personal-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,106 @@
...
+            <not>
+                <columnExists tableName="es_document_data" columnName="document_verification_id"/>
+            </not>
+        </preConditions>
+        <comment>
+            Create a new column document_verification_id in es_document_data table
+        </comment>
+        <addColumn tableName="es_document_data">
+            <column name="document_verification_id" type="varchar(36)">
+                <constraints foreignKeyName="es_document_verification_id_fk" referencedTableName="es_document_verification" referencedColumnNames="id" />
```
zcgandcomp, line 59: There is already a relationship es_document_verification.upload_id =  es_document_data.id. So do we need to create new one?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentDataEntity.java
zcgandcomp: Ok, if there are no other blocker we can change the structure. Are we sure this will not affect current instalations?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -197,8 +170,70 @@ public void terminateExpiredIdentityVerifications() {
...
+        for (final var idsBatch : ListUtils.partition(ids, BATCH_SIZE)) {
+            processedDocumentDataRepository.deleteAllByDocumentVerificationIds(idsBatch);
+            documentDataRepository.deleteAllByDocumentVerificationIds(idsBatch);
+        }
+
+        logger.debug("Document data deleted");
+    }
+
+    private void cleanSelfieData(final List<OnboardingProcessPersonalDataIdsProjection> personalDataIds) {
+        final var ids = personalDataIds.stream()
```
zcgandcomp, line 230: the `ids` will contain dulicit IdentityVerificationIds for processes where the identity verification phase is repeated.

zcgandcomp (changes requested):
Check the environment properties.
The logic is fine. There could be other solutions (like marking the process finish timestamp on the entities for deletion), but this one does the assignment.

I was little bit thinking about (forbidden) dirty solution of just summing the periods, because it's dirty but if the periods are changed and longer,  nothing is affected and if the periods are shortened, also nothing is affected (the processes will start expiring en masse). JUST observation, no need to change anything.

There should not be added so many indexes - only one is chosen by the optimizer anyway.

There is a question of changing data structure - it's fine only if we can guarantee there is no impact on the current installations. It is possible as the data are deleted, but I'm not sure.

There are several non-optimal handling of the list of ids for deletion - the select will return duplicates ids.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
michal-rozehnal-w: Index removed

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-identity-data-cleaned-timestamp.xml
michal-rozehnal-w: Index removed

docs/onboarding/Configuration-Properties.md
michal-rozehnal-w: Discussed with @zcgandcomp that we want a single setting for all process types.

michal-rozehnal-w: I initially wanted to keep it associated with the process and reflect in the name that it applies to completed processes. However, on second thought, that was probably too much, and `data-retention` is sufficient. I moved it to the identity verification config and used kebab-case for consistency.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
michal-rozehnal-w: Good catch - I added `.distinct()` here and also in the `setProcessCleanupTime` method.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-personal-data-cleaned-timestamp.xml
michal-rozehnal-w: Yes, we will have two bindings between the tables—via `upload_id` and `document_verification_id`.
The main reason I added `document_verification_id` is that we can use a foreign key to ensure the data is properly linked (it will be NULL only for legacy records).

The issue with `upload_id` is that it is nullable, and for some providers it refers to an entity in an external service. In such cases, there is no corresponding record in `es_document_data`, so we cannot use a foreign key.

I adjusted the `MicroblinkVerificationProvider::submitDocuments` (this is only provider using the `es_document_data` table) method to ensure that `upload_id` is always propagated to the upper layer and linked with `es_document_verification` records. However, I’m concerned this approach is too fragile—it’s easy to make a mistake and fail to propagate `upload_id`. This could result in orphan (unlinked) records in `es_document_data` that would never be cleaned up and would remain in the database indefinitely. I consider this too risky, and in my opinion, the additional binding is justified.

For the future, it would likely be better to use `upload_id` strictly for entities uploaded to third-party services and link `es_document_data` only via `document_verification_id`.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentDataEntity.java
michal-rozehnal-w: The only issue I can see is that during the upgrade, there may be records with `document_verification_id` set to `NULL` that have not yet been deleted by the old cleaning service. Such records need to be cleaned manually—I added a description to the [migration guide](https://github.com/wultra/enrollment-server/pull/1680/changes#diff-0021ab1900364cc274fa1ccd4e61edfbf4322633f57eae36dbf6ab46b47c5192R33). Otherwise, this is just adding a column, which is generally a safe operation. @banterCZ do you have any concerns about this change?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessEntity.java
michal-rozehnal-w: According to postgres 18 [documentation](https://www.postgresql.org/docs/18/indexes-types.html#INDEXES-TYPES-BTREE), it seems the index can help with `IS NULL` condition

docs/onboarding/Configuration-Properties.md
```diff
@@ -29,26 +29,27 @@ The configuration properties below are global to all the onboarding processes.
...
-| `enrollment-server-onboarding.onboarding-process.verification.expiration` | `1h`      | Expiration of identity verification within an onboarding process.                             |
-| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`   | `5`       | Maximum number of onboarding processes during last 24 hours per user.                         |
-| `enrollment-server-onboarding.onboarding-process.max-error-score`         | `15`      | Maximum error score for an onboarding process.                                                |
-| `enrollment-server-onboarding.onboarding-process.default-type`            | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application. |
+| Property                                                                            | Default   | Note                                                                                                   |
+|-------------------------------------------------------------------------------------|-----------|--------------------------------------------------------------------------------------------------------|
```
banterCZ: Nitpick: Couldn't we reformat the table back?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
```diff
@@ -196,4 +197,37 @@ default List<String> findActiveByTimestampWithLock(Date dateCreatedBefore) {
...
+    @Query("""
+        SELECT p.id AS processId,
+               iv.id AS identityVerificationId,
+               dv.id AS documentVerificationId
+        FROM OnboardingProcessEntity p
+        JOIN IdentityVerificationEntity iv ON iv.processId = p.id
+        JOIN DocumentVerificationEntity dv ON dv.identityVerification.id = iv.id
+        WHERE p.status IN :statuses
+          AND (p.timestampFinished < :processCompletedBefore OR p.timestampFailed < :processCompletedBefore)
+          AND p.timestampPersonalDataCleaned IS NULL
```
banterCZ, line 218: 🤔 What should we do with the existing customers? The first run will be absolutely huge if all `timestamp_personal_data_cleaned` are `null`.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/DocumentDataEntity.java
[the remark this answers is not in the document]
michal-rozehnal-w: I was thinking about the same approach as is used for `OnboardingProcessEntity#consentAccepted`. However, there is no dirty checking, and the record is never modified. Once created, it is only used for reading until it is deleted by the cleanup task.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/entity/OnboardingProcessEntity.java
```diff
@@ -148,6 +148,12 @@ public class OnboardingProcessEntity implements Serializable {
     @NotNull
     private OnboardingProcessConfigurationEntity processConfiguration;
 
+    /**
+     * Timestamp when the identity data (e.g. selfie, uploaded documents) was cleaned up.
+     */
+    @Column(name = "timestamp_personal_data_cleaned")
+    private Date timestampPersonalDataCleaned;
```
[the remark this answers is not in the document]
michal-rozehnal-w: I updated the PR description

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260408-personal-data-cleaned-timestamp.xml
```diff
@@ -0,0 +1,78 @@
...
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="5" author="Michal Rozehnal">
+        <preConditions>
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

docs/onboarding/Configuration-Properties.md
michal-rozehnal-w: Reformated

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -18,3 +18,44 @@ Added a new indexed column `subject_id` holding an identifier linking the audit
...
+
+
+### Add column timestamp_personal_data_cleaned to es_onboarding_process table
+
+The `timestamp_personal_data_cleaned` column track when the personal data associated with an onboarding process has been cleaned up.
+
+
+### Add column document_verification_id to es_document_data table
+
+The `document_verification_id` column is used to link record with the `es_document_verification`. There is a one-to-one relationship and there is foreign key constraint.
```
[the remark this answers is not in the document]
michal-rozehnal-w: Applied

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -18,3 +18,44 @@ Added a new indexed column `subject_id` holding an identifier linking the audit
...
+
+
+### Onboarding process identity data retention
+
+The identity data retention period is configured using the property `enrollment-server-onboarding.identity-verification.data-retention`, and the retention period 
+is measured from the process completion time—either the `timestamp_finished` or `timestamp_failed column` in the `es_onboarding_process` table.
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
```diff
@@ -196,4 +197,37 @@ default List<String> findActiveByTimestampWithLock(Date dateCreatedBefore) {
...
+     * @param processCompletedBefore time before which the process must have been completed to be included in the result
+     * @return list of identity data IDs for completed processes to be cleaned up
+     */
+    @Query("""
+        SELECT p.id AS processId,
+               iv.id AS identityVerificationId,
+               dv.id AS documentVerificationId
+        FROM OnboardingProcessEntity p
+        JOIN IdentityVerificationEntity iv ON iv.processId = p.id
+        JOIN DocumentVerificationEntity dv ON dv.identityVerification.id = iv.id
```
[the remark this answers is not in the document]
michal-rozehnal-w: Interesting question. I think there is always at least one document (a DocumentVerificationEntity always exists). However, to be more resilient, I added a `LEFT JOIN`, filtering null values in cleaning task and integration test. This shouldn't have any negative impact on the query performance

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
michal-rozehnal-w: This is a valid point. I am considering the following solutions:

1. Add a limit to the query (configurable via application properties), so each task run processes only a small portion of the large legacy dataset.
2. Before the upgrade, add the `timestamp_personal_data_cleaned` column manually—it is fine for Liquibase, as it will skip creating the column if it already exists. Then set it manually to the current time, excluding the most recent processes which haven't been cleaned yet by the old cleaning service. For example:

```
UPDATE es_onboarding_process
SET timestamp_personal_data_cleaned = now()
WHERE timestamp_personal_data_cleaned IS NULL
AND (timestamp_failed < :old_retention_time OR timestamp_finished < :old_retention_time)
```

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningTask.java
zcgandcomp: The expected retention period is days. There could be some extra deleting after FAILED attempts, but in standard case the retentions will be something like 30 or 90 days. IMHO
The period is more about how many records are we expect to delete. Because of the construction of the delete this is mostly sensitive to the initial select performance. And the second factor is some recovery - i.e. we are going to delete everything. There is a batching, but the delete job is trying to delete ALL records. We should actually delete only batch and end the transaction, i.e., the job. 
And the interval is our expectation how long the 1000 processes can be cleaned plus some safe buffer.
IMHO something 30s-1minute is fine. 15 seconds are fine in normal conditions but in case of trouble this can start clashing.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -118,44 +118,59 @@ public DocumentsSubmitResult checkDocumentUpload(OwnerId ownerId, DocumentVerifi
...
 
-        logger.info("action: submitDocuments, state: succeeded, provider: microblink, rejectReason: {}", result.getRejectReason());
-        return result;
+            logger.info("action: submitDocuments, state: succeeded, provider: microblink, rejectReason: {}", result.getRejectReason());
+            return result;
+        } catch (final DocumentVerificationException | RemoteCommunicationException | RuntimeException e) {
+            final var errorMessage = e.getMessage();
+            final var documentResults = documentsVerificationData.stream()
+                .map(document -> {
+                    final var submitResult = new DocumentSubmitResult();
```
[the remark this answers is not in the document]
michal-rozehnal-w: Very good catch. It is unlikely that an exception without a message will be thrown, but it can still happen. I adjusted the message so it is never null. Also added unit test.

enrollment-server-onboarding-provider-microblink/src/test/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProviderIntTest.java
```diff
@@ -580,6 +580,7 @@ private void assertIdCardDocumentsData(final DocumentsSubmitResult result) {
         final var frontDocumentData = documentDataByUploadId.get(frontDocumentUploadId);
         assertArrayEquals(idCardFrontImage.getData(), frontDocumentData.getData());
         assertEquals(new Date().getTime(), frontDocumentData.getTimestampCreated().getTime(), TIMESTAMP_ASSERT_DELTA_MS);
+        assertEquals(ID_CARD_FRONT_DOCUMENT_ID, frontDocumentData.getDocumentVerificationId());
```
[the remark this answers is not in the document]
michal-rozehnal-w: True, fixed by setting the property in these methods

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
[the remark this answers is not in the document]
zcgandcomp: I actually reviewed this logic. If no document verification exist it would mean the process didn't reach the document scanning. As a result of your change, you will now start selecting identity verification without documents and feed them to the deletes. But there will be nothing to delete...

So the impact on query performance is very small (but actualyl there is), but you are impacting further actions in the cleaning flow.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/configuration/OnboardingConfig.java
michal-rozehnal-w: Used `enrollment-server-onboarding.identity-verification.data-retention`

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
zcgandcomp: Create the column with a default value, and then drop the default value...

Or use different solution for the deletion...

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningTask.java
michal-rozehnal-w: For activations cleanup we have `@Scheduled(fixedDelayString = "PT60S", initialDelayString = "PT30S")`, I can use the same values for the personal data cleanup

zcgandcomp (changes requested):
Few remarks - the initial load, mainly from marking the clean and let's make job in tranches, not all remaining documents.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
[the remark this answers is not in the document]
michal-rozehnal-w: Correct, I removed the `LEFT JOIN` and put back just `JOIN`

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
michal-rozehnal-w: I like the approach - setting a default value and then dropping it so new values are `NULL`. However, wouldn’t this still create a time window during which some processes could be finished but their personal data not yet deleted?
I added a query to the migration guide to set the timestamp for processes that were already cleaned by the old task, so they are not picked up by the new one. This requires manual execution, but ensures that no personal data is left behind.
Let me know if you see any issues with this approach.

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -27,23 +27,51 @@ The `timestamp_personal_data_cleaned` column track when the personal data associ
...
-and the cleanup will work as described.
+
+#### Reduce the load of the first run of the new cleanup logic
+
+The new cleanup logic uses a new column, `es_onboarding_process.timestamp_personal_data_cleaned`, to track whether personal data has been deleted for a process. 
+Because the default value is `NULL`, all processes would be picked up by the first run of cleanup task. 
+To reduce the load on the task, it is recommended to set this value for processes that have already been cleaned by the old cleanup logic.
+Use the following SQL query to do so:
+
+```sql
```
banterCZ, line 52: May we have Oracle and Postgres variants, please?

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
banterCZ: I like the migration script, but the way with the default value seems more robust to me (if somebody does not read the migration guide properly).

banterCZ (approved):
Migration steps are improved, and the scheduled job processes the items in batches (not all records at once).

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
michal-rozehnal-w: Sure, added

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
michal-rozehnal-w: I agree that the automatic solution in the Liquibase script is probably the best. On the other hand, we will still need some manual action from the user, since we don’t know the value of `enrollment-server-onboarding.onboarding-process.expiration`. That’s why I initially leaned toward this solution.

However, your argument is valid. If the migration is not read and the query for setting the timestamp is not executed, it could cause a very high load on the cleaning task.

If we set a default value and then drop it, we still have to deal with a few legacy records that may not have been cleaned by the old cleanup job. Because they would already have a timestamp, they would not be processed by the new job either. The advantage, however, is that the new cleaning job would be spared the load.

I would rework the approach as follows:

1. Liquibase script: Add the column with a default value of now().
2. Liquibase script: Drop the default value for the column.
3. Manual execution: Set the timestamp to `NULL` for these “ghost” records so they are cleaned by the new cleanup task, using the following query:

```
UPDATE es_onboarding_process
SET timestamp_personal_data_cleaned = NULL
WHERE timestamp_created > (timestamp_of_upgrade - (2 * enrollment-server-onboarding.onboarding-process.expiration + 10 minutes))
  AND timestamp_created <= timestamp_of_upgrade;
```

The `timestamp_of_upgrade` can be obtained from the database at any time, as it will be the minimum value in the timestamp column.

@banterCZ does this solution look good to you?

banterCZ: I let the decision to @zcgandcomp

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -18,3 +18,92 @@ Added a new indexed column `subject_id` holding an identifier linking the audit
...
+Because the default value is `NULL`, all processes would be picked up by the first run of cleanup task. 
+To reduce the load on the task, it is recommended to set this value for processes that have already been cleaned by the old cleanup logic.
+Use the following SQL query to do so:
+
+PostgreSQL:
+
+```sql
+ALTER TABLE es_document_data ADD document_verification_id VARCHAR(36);
+
+UPDATE es_onboarding_process
```
zcgandcomp, line 55: This update needs to search ALL records in the table. 
Create the column `timestamp_personal_data_cleaned` with a default value now() / SYSDATE.
Alter the column `timestamp_personal_data_cleaned` DEFAULT NULL

Mention scripts to remove the remnants - we can delete the records in the tables that are subject to cleaning based on the retention period. (i.e. old behavior) 
i.e., with created_timestamp lower then current date - retention period.

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -18,3 +18,92 @@ Added a new indexed column `subject_id` holding an identifier linking the audit
...
+SET timestamp_personal_data_cleaned = now()
+WHERE timestamp_created < now() - :cleaned_processes_timestamp;
+```
+
+Oracle:
+
+```sql
+ALTER TABLE es_document_data ADD document_verification_id VARCHAR2(36);
```
zcgandcomp, line 63: Why is the ALTER mentioned here? This should be done via Liquibase.

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -18,3 +18,92 @@ Added a new indexed column `subject_id` holding an identifier linking the audit
...
+For manual cleanup, check whether any such records exist by executing the following SQL query:
+
+```sql
+SELECT count(*)
+FROM es_document_data
+WHERE document_verification_id IS NULL;
+```
+
+If any records exist, they can be deleted using the following SQL query:
+
```
zcgandcomp, line 99: This will effectively delete live data.  See my previous comment, delete the data where document_verification_id is null, and created timestamp before current date - retention period.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/OnboardingProcessRepository.java
zcgandcomp: This could also work, but the computation which records should have clean timestamp cleared is tricky. Better to delete old records without such computation.
And the update of the table based on a non-indexed column will result in slow updates on large databases.

zcgandcomp (changes requested):
The migration approach still requires a few optimizations.

banterCZ:
Please describe this in the Changelog.

PR #1692, opened 2026-04-10
Fix #1691: Set ErrorDetail in final check

### LINKED ISSUE #1691: Set ErrorDetail in final check
### Task specification

Right now, when `identityVerificationPrecompleteCheck` fails, the error detail is set to `DOCUMENT_VERIFICATION_FAILED`. Replace it with `result.getErrorDetail()`.

banterCZ:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -281,7 +281,7 @@ public void processVerificationResult(final OwnerId ownerId, final IdentityVerif
             moveToPhaseAndStatus(idVerification, IdentityVerificationPhase.COMPLETED, ACCEPTED, ownerId);
         } else {
             logger.warn("Final validation did not pass, marking identity verification as failed due to '{}', {}", result.getErrorDetail(), ownerId);
-            idVerification.setErrorDetail(ErrorDetail.DOCUMENT_VERIFICATION_FAILED);
+            idVerification.setErrorDetail(result.getErrorDetail());
```
michal-rozehnal-w, line 284: Just want to confirm - we don’t expose this information in the REST API; this is for internal BE usage, right? The `POST /api/identity/status` endpoint contains `rejectReason`, but that’s a different field not touched in this PR, so there isn’t any functional change from my point of view.

michal-rozehnal-w (approved):
Looks good

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
banterCZ: 👏 Well spotted! 🤔 `DOCUMENT_VERIFICATION_FAILED` is definitely misleading. We should store at least a different constant. It could be handy to keep error details in the database for easier investigation. ~Unfortunately, we return error details in the status, as you pointed out.~ `IdentityVerificationPrecompleteCheck` is the very last step, working mainly as a sanity check. Right now, filling values below. @michal-rozehnal-w, please review again.

```java
    private static final String SOME_DOCUMENTS_NOT_ACCEPTED = "Some documents not accepted";
    private static final String REQUIRED_DOCUMENTS_NOT_PRESENT = "Required documents not present";
    private static final String NOT_VALID_PHASE_AND_STATE = "Not valid phase and state";
    private static final String NOT_VALID_USER_VERIFICATION_OTP = "Not valid user verification OTP";
    private static final String NOT_VALID_ACTIVATION_OTP = "Not valid activation OTP";
    private static final String NOT_VALID_ACTIVATION = "Activation is not valid";
    private static final String NOT_VALID_SCA = "Did not pass SCA";
    private static final String NOT_VALID_TARGET_ACTIVATION = "Target activation is not valid";
```

banterCZ: As we have discussed offline, we do not return `error_detail` at REST.

michal-rozehnal-w: After the internal discussion, the change is fine. We don’t expose `error_detail`; only `reject_reason`, which is not affected by the PR. Therefore, this change is safe. Thank you for double checking this.

tech-kratas (approved):
looks good

PR #1726, opened 2026-04-20
Fix #1722: Implicit validation of encryptionContext and PowerAuthApiAuthentication

### LINKED ISSUE #1722: Implicit validation of encryptionContext and PowerAuthApiAuthentication
### Task specification

In controllers, there are too many explicit validations of `PowerAuthApiAuthentication` and `encryptionContext`. Let's use bean validation.

A follow-up to #1589

```java
if (encryptionContext == null) {
    throw new PowerAuthEncryptionException("ECIES decryption failed during cleanup");
}

if (apiAuthentication == null) {
    throw new PowerAuthTokenInvalidException("Unable to verify device registration when " + description);
}
```

banterCZ:
Should be merged after release 2.1

Just cleanup, because unfortunately, `io.swagger.v3.oas.annotations.Parameter#required` is only for documentation purposes. `@NotNull` is not enough, since `MobileTokenController` throws `MobileTokenAuthException` and returns 401.


michal-rozehnal-w (approved):
Looks good

PR #1731, opened 2026-04-23
Fix #1730: Process failing for specific config in ESO

### LINKED ISSUE #1730: Process failing for specific config in ESO
### Task specification

Process will always fail for this specific configuration.

```
{
  "enabled": true,
  "activationType": "CODE",
  "otpForIdentification": true,
  "otpForIdentityVerification": true,
  "useTemporaryActivation": false,
  "approvalEnabled": true,
  "verifyPresenceWithOtp": false,
  "consentRequired": true,
  "documents": {
    "totalRequiredDocumentsCount": 2,
    "groups": [
      {
        "requiredDocumentsCount": 2,
        "items": [
          {
            "type": "ID_CARD",
            "sideCount": 2,
            "country": "CZE"
          },
          {
            "type": "DRIVING_LICENSE",
            "sideCount": 1,
            "country": "CZE"
          }
        ]
      }
    ]
  }
}
```

The process fails with the following message in the log:
`Final validation did not pass, marking identity verification as failed due to ‘Not valid phase and state’, …`

This is caused by an incorrect order of conditions in [this method](https://github.com/wultra/enrollment-server/blob/develop/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationPrecompleteCheck.java#L231). If both OTP and approval are enabled, the OTP is executed before approval. However, in the pre-complete check, the `if` branch for OTP comes first, which triggers an error about an invalid state and phase.

This bug can be reproduce by using following combination of process configuration:
```
"useTemporaryActivation": false,
"approvalEnabled": true,
"otpForIdentityVerification": true
```

michal-rozehnal-w:


banterCZ (approved):
Thanks for the fix. Let's decide with @zcgandcomp whether to include it in release 2.1

PR #1735, opened 2026-04-23
Fix #1734: Document extracted data are not cleaned by cleaning task 

### LINKED ISSUE #1734: Document extracted data are not cleaned by cleaning task

_No response_

### Environment

_No response_

For successfully completed processes, the following database data are not automatically cleaned by the cleaning task:

- `es_document_result.verification_result`
- `es_document_result.extracted_data`

These data must always be cleaned after the [retention period](https://github.com/wultra/enrollment-server/blob/develop/enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java#L202), regardless of the process result.

### Log output

_No response_

michal-rozehnal-w:


enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/DocumentResultRepository.java
```diff
@@ -54,4 +60,19 @@ public interface DocumentResultRepository extends CrudRepository<DocumentResultE
...
+     * @return Number of records cleaned.
+     */
+    @Modifying
+    @Query("""
+            UPDATE DocumentResultEntity d
+            SET d.verificationResult = null,
+                d.extractedData = null
+            WHERE (d.verificationResult IS NOT NULL OR d.extractedData IS NOT NULL)
+            AND d.timestampCreated < :dateCleanup
+    """)
```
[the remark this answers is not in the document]
michal-rozehnal-w: Added the index

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -186,4 +186,6 @@ Previously, the data was deleted immediately after the retention period, which c
 Records from the following tables are deleted according to this calculation:
 - `es_document_data`
 - `es_processed_document_data`
-- `es_selfie`
\ No newline at end of file
+- `es_selfie`
+
+For records in table `es_document_result` values in columns `verification_result` and `extracted_data` are set to `null`after the retention period.
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -162,6 +162,16 @@ public int cleanupProcessedDocumentData() {
         return processedDocumentDataRepository.cleanup(getPersonalDataRetentionTime());
     }
 
+    /**
+     * Clean personal data from document results.
+     *
+     * @return Number of deleted personal data.
```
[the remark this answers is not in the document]
michal-rozehnal-w: Updated

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/DocumentResultRepository.java
michal-rozehnal-w, line 60: Query for cleaning by `documentVerificationIds` is still needed for manual clean by REST API

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/DocumentResultRepository.java
```diff
@@ -54,4 +60,19 @@ public interface DocumentResultRepository extends CrudRepository<DocumentResultE
...
+     * @param dateCleanup Date older than which the records should be cleaned.
+     * @return Number of records cleaned.
+     */
+    @Modifying
+    @Query("""
+            UPDATE DocumentResultEntity d
+            SET d.verificationResult = null,
+                d.extractedData = null
+            WHERE (d.verificationResult IS NOT NULL OR d.extractedData IS NOT NULL)
+            AND d.timestampCreated < :dateCleanup
```
michal-rozehnal-w, line 77: Here I am not 100 % sure about performance. I added index on the `timestamp_created` but I have concern about the condition with `OR`. Since we are in a release, I don't want to make many changes. We also plan to adjust the cleanup logic for all personal data after the release. In any case, we must not introduce any performance issues here.
Any thoughts about the query?

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260423-document-result-timestamp-created-index.xml
```diff
@@ -0,0 +1,18 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.33.xsd">
+
+    <changeSet id="1" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="es_document_result_timestamp_created_idx" tableName="es_document_result" />
+            </not>
+        </preConditions>
+        <comment>Add index on timestamp_created column in es_document_result table</comment>
+        <createIndex tableName="es_document_result" indexName="es_document_result_timestamp_created_idx">
```
banterCZ: Missing in the SQL scripts.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/DocumentResultRepository.java
banterCZ: I let the performance point of view to @zcgandcomp

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningTask.java
```diff
@@ -112,6 +112,18 @@ public void cleanupDocumentData() {
         logger.debug("action: cleanupDocumentData, state: succeeded, cleanedRecords: {}", count);
     }
 
+    /**
+     * Clean personal data from document results older than retention time.
+     */
+    @Scheduled(fixedDelayString = "PT10M", initialDelayString = "PT15S")
```
banterCZ: No configuration property?

banterCZ (approved):
Ok, sets `verificationResult` and `extractedData` to `null`. Not sure whether we need a configuration property for the job.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260423-document-result-timestamp-created-index.xml
michal-rozehnal-w: I always forget to add them—fixed.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningTask.java
michal-rozehnal-w: I wanted to be consistent with the rest of the methods here, since we don’t use any property yet. I can add it if we really want it—just let me know.

zcgandcomp (changes requested):
- introduce new **indexed** NULL column `anonymized` boolean (feel free to give it a better name)
- for new records use value false. (the old records should be kept null)
- - bonus solution create the column with default true and change the default to false
- change the cleaning update to:

```
            UPDATE DocumentResultEntity d
            SET d.verificationResult = null,
                d.extractedData = null
                d.anonymized = true
            WHERE d.anonymized = false
               AND d.timestampCreated < :dateCleanup

```
- As we are creating the column with safe default and we have only 1 SQL statement this doesn't needs to be limited, we can set some very large limit e.g. 10000

docs/onboarding/Configuration-Properties.md
```diff
@@ -44,12 +44,12 @@ See [Configuration of Onboarding Process](./Configuration-Onboarding-Process.md)
...
+| `enrollment-server-onboarding.identity-verification.max-failed-attempts`                 | `5`      | Maximum failed attempts for identity verification.                                                                                                                                                                                                              |
+| `enrollment-server-onboarding.identity-verification.max-failed-attempts-document-upload` | `5`      | Maximum failed attempts for document upload.                                                                                                                                                                                                                    |
+| `enrollment-server-onboarding.identity-verification.data-retention`                      |          | Retention duration for identity verification personal data. The effective retention window is computed as: `timestamp_created` + `enrollment-server-onboarding.onboarding-process.expiration` + this value. An empty value means automatic cleanup is disabled. |
```
michal-rozehnal-w: I set it to `null` — so the cleaning task is disabled by default for compatibility. The ZenID provider already uses `es_document_result`.`extracted_data` and `verification_result` (by propagating them from [provider](https://github.com/wultra/enrollment-server/blob/develop/enrollment-server-onboarding-provider-zenid/src/main/java/com/wultra/app/onboardingserver/provider/zenid/ZenidDocumentVerificationProvider.java#L380) in result), and these were not automatically deleted in the previous version.

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.1.0.md
```diff
@@ -180,10 +180,13 @@ enrollment-server-onboarding.document-verification.checkDocumentSubmitVerificati
 
 ## Cleaning task
 
-Modified the calculation of the retention period for processing personal data (e.g., uploaded documents and selfie photos) as follows. The data are deleted after the expiration time of the process plus the retention period.
-Previously, the data was deleted immediately after the retention period, which could lead to the deletion of data for active processes if the process expiration is higher than the data retention.
+Automatic cleaning of personal data (e.g., uploaded documents and selfie photos) can be enabled by setting the property `enrollment-server-onboarding.identity-verification.data-retention`.
```
michal-rozehnal-w: I adjusted the description because the tables `es_processed_document_data` and `es_selfie` were created after version 1.10. The `es_document_data` table already existed but was never used in production; we are now starting to use it and have changed its original purpose (it was intended as temporary storage in case document data were uploaded before calling the upload endpoint).
The `es_document_result.extracted_data` and `verification_result` columns were not cleaned in the previous version, so this behavior is new and nothing has been changed retroactively.

Because of this, there is no need to describe any manual cleanup for historical data, as there are none.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/DocumentResultRepository.java
michal-rozehnal-w: The query was changed

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningTask.java
```diff
@@ -88,42 +88,6 @@ public void terminateExpiredProcesses() {
         cleaningService.terminateExpiredProcesses();
     }
 
-    /**
```
michal-rozehnal-w, line 91: Methods related to personal data were moved to a separate class so they can be controlled by the data retention property.

michal-rozehnal-w:
I did local testing:

- Migration: I created a few records in the database (via a BE test), then applied Liquibase and created another set of new records. The old records have anonymized = true, while the new ones have false.
- Cleaning task disabled: I set the process expiration to 1 minute and left the retention time unspecified. Personal data were not cleaned.
- Cleaning task enabled: I set both the process expiration and retention time to 1 minute. The data were cleaned after 2 minutes.

michal-rozehnal-w:
> * introduce new **indexed** NULL column `anonymized` boolean (feel free to give it a better name)
> * for new records use value false. (the old records should be kept null)
> * * bonus solution create the column with default true and change the default to false
> * change the cleaning update to:
> 
> ```
>             UPDATE DocumentResultEntity d
>             SET d.verificationResult = null,
>                 d.extractedData = null
>                 d.anonymized = true
>             WHERE d.anonymized = false
>                AND d.timestampCreated < :dateCleanup
> ```
> 
> * As we are creating the column with safe default and we have only 1 SQL statement this doesn't needs to be limited, we can set some very large limit e.g. 10000

- Added new `anonymized` column
- Added default value is `true`, so old records are marked as cleaned, then default value is changed to `false`
- Updated the query

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260423-document-result-timestamp-created-index.xml
```diff
@@ -0,0 +1,52 @@
...
```
zcgandcomp: do not add this index.

docs/onboarding/Configuration-Properties.md
zcgandcomp: but the default value is not null, the default value was 1h, and the default behavior was to clean the data...

zcgandcomp (changes requested):
- the index es_document_result_timestamp_created_idx is not needed and should not be used. Add an index only if you have a select that actually requires such an index.
- The default value of retention period should stay the same as in previous release. IMHO. But yes if the data retention is not used at all, we can disable them, but each deployment has to be instructed to turn this on. As shoud be our default behavior. My preference is to keep the default on.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260423-document-result-timestamp-created-index.xml
michal-rozehnal-w: Removed

docs/onboarding/Configuration-Properties.md
michal-rozehnal-w: I was considering this from a compatibility perspective, because in the previous version this property controlled only the `es_document_data` table, which was effectively never used. In this version, we are using it to clean additional data.
That said, you are right that the default value was historically 1 hour. I have set it back and tested it locally, and confirmed that setting `enrollment-server-onboarding.identity-verification.data-retention=` disables the personal data cleaning task.

michal-rozehnal-w:
> * the index es_document_result_timestamp_created_idx is not needed and should not be used. Add an index only if you have a select that actually requires such an index.
> * The default value of retention period should stay the same as in previous release. IMHO. But yes if the data retention is not used at all, we can disable them, but each deployment has to be instructed to turn this on. As shoud be our default behavior. My preference is to keep the default on.

- The index was removed from migration and retested locally, that new changeset works
- Set default value to 1h and retested, confirmed that the property with empty value disabled the cleaning task

zcgandcomp (approved):
Looks OK to me.
Functional change.
Migration guides and documentation updated.
Change is compatible and should work well on current installations

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/configuration/IdentityVerificationConfig.java
```diff
@@ -51,7 +52,7 @@ public class IdentityVerificationConfig {
     private boolean presenceCheckCleanupEnabled;
 
     @Value("${enrollment-server-onboarding.identity-verification.data-retention:1h}")
-    private Duration dataRetentionTime;
+    private Optional<Duration> dataRetentionTime;
```
banterCZ: From a conceptual point of view, `Optional` is primarily intended for use as a method return type. We should keep the original type and use it in the getter.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
```diff
@@ -201,7 +211,8 @@ public void terminateExpiredIdentityVerifications() {
 
     private Date getPersonalDataRetentionTime() {
         final var processExpirationTime = onboardingConfig.getProcessExpirationTime();
-        final var dataRetentionTime = identityVerificationConfig.getDataRetentionTime();
+        final var dataRetentionTime = identityVerificationConfig.getDataRetentionTime()
```
banterCZ, line 214: In the documentation, it is said:
> An empty value means automatic cleanup is disabled.

But here it throws an exception.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/PersonalDataCleaningTask.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+/**
+ * Task to clean personal data stored during an onboarding process.
+ */
+@Component
+@Slf4j
+@AllArgsConstructor
+@ConditionalOnPropertyNotEmpty("enrollment-server-onboarding.identity-verification.data-retention")
```
banterCZ, line 35: Ok, I see, it should never throw an exception, thanks to this conditional configuration, right?

banterCZ (approved):

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/PersonalDataCleaningTask.java
michal-rozehnal-w: Yes, `the identityVerificationConfig.getDataRetentionTime()` property is used only within the scope of the `PersonalDataCleaningTask` class, which is a conditional bean created only if the property exists. The exception serves as a safety fallback and should never occur.
We can refactor the cleaning service and property classes later. Currently, everything resides in `CleaningService`, but it is triggered by two different tasks, one of which is conditional. On the other hand, I didn’t want to introduce a larger refactoring in this PR.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/task/cleaning/CleaningService.java
michal-rozehnal-w: Explained in this [comment](https://github.com/wultra/enrollment-server/pull/1735/changes#r3135536146)

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/configuration/IdentityVerificationConfig.java
michal-rozehnal-w: Good idea, I used property with `Duration` type, but added getter with `Optional`

PR #1741, opened 2026-04-24
Fix #1725: Publish event after onboarding process success is commited

### LINKED ISSUE #1725: Orchestrate calling User Date Store
### Task specification

If process is successful and integration with User Data Store is enabled then we should call it.
Scope of this issue:

- Add application property for enabling the integration
- Implement everything else for successful flow - calling the REST API - except of https://github.com/wultra/enrollment-server/issues/1724 and https://github.com/wultra/enrollment-server/issues/1723

michal-rozehnal-w:


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -215,6 +222,25 @@ private Mono<Void> persistState(final StateContext<OnboardingState, OnboardingEv
...
+
+    private Function<StateContext<OnboardingState, OnboardingEvent>, Mono<Void>> publishCompletedAcceptedEvent() {
+        return context -> Mono.fromRunnable(() -> {
+            final OwnerId ownerId = (OwnerId) context.getMessageHeader(EventHeaderName.OWNER_ID);
+
+            final String processId = Optional.of(context)
+                    .map(StateContext::getExtendedState)
+                    .map(extendedState -> extendedState.get(ExtendedStateVariable.IDENTITY_VERIFICATION, IdentityVerificationEntity.class))
+                    .map(IdentityVerificationEntity::getProcessId)
+                    .orElseThrow(() -> new IllegalStateException("Process ID not found in extended state for userId: %s, activationId: %s".formatted(ownerId.getUserId(), ownerId.getActivationId())));
```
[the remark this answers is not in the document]
michal-rozehnal-w: Good catch, implemented

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedListener.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
+    @Async
+    public void onOnboardingCompletedAccepted(final OnboardingCompletedAcceptedEvent event) {
+        logger.info("",
+                kv("action", "onOnboardingCompletedAccepted"),
+                kv("state", "initiated"),
+                kv("processId", event.getProcessId()),
+                kv("userId", event.getOwnerId().getUserId()),
+                kv("activationId", event.getOwnerId().getActivationId()));
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is a new way how to use structured logging

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedListener.java
```diff
@@ -0,0 +1,58 @@
...
+
+        // TODO (michal-rozehnal-w, 2026-04-24, #1723) Sending is implemented in separate issue
+
+        logger.info("",
+                kv("action", "onOnboardingCompletedAccepted"),
+                kv("state", "succeeded"));
```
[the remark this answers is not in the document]
michal-rozehnal-w: I think it is fine, there should be `succeeded` in final implementation and there is TODO comment

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedEvent.java
```diff
@@ -0,0 +1,47 @@
...
+     * Creates a new event.
+     *
+     * @param source    the object on which the event initially occurred
+     * @param ownerId   identification of the activation owner
+     * @param processId onboarding process ID
+     */
+    public OnboardingCompletedAcceptedEvent(final Object source, final OwnerId ownerId, final String processId) {
+        super(source);
+        this.ownerId = ownerId;
+        this.processId = processId;
```
banterCZ, line 45: 🤔 In general, the event should carry the whole state without the need to query any detail, because between firing the event and processing it, the data may change. But for the finished process, it could be fine to collect data in the service, as we have been doing in other PRs.

banterCZ (approved):
Looks good.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedEvent.java
michal-rozehnal-w: Good point. In my experience, the choice between having an event with full details versus keeping it lightweight is never fully resolved :) - both approaches have their pros and cons. For this particular event, I agree it doesn’t matter much, since once the process is finished, it reaches a terminal state. The only potential change might be, for example, in the automatic cleaning of personal data.
It shouldn’t be a complicated change to call the service in `StateMachineConfig.registerPublishEventFunction`, put all the data into the event, and then, in `OnboardingCompletedAcceptedListener`, simply call the client in a for loop. However, I don’t see much benefit in doing that, since event is triggered right after commit.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/userdatastore/NoOpUserDataStoreService.java
```diff
@@ -29,8 +27,6 @@
  *
  * @author Lubos Racansky, lubos.racansky@wultra.com
  */
-@ConditionalOnMissingBean(UserDataStoreService.class)
-@Service
```
michal-rozehnal-w, line 33: I used `UserDataStoreService` as a dependency in an event listener, and it caused the combination of `@ConditionalOnMissingBean` and the `@Service` not to be reliably resolved - application doesn't start. I fixed it by adding a NoOp configuration class.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedListener.java
```diff
@@ -0,0 +1,66 @@
...
+                event.getOwnerId().getActivationId());
+
+        try {
+            final var documentData = userDataStoreService.collectDocumentData(processId);
+            userDataStoreService.storeDocumentData(documentData);
+
+            logger.info("action: onOnboardingCompletedAccepted, state: succeeded, storedDocumentCount: {}", documentData.size());
+        } catch (final UserDataStoreClientException | RuntimeException e) {
+            logger.warn("action: onOnboardingCompletedAccepted, state: failed", e);
+            logger.info("action: onOnboardingCompletedAccepted, state: failed, errorMessage: {}", e.getMessage());
```
banterCZ: Why do we have both levels? Isn't WARN enough?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/userdatastore/UserDataStoreConfig.java
```diff
@@ -37,31 +39,46 @@
...
-    @Bean
-    public UserDataStoreClient userDataStoreClient(final UserDataStoreConfigProperties config) throws UserDataStoreClientException {
-        final var clientConfig = config.getRestClientConfig();
+    /**
+     * Configuration when User Data Store is enabled.
+     */
+    @Configuration
+    @ConditionalOnProperty(name = "enrollment-server-onboarding.user-data-store.enabled", havingValue = "true")
+    @EnableConfigurationProperties(UserDataStoreConfigProperties.class)
+    static class EnabledConfig {
```
michal-rozehnal-w, line 52: There is an inner class, so the properties are bound only if the user data store integration is enabled. We don’t want to bind them when it is disabled, because there is validation on `restClientConfig` and it would fail.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedListener.java
michal-rozehnal-w: Right, originally I wanted to keep the INFO level tiny, but with WARN it doesn't make much sense. I removed the INFO one

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/userdatastore/UserDataStoreConfig.java
```diff
@@ -37,31 +39,46 @@
...
+        public UserDataStoreClient userDataStoreClient(final UserDataStoreConfigProperties config) throws UserDataStoreClientException {
+            final var clientConfig = config.getRestClientConfig();
+
+            logger.info("Registering UserDataStore client with url: {}", clientConfig.getBaseUrl());
+            return new UserDataStoreRestClient(clientConfig);
+        }
 
-        logger.info("Registering UserDataStore client with url: {}", clientConfig.getBaseUrl());
-        return new UserDataStoreRestClient(clientConfig);
+        @ConditionalOnBean(UserDataStoreClient.class)
```
banterCZ: Is it still required now?

enrollment-server-onboarding/src/main/resources/application.properties
```diff
@@ -221,7 +221,6 @@ enrollment-server-onboarding.user-data-store.enabled=false
 enrollment-server-onboarding.user-data-store.document-type=ALL
 enrollment-server-onboarding.user-data-store.store-extracted-data=true
 enrollment-server-onboarding.user-data-store.store-document-image-scan=true
-enrollment-server-onboarding.user-data-store.restClientConfig.baseUrl=
```
banterCZ, line 224: 🤔 Other configuration sections using properties with URL as a kind of documentation. Could we comment that out instead of deleting?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/userdatastore/UserDataStoreConfig.java
michal-rozehnal-w: Good catch, since the inner class is conditional, the annotation is no longer needed. Removed

enrollment-server-onboarding/src/main/resources/application.properties
michal-rozehnal-w: Good idea, done

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedListener.java
```diff
@@ -0,0 +1,65 @@
...
+        logger.info("action: onOnboardingCompletedAccepted, state: initiated, processId: {}, userId: {}, activationId: {}",
+                processId,
+                event.getOwnerId().getUserId(),
+                event.getOwnerId().getActivationId());
+
+        try {
+            final var documentData = userDataStoreService.collectDocumentData(processId);
+            userDataStoreService.storeDocumentData(documentData);
+
+            logger.info("action: onOnboardingCompletedAccepted, state: succeeded, storedDocumentCount: {}", documentData.size());
```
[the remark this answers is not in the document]
michal-rozehnal-w: This would duplicate these information as they are in init action log and we can link them via traceId

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/event/OnboardingCompletedAcceptedListener.java
```diff
@@ -0,0 +1,65 @@
...
+        } catch (final UserDataStoreClientException | RuntimeException e) {
+            logger.warn("action: onOnboardingCompletedAccepted, state: failed", e);
```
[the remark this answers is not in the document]
michal-rozehnal-w: This would duplicate these information as they are in init action log and we can link them via traceId

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -63,6 +65,7 @@
 import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Optional;
```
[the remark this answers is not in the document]
michal-rozehnal-w: Removed

PR #1750, opened 2026-04-29
Fix #1619: Refactor ProcessIdentifierGuard

### LINKED ISSUE #1619: Refactor ProcessIdentifierGuard
### Task specification

We should reconsider `ProcessIdentifierGuard`. It does not make much sense to do such verification for each state machine step, but just at the start of the machine. You may see it as a code smell in `createCompositeGuard(...)`.

michal-rozehnal-w:
# Functional changes

- The method responsible for fetching processes whose next event should be sent by the scheduled task no longer uses a read-only transaction. Instead of processing all records using a parallel stream, records are now fetched from the database and processed in batches (without parallelization), with a default batch size of 10 000.
- `ProcessIdentifierGuard` has been removed. Process existence validation is now performed only for processes triggered by the scheduled task that sends the next event.


enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
michal-rozehnal-w: This method can be placed in a different class to allow using an annotation instead of `transactionTemplate`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -163,4 +170,10 @@ private StateMachine<OnboardingState, OnboardingEvent> fetchStateMachine(
         return prepareStateMachine(processId, onboardingState, identityVerification);
     }
 
+    private void lockAndVerifyProcess(final String processId) throws OnboardingProcessException {
```
michal-rozehnal-w, line 169: Generally, there are two sources of events for a state machine:

The first is the REST API, where checking the process—requested vs. actual ID, user ID, and activation ID—makes sense, since the processId comes in the request body from mobile. We also perform this check with a process lock in all endpoints.

The second is the `changeMachineStatesInBatch` method (also start of transaction), which is called via scheduled tasks. This is purely backend, with no user input. In this case, I don’t think the check is necessary (once the process and identity records are created, these IDs should be never changed), so I left only the lock and onboarding status verification. However, I still need to double-check the status as well to ensure there are no changes in behavior. The guard did the check on specific transitions for specific status.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -253,7 +251,7 @@ private void configureInitialTransition(StateMachineTransitionConfigurer<Onboard
                 .withExternal()
                 .source(OnboardingState.INITIAL)
                 .event(OnboardingEvent.IDENTITY_VERIFICATION_INIT)
-                .guard(createCompositeGuard(processIdentifierGuard, consentResolvedGuard))
+                .guard(consentResolvedGuard)
```
banterCZ, line 277: I would prefer to merge this after #1592

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -118,7 +117,6 @@ public class StateMachineConfig extends EnumStateMachineConfigurerAdapter<Onboar
 
     private final PresenceCheckEnabledGuard presenceCheckEnabledGuard;
 
-    private final ProcessIdentifierGuard processIdentifierGuard;
```
banterCZ: Remove this empty line. The guard class could also be removed, right?

```suggestion
```

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/IdentityVerificationRepository.java
```diff
@@ -88,7 +89,7 @@ AND EXISTS (
                    )
             """
     )
-    Stream<IdentityVerificationEntity> streamAllIdentityVerificationsToChangeState(final String documentVerificationProvider);
+    List<IdentityVerificationEntity> findAllIdentityVerificationsToChangeState(final String documentVerificationProvider, final Pageable pageable);
```
michal-rozehnal-w, line 74: Fetching a List with a limit should be more resilient than always processing the entire stream.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
```diff
@@ -0,0 +1,56 @@
...
+    private final StateMachineService stateMachineService;
+
+    private final IdentityVerificationService identityVerificationService;
+
+    /**
+     * Change machine states in batch.
+     */
+    public void changeMachineStatesInBatch() {
+        final var countFinished = identityVerificationService.findAllIdentityVerificationsToChangeState()
+                .parallelStream()
```
michal-rozehnal-w: I’m not entirely sure about the parallelism here. It was already used in the previous implementation, but maybe we can remove it.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -117,40 +139,10 @@ public Message<OnboardingEvent> createMessage(OwnerId ownerId, String processId,
                 .build();
     }
 
-    /**
-     * Change machine states in batch.
-     */
-    @Transactional(readOnly = true)
-    public void changeMachineStatesInBatch() {
```
michal-rozehnal-w, line 124: Method moved into separate class [StateMachineBatchService](https://github.com/wultra/enrollment-server/pull/1750/changes#diff-6a7f25db7b6e61e4121c35a20ce54ac8629c3d55aeffa0f3fbcd6cbebcdb5e29), so the `TransactionTemplate` is not needed.
I also removed the `@Transactional(readOnly = true)`, because I think it is not needed. Identity verification records are now fetched as List with limit instead of stream.
Execution of this method is controlled by `StateMachineTask` using `shedlock` and it should make sure there is only one active task.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -81,6 +79,30 @@ public StateMachine<OnboardingState, OnboardingEvent> processStateMachineEvent(O
...
+
+        try {
+            lockAndVerifyProcess(processId);
+            processStateMachineEvent(ownerId, processId, OnboardingEvent.EVENT_NEXT_STATE);
+            return true;
+        } catch (IdentityVerificationException e) {
+            logger.warn("Unable to change state for process ID: {}", processId, e);
+        } catch (final OnboardingProcessException e) {
+            logger.warn("Process with ID {} not found", processId, e);
+        } catch (final RuntimeException e) {
```
michal-rozehnal-w, line 110: In previous implementation there was `transactionTemplate.executeWithoutResult(...)`. To make sure that exception is not propagated - which would break the for-loop in upper layer - I added catch for the `RuntimeException`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
michal-rozehnal-w: Removed

docs/onboarding/PowerAuth-Enrollment-Onboarding-Server-2.2.0.md
```diff
@@ -0,0 +1,8 @@
+# Migration from 2.1.x to 2.2.x
+
+This guide contains instructions for migration from PowerAuth Enrollment Onboarding Server version `2.1.0` to version `2.2.x`.
+
+
+## Onboarding process state machine
+
+Added limit for identity verification records processed by a single scheduled task calling next state. The default limit is `10_000` and can be configured by property `enrollment-server-onboarding.identity-verification.next-state-batch-size`.
```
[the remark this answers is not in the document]
michal-rozehnal-w: True, this change extends the scope, but in my opinion, it is still related.

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/IdentityVerificationRepository.java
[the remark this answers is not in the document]
michal-rozehnal-w: The ordering is not necessary; the goal here is to process all records matching the criteria. However, ordering can be done in case we want to process from the oldest records. But I still consider is as optional

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
```diff
@@ -508,12 +508,15 @@ public VerificationSdkInfo initVerificationSdk(OwnerId ownerId, Map<String, Stri
...
+     * Return all identity verifications eligible for change to the next state.
      *
      * @return identity verifications
      */
-    public Stream<IdentityVerificationEntity> streamAllIdentityVerificationsToChangeState() {
-        return identityVerificationRepository.streamAllIdentityVerificationsToChangeState(identityVerificationConfig.getDocumentVerificationProvider());
+    public List<IdentityVerificationEntity> findAllIdentityVerificationsToChangeState() {
+        final var provider = identityVerificationConfig.getDocumentVerificationProvider();
+        final var batchSize = identityVerificationConfig.getNextStateBatchSize();
+
```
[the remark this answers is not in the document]
michal-rozehnal-w: Good idea; however, it is not easy to implement. Ideally, I would like the server not to start in the case of an invalid configuration (e.g., a negative number), so the `@Positive` annotation seems like a good choice. However, it will not work for a class with the `@Configuration` annotation. I prefer to leave it as it is in this PR and create a separate issue for adding validations (not only for this field). For now, in the worst-case scenario (a negative value is set), the scheduled task will log an error about the invalid value, and the logic will not be executed.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -81,6 +79,30 @@ public StateMachine<OnboardingState, OnboardingEvent> processStateMachineEvent(O
         return stateMachine;
     }
 
+    @SuppressWarnings("java:S6809") // New transaction is always created in this method, so @Transactional on the method processStateMachineEvent shouldn't cause any issue
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    boolean changeMachineState(final IdentityVerificationEntity identityVerification) {
```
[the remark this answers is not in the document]
michal-rozehnal-w: IMHO this is not true. In Spring version we use (3.5) the annotation is applied to the package-private/protected method as well.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -163,4 +155,9 @@ private StateMachine<OnboardingState, OnboardingEvent> fetchStateMachine(
         return prepareStateMachine(processId, onboardingState, identityVerification);
     }
 
+    private void lockAndVerifyProcess(final String processId) throws OnboardingProcessException {
+        onboardingProcessRepository.findByIdWithLock(processId)
+                .filter(p -> OnboardingStatus.NOT_YET_COMPLETED.contains(p.getStatus()))
+                .orElseThrow(() -> new OnboardingProcessException("Onboarding process not found for process ID: " + processId));
```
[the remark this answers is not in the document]
michal-rozehnal-w: Resolving, since this is same as previous note

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -163,4 +170,10 @@ private StateMachine<OnboardingState, OnboardingEvent> fetchStateMachine(
...
```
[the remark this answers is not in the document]
michal-rozehnal-w: True, but we use this pattern in other parts of code. I look at it from perspective that if the process with given id exists but the status is invalid, then there isn't suitable record. Anyway, I am open to any other suggestions

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
michal-rozehnal-w: I added the issue as blocker to this one

enrollment-server-onboarding-common/src/main/java/com/wultra/app/onboardingserver/common/database/IdentityVerificationRepository.java
banterCZ: Yes, the stream was not a good idea.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Service for changing state machine states in batch.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@Service
+@Slf4j
+@AllArgsConstructor
+@ConditionalOnProperty(value = "enrollment-server-onboarding.identity-verification.enabled", havingValue = "true")
```
banterCZ, line 35: 🤔 We should revisit `enrollment-server-onboarding.identity-verification.enabled` later on. I'm afraid this is a leftover from times when Enrollment and Onboarding were a single WAR. It does not make much sense to run Onboarding with this set to `false`.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
banterCZ: I would rather remove it.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public void changeMachineStatesInBatch() {
+        final var countFinished = identityVerificationService.findAllIdentityVerificationsToChangeState()
+                .parallelStream()
+                .map(stateMachineService::changeMachineState)
+                .filter(result -> result)
+                .count();
+
+        if (countFinished > 0) {
+            logger.debug("Changed state of {} identity verifications", countFinished);
```
banterCZ: 🤔 What about logging the count at INFO level?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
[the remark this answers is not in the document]
banterCZ: Interesting, I was not sure whether package-private methods are proxied.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
banterCZ: I am thinking about a possible poisoning the queue. This glitch used to already be here before.

enrollment-server-onboarding/src/main/resources/application.properties
```diff
@@ -109,6 +109,7 @@ enrollment-server-onboarding.client-evaluation.include-extracted-data=false
 enrollment-server-onboarding.identity-verification.enabled=${ENROLLMENT_SERVER_ONBOARDING_IDENTITY_VERIFICATION_ENABLED:false}
 enrollment-server-onboarding.identity-verification.max-failed-attempts=5
 enrollment-server-onboarding.identity-verification.max-failed-attempts-document-upload=5
+#enrollment-server-onboarding.identity-verification.next-state-batch-size=10000
```
banterCZ: Why is it commented out?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -163,4 +155,9 @@ private StateMachine<OnboardingState, OnboardingEvent> fetchStateMachine(
         return prepareStateMachine(processId, onboardingState, identityVerification);
     }
 
+    private void lockAndVerifyProcess(final String processId) throws OnboardingProcessException {
+        onboardingProcessRepository.findByIdWithLock(processId)
+                .filter(p -> OnboardingStatus.NOT_YET_COMPLETED.contains(p.getStatus()))
```
banterCZ, line 171: I believe that this should never happen, but I do not mind being safe.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
```diff
@@ -120,8 +119,6 @@ public class StateMachineConfig extends EnumStateMachineConfigurerAdapter<Onboar
...
```
banterCZ, line 108: May we delete the `ProcessIdentifierGuard` class completely?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -81,6 +79,30 @@ public StateMachine<OnboardingState, OnboardingEvent> processStateMachineEvent(O
         return stateMachine;
     }
 
+    @SuppressWarnings("java:S6809") // New transaction is always created in this method, so @Transactional on the method processStateMachineEvent shouldn't cause any issue
```
banterCZ: 🤔 May we introduce `private processStateMachineEventInternal` (or some better name) without annotation and keep `public processStateMachineEvent` to avoid confusion?

banterCZ (commented):
In general, I approve. Thanks for the cleanup. I will formally approve it after merging the refactoring of `NEXT_EVENT`. I have added several comments and ideas to consider.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
michal-rozehnal-w: I agree

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
michal-rozehnal-w: Removed

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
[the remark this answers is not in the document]
michal-rozehnal-w: They are applied for class-based proxies according to [this](https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html#transaction-declarative-annotations-method-visibility) documentation

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
michal-rozehnal-w: Good idea, cleaner than using the `@SuppressWarnings`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
michal-rozehnal-w: Changed to the INFO level

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
michal-rozehnal-w: Well spotted. I can create a separate issue for it. The invalid record will appear in the iteration again and again until it is marked as FAILED by the expiration task.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/StateMachineConfig.java
michal-rozehnal-w: Of course, I forgot about it. Deleted

enrollment-server-onboarding/src/main/resources/application.properties
michal-rozehnal-w: I put default value directly into `@Value` property. Anyway, it can be uncommented here as well since the value is same - `10_000`

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineServiceTest.java
```diff
@@ -45,31 +61,58 @@ class StateMachineServiceTest {
...
+    @MockitoBean
+    private DocumentVerificationProvider documentVerificationProvider;
+
+    @MockitoBean
+    private VerificationProcessingService verificationProcessingService;
+
     @Sql
-    void testProcessStateMachineEvent_documentUploaded() throws Exception {
+    @Test
+    void testProcessStateMachineEvent_fail() throws Exception {
```
michal-rozehnal-w, line 78: This test's implementation and purpose were changed because originally it did not pass the guard and no exception was thrown. Instead, neither the phase nor the state was changed. As a result, the test was effectively verifying that the process was not found.

For the `processStateMachineEvent` method (which is always called by REST API endpoint), there is currently no guard or similar process existence check. This is because such validation is performed before the method is called:

- `AcknowledgeService`: `onboardingProcessRepository.findByIdWithLock(...)`
- `IdentityVerificationRestService`: `onboardingService.verifyProcessIdAndLock(...)`

Therefore, having an additional guard in `processStateMachineEvent` would be redundant.

The test now verifies that an exception thrown during processing causes the transaction to rollback and that neither the phase nor the state is changed.

michal-rozehnal-w:
# Local testing

Onboarding for process type `onboardingSimple` passes

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
banterCZ: I have opened an issue #1788

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
```diff
@@ -0,0 +1,56 @@
...
+     * Change machine states in batch.
+     */
+    public void changeMachineStatesInBatch() {
+        final var countFinished = identityVerificationService.findAllIdentityVerificationsToChangeState()
+                .stream()
+                .map(stateMachineService::changeMachineState)
+                .filter(result -> result)
+                .count();
+
+        if (countFinished > 0) {
```
banterCZ, line 52: 🤔 No strong opinion, but shouldn't we log that always, even with zero?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
```diff
@@ -81,6 +87,29 @@ public StateMachine<OnboardingState, OnboardingEvent> processStateMachineEvent(O
...
+        ownerId.setActivationId(identityVerification.getActivationId());
+        ownerId.setUserId(identityVerification.getUserId());
+        logger.debug("Changing state of machine for process ID: {}", processId);
+
+        try {
+            lockAndVerifyProcess(processId);
+            processStateMachineEventInternal(ownerId, processId, OnboardingEvent.EVENT_NEXT_STATE);
+            return true;
+        } catch (IdentityVerificationException e) {
+            logger.warn("Unable to change state for process ID: {}", processId, e);
```
banterCZ: 🤔 Shouldn't we use `processId:` or even `kv()` pattern to improve log parsing?

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchServiceTest.java
```diff
@@ -0,0 +1,124 @@
+/*
+ * PowerAuth Enrollment Server
+ * Copyright (C) 2023 Wultra s.r.o.
```
banterCZ:
```suggestion
 * Copyright (C) 2026 Wultra s.r.o.
```

banterCZ (approved):

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchServiceTest.java
michal-rozehnal-w: Fixed

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
michal-rozehnal-w: Good idea, I added structured logging for start, success and failure

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
michal-rozehnal-w: The `if` block was originally here, and it's probably useful to keep it because the scheduled next-event task is triggered every 3 seconds by default. Without this check, the logs can be flooded with information that isn't particularly useful.
Maybe we can add the logging, but with debug level?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationService.java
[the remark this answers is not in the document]
michal-rozehnal-w: I added issue for it https://github.com/wultra/enrollment-server/issues/1789

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineService.java
michal-rozehnal-w: Created issue for poisoning the queue https://github.com/wultra/enrollment-server/issues/1790

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/statemachine/service/StateMachineBatchService.java
banterCZ: It sounds reasonable, so let's keep it as is.

michal-rozehnal-w:
@zcgandcomp adding you as an optional reviewer

PR #1757, opened 2026-05-04
Fix #1756: Hash#sha256 throws CryptoProviderException

### LINKED ISSUE #1756: Hash#sha256 throws CryptoProviderException
### Task specification

https://github.com/wultra/powerauth-crypto/issues/927

banterCZ:


enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/integration/OwnerId.java
```diff
@@ -83,4 +84,11 @@ public String getUserIdSecured() {
         return userIdSecured;
     }
 
+    private static byte[] sha256(final String source) {
+        try {
+            return Hash.sha256(source);
+        } catch (CryptoProviderException e) {
+            throw new IllegalStateException("sha256 is not supported", e);
```
romanstrobl: This will lead to a `500` error, do we want it?

banterCZ: 🤔 „It should never happen“, in that case, 500 is ok. Do you see any scenario when it could fail, and how we should react?

romanstrobl: Thoretically in case of a limit FIPS profile the algorithm may not be available in the future.

The reaction is to log the problem as an error, thrown an exception and handle it and return a 4xx code.

romanstrobl: This is how I handled it on the server: https://github.com/wultra/powerauth-server/pull/2364/changes The logging is enough by the exception handler, I think.

banterCZ: 🤔 I have two concerns:
- This code is called in the context of the state machine, not from the controller.
- How is it the client's error that the hash of the userId fails?

banterCZ: Since it is a security exception, I am adding at least the logger.

enrollment-server-onboarding-domain-model/src/main/java/com/wultra/app/enrollmentserver/model/integration/OwnerId.java
```diff
@@ -83,4 +86,12 @@ public String getUserIdSecured() {
         return userIdSecured;
     }
 
+    private static byte[] sha256(final String source) {
+        try {
+            return Hash.sha256(source);
+        } catch (CryptoProviderException e) {
```
michal-rozehnal-w, line 92: Nitpick: The exception can be `final`

michal-rozehnal-w (approved):
Looks good

PR #1758, opened 2026-05-04
Remove todo

banterCZ:
We need to keep `v3.PowerAuthClient`


romanstrobl (approved):
Thanks for the cleanup.

PR #1768, opened 2026-05-07
Fix #1751: Implement Process Events for Digital Onboarding

### LINKED ISSUE #1751: Implement Process Events for Digital Onboarding
### Description

Implement following process events for Digital Onboarding:
- DOCUMENT_VERIFICATION_FINISHED
- FINAL_DOCUMENT_VERIFICATION_FINISHED
- PRESENCE_CHECK_FINISHED
- PROCESS_FINISHED

Detailed [specification can be found here](https://github.com/wultra/enrollment-server/blob/features/1738-events-documentation/docs/onboarding/Events.md).

**Notes**
- we already have `processEvent` implemented... please check the possibility to extend this old event (preferred) or decide to create a new one

### Acceptance criteria
- events are implemented
- documentation is updated

### Technical specification
- events can share generic `onboarding-adapter` configuration (url, basic auth)
- add possibility to configure list of event types via system property (e.g. `enrollment-server-onboarding.process-event.types`)

### QA specification
- check configuration
- check if the events are sent to the target (try different configuration)
- check the structure of all event types
- check if the documentation on the portal is updated

banterCZ:


banterCZ:
Depends on #1774

enrollment-server-onboarding/src/test/java/com/wultra/app/onboardingserver/provider/rest/ProcessEventRequestDtoSerializationTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Tests verifying that {@link ProcessEventRequestDto} serializes
+ * to the JSON shape documented in {@code docs/onboarding/Events.md}.
```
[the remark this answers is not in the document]
banterCZ: It will be added in #1774

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/IdentityVerificationFinishService.java
```diff
@@ -113,8 +112,9 @@ public void finishIdentityVerification(OwnerId ownerId) throws RemoteCommunicati
 
     private void sendFinishedEvent(final OnboardingProcessEntity process, final IdentityVerificationEntity identityVerification, final OwnerId ownerId) {
         final ProcessEventRequest request = ProcessEventRequest.builder()
-                .type(ProcessEventRequest.EventType.FINISHED)
+                .type(ProcessEventRequest.EventType.PROCESS_FINISHED)
                 .userId(identityVerification.getUserId())
+                .externalUserId(process.getUserId()) // TODO Lubos store and get iProov userId
                 .processId(process.getId())
```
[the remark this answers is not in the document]
banterCZ: In this stage, the userId has to be already known.

docs/onboarding/Configuration-Properties.md
```diff
@@ -29,18 +29,19 @@ The configuration properties below are global to all the onboarding processes.
...
+| `enrollment-server-onboarding.onboarding-process.max-processes-per-day`   | `5`       | Maximum number of onboarding processes during last 24 hours per user.                                                                                                |
+| `enrollment-server-onboarding.onboarding-process.max-error-score`         | `15`      | Maximum error score for an onboarding process.                                                                                                                       |
+| `enrollment-server-onboarding.onboarding-process.default-type`            | `_empty_` | Default process type used for starting onboarding if not specified by the mobile application.                                                                        |
+| `enrollment-server-onboarding.onboarding-process.process-event.types`     | `_all_`   | Event types to publish; possible values `PROCESS_FINISHED`, `DOCUMENT_VERIFICATION_FINISHED`, `FINAL_DOCUMENT_VERIFICATION_FINISHED`, and `PRESENCE_CHECK_FINISHED`. |
```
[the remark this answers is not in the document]
banterCZ: I do not want to list too many options, I think that _all_ is clear.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
```diff
@@ -0,0 +1,380 @@
...
+                .processId(process.getId())
+                .processType(process.getProcessConfiguration().getProcessType())
+                .userId(identityVerification.getUserId())
+                .externalUserId(process.getExternalUserId())
+                .identityVerificationId(identityVerification.getId());
+    }
+
+    private boolean isEventTypeNotEnabled(final EventType eventType) {
+        final List<EventType> enabled = onboardingConfig.getEventTypes();
+        if (enabled == null || !enabled.contains(eventType)) {
```
michal-rozehnal-w: We can use empty collection as default value for `onboardingConfig.getEventTypes()`, so the check for `null` is not needed.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
```diff
@@ -0,0 +1,380 @@
...
+            return null;
+        }
+        final DocumentExtractedDataValue value;
+        try {
+            value = objectMapper.readValue(result.getExtractedData(), DocumentExtractedDataValue.class);
+        } catch (JacksonException e) {
+            logger.warn("Unable to parse extracted data for documentResultId={}: {}", result.getId(), e.getMessage());
+            return null;
+        }
+        return DocumentVerificationFinishedEventData.DocumentData.builder()
```
michal-rozehnal-w, line 303: In case of `value = null` we should just return empty object. In current implementation, the `value.surname()` (and any other property) is always called and would throw NPE.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
```diff
@@ -0,0 +1,380 @@
...
+
+    private EventData createFinalDocumentVerificationFinishedEventData(
+            final IdentityVerificationEntity identityVerification,
+            final EventStatus status,
+            final String rejectReason,
+            final String errorDetail) {
+
+        final List<String> documentIds = identityVerification.getDocumentVerifications().stream()
+                .filter(DocumentVerificationEntity::isUsedForVerification)
+                .map(it -> it.getUploadId() != null ? it.getUploadId() : it.getId())
```
michal-rozehnal-w: Isn't `uploadId` always set? If not, maybe we should log it. I am just little bit confused we use 2 different ids for same field

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
```diff
@@ -0,0 +1,380 @@
...
+                .rejectReason(rejectReason)
+                .errorDetail(errorDetail)
+                .provider(identityVerificationConfig.getDocumentVerificationProvider())
+                .documentIds(documentIds)
+                .build();
+    }
+
+    private EventData createPresenceCheckFinishedEventData(final PresenceCheckResult result) {
+        final Image photo = result.getPhoto();
+        final PresenceCheckFinishedEventData.PresenceCheckResult presenceCheckResult = (photo == null || photo.getData() == null)
```
michal-rozehnal-w: The condition is little complex. Maybe in this case `Optional` is better

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
```diff
@@ -0,0 +1,380 @@
...
+    private List<DocumentVerificationFinishedEventData.Image> buildImages(final DocumentVerificationEntity doc) {
+        final List<ProcessedDocumentDataEntity> entities =
+                processedDocumentDataRepository.findAllByDocumentVerificationIds(Set.of(doc.getId()));
+        if (entities.isEmpty()) {
+            return List.of();
+        }
+        return entities.stream()
+                .map(it -> DocumentVerificationFinishedEventData.Image.builder()
+                        .type(it.getDataType().name())
+                        .data(Base64.getEncoder().encodeToString(it.getData()))
```
michal-rozehnal-w, line 327: I am fine with the current implementation. One thing to consider is creating a separate utility method for Base64 encoding/decoding that handles exceptions and performs logging. This is a fairly common operation and could be centralized in a utility method.

I think the current flow is correct and should remain unchanged - we want to throw the exception and propagate it to prevent sending corrupted data.

On the other hand, an exception thrown here would indicate that we have corrupted data in the database, which should never happen under normal circumstances.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/model/request/DocumentVerificationFinishedEventData.java
```diff
@@ -0,0 +1,115 @@
...
+     * Present only when {@code status} is {@code ACCEPTED} or {@code REJECTED}.
+     */
+    private DocumentVerificationResult documentVerificationResult;
+
+    @Builder
+    @Getter
+    @ToString
+    @PublicApi
+    @EqualsAndHashCode
+    public static class DocumentVerificationResult {
```
michal-rozehnal-w: Are `DocumentVerificationResult`, `DocumentData` and `Image` of type `class` on purpose or can we use `record`? I can see for *`.Dto` objects we use `record` type

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/model/request/EventData.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package com.wultra.app.onboardingserver.provider.model.request;
+
+/**
+ * Marker interface for event payload data.
+ * The concrete shape differs per {@link EventType}.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public sealed interface EventData
```
michal-rozehnal-w, line 27: Nice usage of `sealed` and `permits`

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/model/request/ProcessEventRequest.java
```diff
@@ -48,6 +47,13 @@ public final class ProcessEventRequest {
...
 
+    /**
+     * User ID in the external system, which is used by the provider.
+     * It can be the same as {@code userId} or different, depending on the provider implementation.
+     */
+    @NonNull
+    private String externalUserId;
+
     @NonNull
     private String identityVerificationId;
```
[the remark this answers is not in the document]
michal-rozehnal-w: We are aware of it, and it is acceptable

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/rest/RestOnboardingProvider.java
```diff
@@ -228,23 +230,150 @@ private static ApproveClientRequestDto.Status convert(final ApproveClientRequest
...
+        return switch (source) {
+            case PROCESS_FINISHED -> EventTypeDto.PROCESS_FINISHED;
+            case DOCUMENT_VERIFICATION_FINISHED -> EventTypeDto.DOCUMENT_VERIFICATION_FINISHED;
+            case FINAL_DOCUMENT_VERIFICATION_FINISHED -> EventTypeDto.FINAL_DOCUMENT_VERIFICATION_FINISHED;
+            case PRESENCE_CHECK_FINISHED -> EventTypeDto.PRESENCE_CHECK_FINISHED;
+        };
+    }
+
+    private static EventDataDto convert(final EventData source) {
+        if (source instanceof ProcessFinishedEventData data) {
```
michal-rozehnal-w, line 255: Can we use `switch` with pattern matching, something like:

```
return switch (source) {
    case ProcessFinishedEventData data -> convert(data);
    case DocumentVerificationFinishedEventData data -> convert(data);
    case FinalDocumentVerificationFinishedEventData data -> convert(data);
    case PresenceCheckFinishedEventData data -> convert(data);
    default -> throw new IllegalArgumentException(
            "Unsupported event data type: " + source.getClass());
};
```

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/configuration/OnboardingConfig.java
```diff
@@ -59,4 +61,9 @@ public class OnboardingConfig extends CommonOnboardingConfig {
     @Value("${enrollment-server-onboarding.onboarding-process.default-type:}")
     private String defaultProcessType;
 
+    /**
+     * Event types to publish.
+     */
+    @Value("${enrollment-server-onboarding.onboarding-process.process-event.types}")
+    private List<EventType> eventTypes;
```
michal-rozehnal-w, line 68: Can we use `EnumSet` and empty collection as default value?

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
```diff
@@ -0,0 +1,380 @@
...
+ * {@link OnboardingProvider}.
+ * <p>
+ * Failures during event publishing never abort the underlying process; they are only logged.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@Service
+@Slf4j
+@AllArgsConstructor
+public class OnboardingEventService {
```
michal-rozehnal-w, line 60: Is this class sufficiently covered by tests?
I am thinking about test for each method creating a `ProcessEventRequest`, verifying that all expected values are correctly mapped.
Additionally, we could add tests for `enrollment-server-onboarding.onboarding-process.process-event.types`:
- event type is configured → event is sent
- event type is not configured → event is not sent

I do not insist, just idea what can be consider

michal-rozehnal-w (commented):
Looks good, I added a few comments

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
banterCZ: You are right, we cannot get null, since the property is specified and may be overridden only by an empty string, which leads to an empty collection anyway.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
michal-rozehnal-w: Sorry, my bad, I was thinking about `value = null`, but there is actually `return null`, so null value is never passed to the builder. This comment is no relevant

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/rest/RestOnboardingProvider.java
banterCZ: Sounds good, but works since JDK 21, not in 17.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/configuration/OnboardingConfig.java
banterCZ: We cannot get `null`, since the property is specified and may be overridden only by an empty string, which leads to an empty collection anyway. I have already improved the logic.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
banterCZ: I would postpone that in this PR. In general, I prefer to avoid many util classes and inline the code.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/provider/rest/RestOnboardingProvider.java
michal-rozehnal-w: Right, I didn't realize we are still on 17

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
banterCZ: A good idea, I have added a test.

enrollment-server-onboarding/src/main/java/com/wultra/app/onboardingserver/impl/service/OnboardingEventService.java
banterCZ: 😱 Well spotted, this was complete nonsense, mixing `uploadId` with `id`.

michal-rozehnal-w (approved):
Thank you for additional changes, looks  good to me know

PR #1782, opened 2026-05-14
Fix #1781: Configure annotationProcessorPaths for Lombok

### LINKED ISSUE #1781: Fix: Support Lombok with Java 25
## Task specification

Lombok 1.18.x annotation processing fails on Java 25 — the `logger` field generated by `@Slf4j` is not visible to `javac`:

```
cannot find symbol
  symbol:   variable logger
```

Affects local development on Java 25 (any vendor). CI uses Java 21 and is unaffected.

**Root cause:** Java 25 changed internal compiler APIs that Lombok hooks into via implicit annotation processing.

**Fix:** Explicitly register Lombok via `annotationProcessorPaths` in the root `pom.xml`:

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <configuration>
        <annotationProcessorPaths>
            <path>
                <groupId>org.projectlombok</groupId>
                <artifactId>lombok</artifactId>
            </path>
        </annotationProcessorPaths>
    </configuration>
</plugin>
```

vita-kotacka:
## Summary

Fixes Lombok annotation processing on Java 25 by explicitly registering Lombok via `annotationProcessorPaths` in `maven-compiler-plugin`.

Closes #1781

## Problem

Lombok 1.18.x relies on implicit annotation processor discovery which breaks on Java 25 — the `logger` field (and other generated members) are not visible to `javac`, causing:

```
cannot find symbol
  symbol: variable logger
```

CI (Java 21) is unaffected; the issue only manifests locally when using Java 25.

## Solution

Add explicit `annotationProcessorPaths` configuration to `maven-compiler-plugin` in the root `pom.xml`. This bypasses the implicit APT discovery and works correctly on Java 25.

The existing `provided` Lombok dependency is retained — it is still needed to make Lombok annotations (`import lombok.*`) available on the compile classpath.

## Changes

| File | Change |
|---|---|
| `pom.xml` | Added `maven-compiler-plugin` with `annotationProcessorPaths` for Lombok |

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -312,6 +312,19 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.projectlombok</groupId>
+                            <artifactId>lombok</artifactId>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <excludedGroups>external-service</excludedGroups>


banterCZ (approved):
Thanks!

michal-rozehnal-w (approved):

PR #1786, opened 2026-05-21
Add `onFail=MARK_RAN` to preconditions in processed document data table changelog

banterCZ:
A follow-up to #1456


michal-rozehnal-w (approved):
Thank you very much for the fix

PR #1792, opened 2026-05-26
Fix #1791: Change iProov resource attribute

### LINKED ISSUE #1791: Change iProov resource attribute
### Task specification

When calling the claim verification token on iProov, the `resource` attribute should be:
`The resource being accessed (e.g., URL)` (see iProov documentation)

So we can give some means to our API to enable setting what resource is being accessed

In our case, use `enrollment-onboarding-server` as a resource

banterCZ:


enrollment-server-onboarding-provider-iproov/src/main/java/com/wultra/app/onboardingserver/provider/iproov/IProovRestApiService.java
```diff
@@ -178,7 +178,7 @@ public ResponseEntity<String> validateVerification(OwnerId id, String token) thr
         final ClaimValidateRequest request = new ClaimValidateRequest();
         request.setApiKey(configProps.getApiKey());
         request.setSecret(configProps.getApiSecret());
-        request.setClient("Wultra Enrollment Server, activationId: " + id.getActivationId());
+        request.setClient(I_PROOV_RESOURCE);
```
banterCZ, line 181:
```
"client": {
          "description": "Fingerprint or client identifier of the device making the request (e.g. User Agent)",
          "type": "string"
        },
```

enrollment-server-onboarding-provider-iproov/src/main/java/com/wultra/app/onboardingserver/provider/iproov/IProovRestApiService.java
```diff
@@ -76,7 +76,7 @@ class IProovRestApiService {
      */
     private static final Pattern USER_ID_REGEX_PATTERN = Pattern.compile("[a-zA-Z0-9'+_@.-]{1,256}");
 
-    private static final String I_PROOV_RESOURCE_CONTEXT = "presence_check/";
+    private static final String I_PROOV_RESOURCE = "enrollment-onboarding-server";
```
michal-rozehnal-w, line 79: It seems we don't have tests verifying that the request body sent to iProov is constructed correctly and that all required fields are present. We can add these tests in a separate PR.

michal-rozehnal-w (approved):
Looks good from implementation perspective

PR #1793, opened 2026-05-26
Fix #1788: Remove `enrollment-server-onboarding.identity-verification.enabled`

### LINKED ISSUE #1788: Remove `enrollment-server-onboarding.identity-verification.enabled`
### Task specification

We should revisit `enrollment-server-onboarding.identity-verification.enabled`. I'm afraid this is a leftover from times when Enrollment and Onboarding were a single WAR. It does not make much sense to run Onboarding with this set to `false`.

banterCZ:


michal-rozehnal-w (approved):
Nice cleanup!

PR #1794, opened 2026-05-29
Fix #1760: Change user data parsing from document verification response

### LINKED ISSUE #1760: Onboarding - change user data parsing from document verification response
### Task specification

Response from document verification provider contains extracted data. The data has following structure:
```
"extraction": {
  "overall":  [extracted data],
  "viz": {
    "front": [extracted data],
    "back": [extracted data]
  }
}
```

Currently, we store front and back side of the document in the database as separate rows and we are PROBABLY parse front side data from `extraction.viz.front` and back side data from`extraction.viz.back`.

I would take data for both sides from `extraction.overall` which contains merged data from whole document. It would create more consistent results.

In addition, response can contain selected fields in multiple languages, for example `Latin` and `Cyrilic`. We probably didn't expect that there can be same fields multiple times (in different languages).

Example:
```
  "extraction": {
    "overall": [
      {
        "side": "Unknown",
        "script": "Latin",
        "value": "OCTAVIAN",
        "type": "DetailedStringResult",
        "field": "FirstName"
      },
      {
        "side": "Unknown",
        "script": "Cyrillic",
        "value": "ОКТАВИАН",
        "type": "DetailedStringResult",
        "field": "FirstName"
      },
      ...
      ]
    }
```
Always use `Latin` script (if available).

The rules have to be applied everywhere:
- storing the data in the DB
- document crosscheck
- Evaluation Service outputs

michal-rozehnal-w:
# Functional description

Data extracted from the document via the Microblink provider is taken from the `extraction.overall` field in the JSON response. In the case of documents with multiple sides, the same value is stored in `es_document_result.extracted_data` because it is not possible to filter fields by side.

A value with `"script": "Latin"` is preferred from JSON response. If no such value exists, the first one is used.


enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkDocumentVerificationProvider.java
```diff
@@ -379,13 +376,8 @@ private List<DocumentSubmitResult> processMicroblinkResults(
         for (final var documentVerificationData : documentsVerificationData) {
             final var microblinkResponse = microblinkResponseByDocumentType.get(documentVerificationData.type());
 
-            final var extractedData = switch (documentVerificationData.side()) {
-                case FRONT -> microblinkResponse.getExtractionFront();
-                case BACK -> microblinkResponse.getExtractionBack();
-            };
-
+            final var extractedData = microblinkResponse.getOverallExtraction();
```
michal-rozehnal-w, line 379: This causes the same data to be stored for both the front and back sides in the database `es_document_result.extracted_data`. This shouldn't change any business logic, and we already handle `es_document_result.verification_result` the same way.
I was thinking about filtering by the `side` field in the original Microblink JSON, but the field is set to `Unknown` for all properties.

michal-rozehnal-w:
# Local testing

- A value with `"script": "Latin"` is preferred. If no such script is available, the first matching value is used for the field.
- Correct values extracted from Microblink are stored in the `es_document_result.extracted_data` database field.
- Correct values are sent to the `evaluate` endpoint.
- Correct values are stored in the `user-data-store`.

docs/onboarding/Changelog.md
```diff
@@ -11,6 +11,7 @@
 - Removed `Serializable` from JPA entities. [(1425)](https://github.com/wultra/enrollment-server/issues/1425)
 - Refactor state machine configuration, limit `EVENT_NEXT_STATE` and speed up processing. [(1592)](https://github.com/wultra/enrollment-server/issues/1592)
 - Migrated to Spring Boot 4 and Jackson 3. [(1775)](https://github.com/wultra/enrollment-server/issues/1775)
+- Changed field for parsing Microblink document extracted data. [#1760](https://github.com/wultra/enrollment-server/issues/1760)
```
banterCZ: Nitpick, we should define copilot instructions, but I think that guys are using this guideline and don't use hash.
```suggestion
- Changed field for parsing Microblink document extracted data. [(1760)](https://github.com/wultra/enrollment-server/issues/1760)
```

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkExtractedDataParser.java
```diff
@@ -102,15 +102,21 @@ private static Map<ExtractedDataField, ExtractedValue> extractFieldValues(final
             final ExtractedValue value = extractValue(node);
 
             if (value != null) {
-                result.put(field, value);
+                final ExtractedValue existingValue = result.get(field);
+                if (existingValue == null || isLatinScript(value)) {
```
banterCZ, line 108: 🤔 Ok, so this is a fallback to have at least Cyrillic one, and if Latin is available, rewrite the value, right?

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkExtractedDataParser.java
```diff
@@ -126,6 +132,10 @@ private static boolean isSuccessfullyParsed(final JsonNode node) {
         return node.path("successfullyParsed").asBoolean(false);
     }
 
+    private static boolean isLatinScript(final ExtractedValue value) {
+        return (value instanceof ExtractedValue.Text text) && "Latin".equalsIgnoreCase(text.script());
```
banterCZ: Nitpick, don't we want to have a private constant for that?

banterCZ (approved):

docs/onboarding/Changelog.md
michal-rozehnal-w: Good catch, fixed

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkExtractedDataParser.java
michal-rozehnal-w: Yes, this is how I understand it from the issue description:
_Always use Latin script (if available)._

@tech-kratas could you plese confirm the logic for field extraction:
For any text field (the date field does not have a script property according to the [documentation](https://docs.microblink.com/verify/api/ref/v2#model/DetailedDateResult)), we first try to find a value with `script: Latin`. If nothing is found, we then parse any other value, regardless of its script. If the value is missing completely, nothing is parsed.

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkExtractedDataParser.java
michal-rozehnal-w: Right, I added the private constant

enrollment-server-onboarding-provider-microblink/src/main/java/com/wultra/app/onboardingserver/provider/microblink/MicroblinkExtractedDataParser.java
tech-kratas: Yes, it is correct.

I would expect the Latin script to be present almost every time, but we don't know about all the variations in documents. There might be documents in different languages, and we don't know the customer's preferences. So, in the future, we may need to add some configuration for the preferred language.

michal-rozehnal-w: Thank you, today only 4 [scripts](https://docs.microblink.com/verify/api/ref/v2#model/DetailedStringResult) are supported. Adding configuration in future is not a big deal, basically it will be limited just to this function.

PR #1799, opened 2026-06-04
Fix #1783: Condition for evaluation process state based on required documents

### LINKED ISSUE #1783: Condition for evaluation process state based on required documents
### Task specification

If the user uploads one document that fails and then starts uploading different documents, all other documents pass, and there are enough ACCEPTED documents, we will still get NOK. And we will go to the handleDocumentStatus, which checks all uploaded document statuses - i.e., the process can only fail.

There might be a mechanism to discard FAILED/REJECTED documents to prevent such situations, but we should not rely on it.

The criteria for the process evaluation:
**If the required number of documents is met, the process should proceed to the next stage.**

banterCZ:


michal-rozehnal-w (approved):
Looks good

michal-rozehnal-w (approved):

PR #1816, opened 2026-06-17
Fix #1815: Add transformer for onboarding events redirect

### LINKED ISSUE #1815: Add event redirection in mock adapter
### Task specification

It should be configurable, that `POST /process/event` is redirected to liveness-check-proxy POST /onboarding-events

michal-rozehnal-w:
# Function description

Added WireMock transformer for  `POST /process/event`. When following env variables are set:

```
LIVENESS_CHECK_PROXY_BASE_URL
LIVENESS_CHECK_PROXY_USERNAME
LIVENESS_CHECK_PROXY_PASSWORD
```

Then any traffic to that endpoint is redirected to the liveness-check-proxy `POST LIVENESS_CHECK_PROXY_BASE_URL/onboarding-events`.

If these env variables are not set, then static response 200 OK is returned and liveness-check-proxy is not called at all.

### DIFF
--- a/enrollment-server-onboarding-adapter-mock/bin/wiremock.sh
+++ b/enrollment-server-onboarding-adapter-mock/bin/wiremock.sh
@@ -3,10 +3,17 @@
 mydir=`dirname $0`
 cd $mydir/..
 
+# Liveness check proxy configuration (all optional). When LIVENESS_CHECK_PROXY_BASE_URL is set,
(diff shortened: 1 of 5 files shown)


michal-rozehnal-w:
# Local testing

Tested via `/bin/wiremock.sh`. When the env variables are set, the redirection to LCP works; without the env variables, the LCP is not called and static response is returned

michal-rozehnal-w:
Not needed, we solved redirection via mapping JSON file

PR #1820, opened 2026-06-18
Fix #1819: Populate message field for structured (kv) logs

### LINKED ISSUE #1819: Missing `message` field in Elasticsearch for structured (kv) logs
Parent epic: wultra/tasklist#992

### Problem
The `message` field in Elasticsearch is `null` for log lines that use the new structured `kv(...)` format. Old-style logs (with a real message string) display `message` correctly.

### Root cause
Log statements use an **empty message string** with structured arguments, e.g.:
```java
logger.info("", kv("action", "operationList"), kv("state", "initiated"));
```
The Logstash encoder (`net.logstash.logback.encoder.LogstashEncoder`) is configured with `<omitEmptyFields>true</omitEmptyFields>`, so the empty `message` field is dropped and indexed as `null` in Elasticsearch.

### Proposed fix
Add a concise, human-readable message to each structured log call while keeping the `kv(...)` fields:
```java
logger.info("Operation list initiated", kv("action", "operationList"), kv("state", "initiated"));
```
No logback/encoder config change required.

### Affected component
enrollment-server / enrollment-server-onboarding

vita-kotacka:
🤖

## Summary
Fixes #1819 (part of epic wultra/tasklist#992).

Structured log calls were logging an **empty message string**, e.g.:
```java
logger.info("", action("operationList"), stateInitiated(), kv("activationId", id));
```
Because the Logstash encoder is configured with `<omitEmptyFields>true</omitEmptyFields>`, the empty `message` field was dropped and indexed as `null` in Elasticsearch. Old-style logs (with a real message) were unaffected.

## Change
Add a concise, human-readable message to every structured log call, derived deterministically from the existing structured arguments — `action` + `state` (+ `reason` on failure/skip). All structured arguments (`action(...)`, `state...()`, `kv(...)`) are left **unchanged**, so existing Elastic dashboards/queries keyed on those fields keep working.

Example:
```java
logger.warn("Operation list failed: validation", action("operationList"), stateFailed(), kv("reason", "validation"));
```

Message templates by state: `initiated`→"… initiated", `succeeded`→"… succeeded", `finished`→"… finished", `failed`/`skipped`→"… failed/skipped: <reason>", `unsupported`→"… not supported".

## Scope
165 call sites across 21 files in `enrollment-server`, `enrollment-server-onboarding`, and the microblink/innovatrics provider modules. No logback/encoder config change (`omitEmptyFields=true` stays).

The other two components from the epic — `liveness-check-proxy` and `user-data-store` — are tracked by their own bug issues in their repositories.

## Validation
- `BUILD SUCCESS` (compile, JDK 25).
- No tests assert on log output.

### DIFF
--- a/enrollment-server-onboarding-provider-innovatrics/src/main/java/com/wultra/app/onboardingserver/provider/innovatrics/InnovatricsLivenessController.java
+++ b/enrollment-server-onboarding-provider-innovatrics/src/main/java/com/wultra/app/onboardingserver/provider/innovatrics/InnovatricsLivenessController.java
@@ -79,7 +79,7 @@ class InnovatricsLivenessController {
             @Parameter(hidden = true) final EncryptionContext encryptionContext,
             @Parameter(hidden = true) final PowerAuthApiAuthentication apiAuthentication) throws IdentityVerificationException, PowerAuthAuthenticationException, PowerAuthEncryptionException, RemoteCommunicationException {
 
-        logger.info("", action("upload"), stateInitiated(), kv("activationId", extractActivationId(apiAuthentication)));
+        logger.info("Upload initiated", action("upload"), stateInitiated(), kv("activationId", extractActivationId(apiAuthentication)));
         if (apiAuthentication == null) {
             throw new PowerAuthTokenInvalidException("Unable to verify device registration when uploading liveness");
         }
@@ -89,7 +89,7 @@ class InnovatricsLivenessController {
         }
 
         innovatricsLivenessService.upload(requestData, encryptionContext);
-        logger.info("", action("upload"), stateSucceeded());
+        logger.info("Upload succeeded", action("upload"), stateSucceeded());
         return new Response();
     }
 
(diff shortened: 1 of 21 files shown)


michal-rozehnal-w:
Thank you for the fix. I understand the issue now.
I'm approving this PR since it resolves the `null` `message` problem.

One minor observation: the information is now duplicated in both the `message` and the `kv()` fields. That said, they serve different purposes - the `message` is for human readers, while the `kv()` fields support log grouping and analysis. So I'm fine with this approach. I just wanted to mention it.

michal-rozehnal-w (approved):

vita-kotacka:
> One minor observation: the information is now duplicated in both the `message` and the `kv()` fields. That said, they serve different purposes - the `message` is for human readers, while the `kv()` fields support log grouping and analysis.

:hurtrealbad: @michal-rozehnal-w Actually, even the previous solution (with `null` message) was not a bug, but feature. 😁  It's rather about loggin convention and how users (developers, DevOps, customers, etc.) and tools work with logging records.

So rather than a specific solution - now with duplicities (which are fine in this case) - I would consider this as a continuing discussion about _Observability_ where we are still at the beginning. 😅

PR #1854, opened 2026-08-03
Fix #1853: [2.1.2] Fix failing checksum check for Liquibase migration

### LINKED ISSUE #1853: Fix failing liquibase migration from 2.1
### Task specification

The `2.1` release doesn't contain this fix: https://github.com/wultra/enrollment-server/issues/1752

Do this:
- use liquibase `validCheckSum` so the migration doesn't fail
- drop index `es_document_data_timestamp_created_idx` if exists

For following branches:
- `releases/2.1.x`
- `releases/2.2.x`
- `develop`

michal-rozehnal-w:
# Functional description

- Add liquibase `<validCheckSum>` for overriding already released and modified changeset
- Drop duplicated index `es_document_data_timestamp_created_idx` if exists

### DIFF
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml
@@ -64,7 +64,7 @@
     </changeSet>
 
     <changeSet id="6" logicalFilePath="enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml" author="Michal Rozehnal">
-        <validCheckSum>ANY</validCheckSum>
+        <validCheckSum>9:d4d9d5d4f6b825c46f301e0fdd12fd1c</validCheckSum>
         <preConditions>
             <not>
                 <indexExists tableName="es_document_verification" indexName="es_document_verification_upload_id_idx"/>
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml
@@ -64,6 +64,7 @@
     </changeSet>
 
     <changeSet id="6" logicalFilePath="enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml" author="Michal Rozehnal">
+        <validCheckSum>ANY</validCheckSum>
         <preConditions>
             <not>
                 <indexExists tableName="es_document_verification" indexName="es_document_verification_upload_id_idx"/>
--- a/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260430-drop-duplicate-document-data-index.xml
+++ b/docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260430-drop-duplicate-document-data-index.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.33.xsd">
+
+    <changeSet id="1" logicalFilePath="enrollment-server-onboarding/2.1.x/20260430-drop-duplicate-document-data-index.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <indexExists tableName="es_document_data" indexName="es_document_data_timestamp_created_idx"/>
+        </preConditions>
+        <comment>Drop duplicate index es_document_data_timestamp_created_idx (already covered by document_data_timestamp from 1.4.x)</comment>
+        <dropIndex tableName="es_document_data" indexName="es_document_data_timestamp_created_idx"/>
+    </changeSet>
+
+</databaseChangeLog>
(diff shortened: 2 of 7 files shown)


michal-rozehnal-w:
# Local testing

1. Checkout to git tag `1.10.2` and execute liquibase migration
2. Checkout to this feature branch with fix and execute again the migration

Migration passes and index `es_document_data_timestamp_created_idx` does not exist (applicable only to postgres).
I tested it locally both postgres and oracle

docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml
```diff
@@ -64,6 +64,7 @@
     </changeSet>
 
     <changeSet id="6" logicalFilePath="enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml" author="Michal Rozehnal">
+        <validCheckSum>ANY</validCheckSum>
```
banterCZ: 🤔  I would rather have specific values than `ANY`.

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260803-add-tag-2.1.2.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.33.xsd">
+
+    <changeSet id="1" logicalFilePath="enrollment-server-onboarding/2.1.x/20260803-add-tag-2.1.2.xml" author="Michal Rozehnal">
+        <tagDatabase tag="enrollment-server-onboarding/2.1.2"/>
```
banterCZ, line 7: Why not, I have been tagging just minor versions so far.

docs/sql/postgresql/onboarding/create-schema.sql
```diff
@@ -362,4 +362,7 @@ ALTER TABLE es_document_result ALTER COLUMN  anonymized SET DEFAULT FALSE;
 
 -- Changeset enrollment-server-onboarding/2.1.x/20260423-document-result-anonymized-column.xml::3::Michal Rozehnal
 -- Add index on anonymized column in es_document_result table
-CREATE INDEX es_document_result_anonymized_idx ON es_document_result(anonymized);
\ No newline at end of file
+CREATE INDEX es_document_result_anonymized_idx ON es_document_result(anonymized);
+-- Changeset enrollment-server-onboarding/2.1.x/20260430-drop-duplicate-document-data-index.xml::1::Michal Rozehnal
+-- Drop duplicate index es_document_data_timestamp_created_idx (already covered by document_data_timestamp from 1.4.x)
+DROP INDEX IF EXISTS es_document_data_timestamp_created_idx;
```
banterCZ: 🤔 Shouldn't this be in the particular version file instead of the `create-scheme.sql`? And I see only _postgres_ folder here.

banterCZ:
Shouldn't it be mentioned in the changelog?

docs/db/changelog/changesets/enrollment-server-onboarding/2.0.x/20260116-add-processed-document-data-table.xml
michal-rozehnal-w: Right, I used the exact value

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260803-add-tag-2.1.2.xml
michal-rozehnal-w: I don't have strong opinion about it, but since it is officially released version, we can use it.

docs/sql/postgresql/onboarding/create-schema.sql
michal-rozehnal-w: Right, the index is not created via `create-scheme.sql`, so no need to drop it.
On purpose there is not sql for oracle, because it is not possible to have more than one index for the same column. Postgres allows it, that's why the drop command is in place

docs/db/changelog/changesets/enrollment-server-onboarding/2.1.x/20260430-drop-duplicate-document-data-index.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.33.xsd">
+
+    <changeSet id="1" logicalFilePath="enrollment-server-onboarding/2.1.x/20260430-drop-duplicate-document-data-index.xml" author="Michal Rozehnal"
+               runInTransaction="false">
```
michal-rozehnal-w, line 7: The `runInTransaction="false"` was recommended by AI and according to documentation it is needed.

According to Liquibase [doc](https://docs.liquibase.com/oss/reference-guide-4-33/changelog-attributes/runintransaction):

> The runInTransaction attribute specifies whether a changeset can be run as a single transaction (if possible). The default value is true.

According to Postgres [doc](https://www.postgresql.org/docs/current/sql-dropindex.html):

> There are several caveats to be aware of when using this option. Only one index name can be specified, and the CASCADE option is not supported. (Thus, an index that supports a UNIQUE or PRIMARY KEY constraint cannot be dropped this way.) Also, regular DROP INDEX commands can be performed within a transaction block, but DROP INDEX CONCURRENTLY cannot. Lastly, indexes on partitioned tables cannot be dropped using this option.

michal-rozehnal-w:
> Shouldn't it be mentioned in the changelog?

Good catch, it seems we don't have changelog in `2.1` (I am not sure we want to add it just for minor release `2.1.0` -> `2.1.2`). But for sure it should be mentioned in migration guide. I added description there.

banterCZ:
> > Shouldn't it be mentioned in the changelog?
> 
> Good catch, it seems we don't have changelog in `2.1` (I am not sure we want to add it just for minor release `2.1.0` -> `2.1.2`). But for sure it should be mentioned in migration guide. I added description there.

I see, we started to use the changelog since 2.2.0, but maybe we could backport that; it helps to track changes.

banterCZ (approved):

